PR #101, opened 2023-07-25
fix: re-compute quantization params when re-fitting linear models

RomanBredehoft:
when re-fitting linear models, quantizers were appended to a list without ever being used. As for tree-baeed models, we now re-compute them at each fit

closes [zama-ai/concrete-ml-internal/issues/#3865](https://github.com/zama-ai/concrete-ml-internal/issues/3865)


andrei-stoian-zama (approved):
Nice fix!

jfrery (approved):
Thanks for the fix!

kcelia (approved):

andrei-stoian-zama (approved):

PR #103, opened 2023-07-26
chore: remove randomness from gpt2 hybrid test

jfrery:
should fix the gpt2 test now using the default p_error = 2^-40 instead of 0.01.


RomanBredehoft (approved):
thanks !

PR #108, opened 2023-07-28
chore: improve make check_links

RomanBredehoft:
We no longer have two `check_links` as the repository is now public (therefore, most links as well)

closes https://github.com/zama-ai/concrete-ml-internal/issues/3881

### DIFF
--- a/script/make_utils/check_internal_links.sh
+++ b/script/make_utils/check_internal_links.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+# We don't want links to the main branch, even if it's public. Instead, release branch should be
+# targeted
+if grep -r "tree/main" docs | grep "\.md:" | grep -v "https://huggingface.co/spaces/"; then
+    echo -n -e "\nThe above links contain references to the main banch. Please remove them as only "
+    echo "release branches should be referenced."
+    exit 255
+fi
+
+
+# We don't want links to our internal repositories (COncrete ML or Concrete), expect if they are
+# GitHub issues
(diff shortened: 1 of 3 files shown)


andrei-stoian-zama (dismissed):

script/make_utils/check_internal_links.sh
```diff
@@ -0,0 +1,31 @@
...
+# We don't want links to the main branch, even if it's public. Instead, release branch should be
+# targeted
+if grep -r "tree/main" docs | grep "\.md:" | grep -v "https://huggingface.co/spaces/"; then
+    echo -n -e "\nThe above links contain references to the main banch. Please remove them as only "
+    echo "release branches should be referenced."
+    exit 255
+fi
+
+
+# We don't want links to our internal repositories (COncrete ML or Concrete), expect if they are
```
fd0r, line 12: Small typo "COncrete ML" instead of "Concrete ML"

fd0r (dismissed):
Small typo but otherwise I'm fine with the change.
Basically we don't have a check for gitbook links anymore (not sure why?)

There is a new check for internal and main links.

script/make_utils/check_internal_links.sh
RomanBredehoft: indeed, thanks

RomanBredehoft:
>  Basically we don't have a check for gitbook links anymore (not sure why?)

yes it's still here, I've just put it in the check_internal_links file !

andrei-stoian-zama (approved):
I love it when we remove code!! It means things are getting simpler

Issue #96, opened 2023-07-16
ImportError: cannot import name 'enc_split' from 'utils' in LLM case.


Lance729:
## Summary

When I attempt to run the LLM case, I the prosed error is `ImportError: cannot import name 'enc_split' from 'utils' in LLM case`.
The error appears in file `quant_framework.py` 
and sequence that `from utils import enc_split, max_fhe_relu, simple_slice`

In detail, the error is :
> ImportError                               Traceback (most recent call last)
> Cell In[1], line 5
>       3 import numpy as np
>       4 from concrete.fhe.tracing import Tracer
> ----> 5 from quant_framework import DualArray, Quantizer
>       7 from concrete import fhe
> 
> File /test_for_docker_jupyter/quant_framework.py:7
>       5 import numpy as np
>       6 from concrete.fhe.tracing import Tracer
> ----> 7 from utils import enc_split, max_fhe_relu, simple_slice
>       9 EPSILON = 2**-11
>      11 import numpy as np
> 
> ImportError: cannot import name 'enc_split' from 'utils' (/opt/conda/envs/py39/lib/python3.9/site-packages/utils/__init__.py)

## Description

- versions affected: concrete-ml 1.1
- python version: 3.9
- config (optional: HW, OS): Docker, Linux container

Lance729:
I fix it according to your suggest, danke!

Lance729:
I fix it according to your suggest.
Danke!

PR #111, opened 2023-07-31
chore: avoid utils.py files in use case examples

RomanBredehoft:
This should fix import errors some users faced 

closes https://github.com/zama-ai/concrete-ml-internal/issues/3854
refs https://github.com/zama-ai/concrete-ml/issues/96


andrei-stoian-zama (commented):
Can we have a make pcc thing that prevents utils.py files on our source tree ? 

It can be done in a second PR if you want

RomanBredehoft:
> fine with me even if not sure what's the point

@fd0r we've recently had users that encountered issues with the utils.py files not being recognized as a local file even in a virtual env (for example: https://github.com/zama-ai/concrete-ml/issues/96)

fd0r:
hummmm do we have an explanation for this?
looks weird to me

andrei-stoian-zama (approved):
Great!

PR #113, opened 2023-07-31
chore: add workflow to sync public and private repositories.

fd0r:
This commit adds a workflow that force pushes the public repository state to the private repository on pushes to main.


.github/workflows/sync_on_push.yaml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce
+        with:
+          name: repo-archive
+          path: '.'
+      - name: git-sync
+        uses: wei/git-sync@55c6b63b4f21607da0e9877ca9b4d11a29fc6d83
+        with:
+          source_repo: "zama-ai/concrete-ml"
+          source_branch: "main"
+          destination_repo: "https://${{ secrets.BOT_USERNAME }}:${{ secrets.CONCRETE_ACTIONS_TOKEN }}@github.com/${{ secrets.SYNC_DEST_REPO }}"
```
RomanBredehoft, line 29: so basically, this is targeting concrete-ml-internal right ?

.github/workflows/sync_on_push.yaml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  sync-repo:
+    if: ${{ github.repository == 'zama-ai/concrete-ml' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
+        with:
+          fetch-depth: 0
+      - name: Save repo
```
RomanBredehoft, line 19: what's the use of save repo ?

RomanBredehoft (approved):
looks good !

.github/workflows/sync_on_push.yaml
fd0r: yes

.github/workflows/sync_on_push.yaml
fd0r: My guess is that it's just a safeguard.
I copy-pasted it from @IceTDrinker 's workflow so maybe he can confirm.

IceTDrinker: I preferred doing a snapshot of the repo as a safeguard as we are doing some force pushes in there

PR #115, opened 2023-08-01
chore: disable CI if internal and push to main

fd0r:
Now that we have a workflow to force pushes public main to internal main we want to avoid running the CI twice. We can skip it directly in the workflow.


RomanBredehoft (approved):
thanks !

PR #116, opened 2023-08-01
chore: skip steps in regular CIs if possible

RomanBredehoft:
Basically, this will make regular CIs (from PRs) skip unnecessary checks (like running pytest for documentation-only PRs) !

refs https://github.com/zama-ai/concrete-ml-internal/issues/3893
closes https://github.com/zama-ai/concrete-ml-internal/issues/2870

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -466,33 +466,31 @@ jobs:
             tests:
               - tests/**
               - src/concrete/ml/pytest/**
-              - conftest.py
-              - deps_licenses/licenses_linux_user.txt.md5
             determinism:
-              - src/**
               - tests/seeding/test_seeding.py
-              - conftest.py
-              - deps_licenses/licenses_linux_user.txt.md5
             docs:
               - docs/**
             codeblocks:
-              - src/**
               - '**.md'
               - '!.*/**'
               - '!docs/_*/**'
-              - deps_licenses/licenses_linux_user.txt.md5
             dependencies:
               - deps_licenses/licenses_linux_user.txt.md5
+            conftest:
+              - conftest.py
 
       # Run determinism test. The only case where this is not run is for a PR that does not modify
       # anything in the source code or the determinism test file. This step is also triggered
-      # if any dependency has been updated 
+      # if any dependency has been updated or the conftest.py file has changed
       - name: Determinism
         id: determinism
         if: |
           !(
             steps.changed-files-in-pr.outcome == 'success' 
             && steps.changed-files-in-pr.outputs.determinism_any_changed == 'false'
+            && steps.changed-files-in-pr.outputs.src_any_changed == 'false'
+            && steps.changed-files-in-pr.outputs.dependencies_any_changed == 'false'
+            && steps.changed-files-in-pr.outputs.conftest_any_changed == 'false'
           )
           && steps.install-deps.outcome == 'success' 
           && steps.make-pcc.outcome == 'success' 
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -433,15 +433,62 @@ jobs:
...
+        if: ${{ env.IS_PR && steps.conformance.outcome == 'success' && !cancelled() }}
+        uses: tj-actions/changed-files@v37
+        with:
+          files_yaml: |
+            src:
+              - src/**
+              - deps_licenses/licenses_linux_user.txt.md5
+            tests:
+              - tests/**
+              - deps_licenses/licenses_linux_user.txt.md5
```
bcm-at-zama, line 452: strictly speaking, I would add `src/concrete/ml/pytest` in `tests`

RomanBredehoft: ah yes indeed !

.github/workflows/continuous-integration.yaml
```diff
@@ -427,21 +427,81 @@ jobs:
           make actionlint
 
       - name: Source code conformance
-        id: cs
+        id: make-conformance
```
fd0r, line 430: This is actually make pcc but alright 😉

.github/workflows/continuous-integration.yaml
```diff
@@ -427,21 +427,81 @@ jobs:
...
+            docs:
+              - docs/**
+            codeblocks:
+              - src/**
+              - '**.md'
+              - '!.*/**'
+              - '!docs/_*/**'
+              - deps_licenses/licenses_linux_user.txt.md5
+            dependencies:
+              - deps_licenses/licenses_linux_user.txt.md5
```
fd0r, line 474: Maybe poetry.lock would be better no?

andrei-stoian-zama (commented):
Just a few things to explain please

fd0r (dismissed):
A small question about adding the lock file to tracked files.
Otherwise everything looks good.

The formatting of the conditions looks much better than before.

.github/workflows/continuous-integration.yaml
RomanBredehoft: mo because the md5 also takes into account the CP version that we install in the makefile), which poetry lock does not ! We thought with @bcm-at-zama that his was simpler and more robust ! But I'll add a comment to make this clear yes

.github/workflows/continuous-integration.yaml
```diff
@@ -427,21 +427,81 @@ jobs:
...
+              - tests/**
+              - src/concrete/ml/pytest/**
+              - conftest.py
+              - deps_licenses/licenses_linux_user.txt.md5
+            determinism:
+              - src/**
+              - tests/seeding/test_seeding.py
+              - conftest.py
+              - deps_licenses/licenses_linux_user.txt.md5
+            docs:
```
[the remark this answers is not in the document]
RomanBredehoft: not sure I understand the question ? I use it in the build-docs step !

.github/workflows/continuous-integration.yaml
```diff
@@ -427,21 +427,81 @@ jobs:
...
+              - docs/**
+            codeblocks:
+              - src/**
```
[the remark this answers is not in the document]
RomanBredehoft: so this was a bit trickier : codeblocks are found directly in the `script/make_utils/pytest_codeblocks.sh` script, using `MD_FILES=$(find . -type f -name "*.md" | grep -v "^\./\." | grep -v "api/concrete\.ml")`

my initial idea was maybe to have a previous step that does this grep in the CI, and then give it to the action + the `make pytest_codeblock` later, but I'm unsure that I'm able to give such a variable to the action like this, and thus I preferred to manually do it here : 
- `src`: because if source changed, codeblocks should be tested again
- `**.md`: because that's what we do in the above find / grep
- `!.*/**` and `!docs/_*/**` sort of the same as in the find, we don't want to consider md from hidden directories or stuf like `docs/_apidoc`
- `deps_licenses/licenses_linux_user.txt.md5` same as most, if some dependencies were changed, we need to test codeblocks

but I'll add a comment to make this more clear

andrei-stoian-zama (approved):

PR #125, opened 2023-08-03
chore: test that changing docs notebook triggers right CI steps

RomanBredehoft:
just a PR testing the new CI skip feature, will be deleted once everything works well

refs https://github.com/zama-ai/concrete-ml-internal/issues/2870


RomanBredehoft:
looks good !

PR #129, opened 2023-08-03
chore: add pytest_no_flaky command and flag flaky tests

RomanBredehoft:
closes https://github.com/zama-ai/concrete-ml-internal/issues/3891

cc @bcm-at-zama @youben11

### DIFF
--- a/tests/parameter_search/test_p_error_binary_search.py
+++ b/tests/parameter_search/test_p_error_binary_search.py
@@ -30,6 +30,10 @@ from concrete.ml.pytest.utils import (
 from concrete.ml.search_parameters import BinarySearch
 from concrete.ml.sklearn import get_sklearn_linear_models
 
+# This test module is a known flaky
+# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3812
+pytestmark = pytest.mark.flaky
+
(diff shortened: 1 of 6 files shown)


andrei-stoian-zama (approved):

tests/parameter_search/test_p_error_binary_search.py
```diff
@@ -30,6 +30,10 @@
 from concrete.ml.search_parameters import BinarySearch
 from concrete.ml.sklearn import get_sklearn_linear_models
 
+# This test module is a known flaky
+# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3812
+pytestmark = pytest.mark.flaky
```
fd0r, line 35: Didn't know this was the way to mark the whole file.
Maybe a comment could be good to understand why we have that here

fd0r (approved):

tests/parameter_search/test_p_error_binary_search.py
RomanBredehoft: yes it's from their doc :  https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules

I'll add the comment on a following PR so that we merge this one right now

PR #133, opened 2023-08-04
chore: move collection of flaky tests out of pytest_no_flaky

RomanBredehoft:
Collecting tests was enough for triggering the parameter search flaky error, so I moved it out of `pytest_no_flaky`


bcm-at-zama (approved):
Looks good to me

PR #135, opened 2023-08-07
chore: remove flaky flag and skip test

fd0r:
\#3389 was previously flaky but updates in Concrete fixed the issue. We can now safely remove the skip and the flaky flag.

closes https://github.com/zama-ai/concrete-ml-internal/issues/3389


RomanBredehoft (approved):
great thanks !

PR #138, opened 2023-08-07
chore: re-run failed flaky tests in CI

RomanBredehoft:
Add a step in regular CIs that is triggered when `make pytest` fails. In this case, if failed tests are known flaky tests, it tries to re-run them (at most 3 times). If they finally pass, the CI adds a warning comment in the PR, listing the initial failed tests, and then continues on to run. However, if at least one failed test is not a know flaky tests, or if the flaky tests failed 3 times in a row, the CI becomes red as usual (and a simple comment is added in the PR).

I also added the [nick-fields/retry@v2.8.3](https://github.com/nick-fields/retry) action as a trusted action in the repository

closes https://github.com/zama-ai/concrete-ml-internal/issues/3888


RomanBredehoft:
all is good, closing the PR

PR #144, opened 2023-08-08
fix: with the new VL, this workaround is no more needed

bcm-at-zama:
linear models are well simulated

closes #https://github.com/zama-ai/concrete-ml-internal/issues/3906


tests/sklearn/test_sklearn_models.py
```diff
@@ -1531,7 +1532,7 @@ def check_for_divergent_predictions(x, model, fhe, max_iterations=N_ALLOWED_FHE_
     fhe_diff_found = check_for_divergent_predictions(x, model, fhe="execute")
 
     # Check for differences in predictions
-    if is_linear_model:
+    if is_linear_model and USE_OLD_VL:
```
RomanBredehoft, line 1535: I've read the initial issue:  so you are saying that the new VL now properly simulates fully-leveled circuits with hgh p-error ? 

and I think it was not indicated as a workaround / fixme since we didn't know this was something that was going to be fixed one day !

bcm-at-zama: at least it's what I see, and yes I think it's expected it's fixed at some points!

RomanBredehoft: maybe we should make sure that this is expected, but if it then yes thanks !

bcm-at-zama: I don't understand why fixing bugs would not be expected! But I can ask if you prefer

bcm-at-zama: @BourgerieQuentin , is it expected that, with the new VL, we simulate better with programs with "large p_error"? With the old VL, for linear models, it was not well simulated.

RomanBredehoft: more specifically, it seems that `p-error` was not simulated with fully-leveled circuits using the old VL, but the new does

BourgerieQuentin: Let's change the terminology don't use VL for the new one but simulation, I guess it give a better idea of what is done. We don't try to mimic the behavior of TFHE we actually run the FHE evaluation but without the mask, that all. That means that the simulation should act exactly same as the actual FHE evaluation (as well if no bugs), so yes it's expected than fully-levelled circuit are actually simulated.

RomanBredehoft: ok great thanks !

RomanBredehoft: @bcm-at-zama then maybe add a small comment here to explain why this test only happens with "old_vl" ?

bcm-at-zama: Sure, done

RomanBredehoft (approved):
thanks a lot !

fd0r (approved):
Fine with me

PR #145, opened 2023-08-09
chore: udpate cifar 16-bit code to run using multi-param

fd0r:


.github/workflows/cifar_benchmark.yaml
```diff
@@ -132,6 +132,7 @@ jobs:
           python3 ./benchmarks/convert_cifar.py --model-name "16-bits-trained-v0"
 
       - name: Upload results
+        if: ${{ github.repository == 'zama-ai/concrete-ml-internal' }}
```
RomanBredehoft, line 135: why should we only upload results from the private repo ?

fd0r: The public repository doesn't have the secret needed to updload and we don't use this DB anymore.

If we don't skip this in the public repository this fails the action.

RomanBredehoft (approved):
thanks ! I guess updating the benchmarks with the DB will be done along https://github.com/zama-ai/concrete-ml-internal/issues/3533

PR #152, opened 2023-08-10
chore: re-run failed flaky tests in CI

RomanBredehoft:
This PR improves regular CIs (non-weekly, non-release) in order to tackle the flaky issue problem. 

Essentially, until now we were running pytest but then the CI would be stopped if at least one test failed. This is an issue as the CI currently takes around 1 hour but many failures come known flaky tests (tests that fail occasionally, waiting to be fixed).

What I suggest in this PR is : 
- run pytest as before and build a full report (using pytest-json plugin)
- if pytest failed, use the report to check if all failed tests are known flaky tests (meaning we already tagged them as flaky)
- if this is the case, then call pytest once again but only on the failed flaky tests
- if the re-run succeeded, a warning comment is added to the on-going PR (example of such a comment : https://github.com/zama-ai/concrete-ml/pull/198#issuecomment-1695740532). It notably list the failed flaky tests that were re-run
- in any other cases (including if the re-run fails), the CI fails as before and no comment is added

Besides, we could also add comment whenever pytest fails (with flaky tests or not, without re-running anything) if we think it's useful. For now I haven't added it since it might make PRs a bit too noisy. We'll see

closes https://github.com/zama-ai/concrete-ml-internal/issues/3888


fd0r:
What are we waiting for here?

fd0r:
> that it works 😅

Alright I would be expecting the PR to be a draft one until ready 😉

fd0r (approved):
That's great, thanks a lot!

PR #158, opened 2023-08-21
chore: update concrete-python to 2.2.0

fd0r:


fd0r:
3rd time I get the same flaky, what do you think we should do here @RomanBredehoft ?

tests/torch/test_hybrid_converter.py
```diff
@@ -53,12 +53,13 @@ def run_hybrid_model_test(
 
     # Compute average of these counts (the accuracy)
     accuracy = is_in.float().mean()
+    del accuracy
```
RomanBredehoft, line 56: why this del ? 😅

fd0r: Just to avoid my linter from saying that the variable was unused.

RomanBredehoft: ah yes ok got it

fd0r: actually not enough to have pcc happy

RomanBredehoft (approved):
thanks !

PR #176, opened 2023-08-23
chore: remove wrong Concrete url from docstring

RomanBredehoft:
This URL does not exist anymore (preventing us from building the documentation)

### DIFF
--- a/src/concrete/ml/sklearn/base.py
+++ b/src/concrete/ml/sklearn/base.py
@@ -242,8 +242,8 @@ class BaseEstimator:
         """Get the FHE circuit.
 
         The FHE circuit combines computational graph, mlir, client and server into a single object.
-        More information available in Concrete documentation:
-        https://docs.zama.ai/concrete/developer/terminology_and_structure#terminology
+        More information available in Concrete documentation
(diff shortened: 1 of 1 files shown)


src/concrete/ml/sklearn/base.py
```diff
@@ -242,8 +242,7 @@ def fhe_circuit(self) -> Optional[Circuit]:
         """Get the FHE circuit.
 
         The FHE circuit combines computational graph, mlir, client and server into a single object.
-        More information available in Concrete documentation:
-        https://docs.zama.ai/concrete/developer/terminology_and_structure#terminology
+        More information available in Concrete documentation (https://docs.zama.ai/concrete)
```
fd0r, line 245: Maybe https://docs.zama.ai/concrete/getting-started/terminology_and_structure would be better?

fd0r (changes requested):

src/concrete/ml/sklearn/base.py
RomanBredehoft: ah yes sure

fd0r (approved):

PR #177, opened 2023-08-23
chore: avoid triggering codeblock tests if SUMMARY.md changed

RomanBredehoft:
the docs/SUMMARY.md only has links but no codeblocks, but is changed when updating apidocs (which makes the prepare release CI trigger unnecessary codeblock tests)

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -488,6 +488,7 @@ jobs:
               - '**.md'
               - '!.*/**'
               - '!docs/_*/**'
+              - '!docs/SUMMARY.md'
             dependencies:
               - deps_licenses/licenses_linux_user.txt.md5
             conftest:


.github/workflows/continuous-integration.yaml
```diff
@@ -488,6 +488,7 @@ jobs:
               - '**.md'
               - '!.*/**'
               - '!docs/_*/**'
+              - 'docs/SUMMARY.md'
```
fd0r, line 491: Missing !

RomanBredehoft: indeed ...

fd0r:
Would be great to have something that checks for the markdowns with code-blocks and trigger the CI only if some code-block was modified or added.

RomanBredehoft:
yes indeed : https://github.com/zama-ai/concrete-ml-internal/issues/3929

PR #185, opened 2023-08-24
chore: add flaky mark to double_fit test

RomanBredehoft:
refs https://github.com/zama-ai/concrete-ml-internal/issues/3932

### DIFF
--- a/tests/sklearn/test_sklearn_models.py
+++ b/tests/sklearn/test_sklearn_models.py
@@ -1230,6 +1230,9 @@ def test_serialization(
     check_serialization(model, x, use_dump_method)
 
 
+# This test is a known flaky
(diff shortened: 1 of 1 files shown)


tests/sklearn/test_sklearn_models.py
```diff
@@ -1230,6 +1230,9 @@ def test_serialization(
     check_serialization(model, x, use_dump_method)
 
 
+# This test is a known flaky
+# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3932
+@pytest.mark.flaky
```
fd0r, line 1235: A bit weird to flag the whole test as flaky when only one configuration of this if flaky but don't know if there is a better way 🤔

fd0r (approved):

PR #200, opened 2023-08-29
fix: flaky test double_fit

RomanBredehoft:
Basically the flaky appeared because we were very unlucky : we were generating 2 datasets sequentially (with fixed seeds) using a new `random_state` each time. This `random_state` was generated through `random_state = numpy.random.randint(0, 2**15)`. Turns out, for seed `1687196373446039357`, two consecutive `randint` calls were outputting the same number (`26567`), which led `get_dataset` / `load_data` to generate the same dataset twice (and thus making the model's predictions identical in the test, while we expect a difference)

I therefore suggest to generate a random state manually once for generating the first dataset and then add 1 to it for generating the second dataset (so that we make sure both random states are different). I believe this should be enough but feel free to suggest other ideas

closes https://github.com/zama-ai/concrete-ml-internal/issues/3932

### DIFF
--- a/tests/sklearn/test_sklearn_models.py
+++ b/tests/sklearn/test_sklearn_models.py
@@ -1230,9 +1230,6 @@ def test_serialization(
     check_serialization(model, x, use_dump_method)
 
 
-# This test is a known flaky
-# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3932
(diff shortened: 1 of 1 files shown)


andrei-stoian-zama (approved):
Nice debugging here!

PR #217, opened 2023-09-01
Knn classifier in cml

kcelia:
As of now:
- only predict function is supported, predict_proba is not.
- what's done on the client side?  the majority_vote.
- what's done on the server side? pairwise_euclidean_distance and the k nearest labels.
   - sqrt is not performed in FHE, since sqrt is a  monotonic function, it doesn't affect the argmax calculation. So, removing it will increase the computation,

closes [#3818](https://github.com/zama-ai/concrete-ml-internal/issues/3818)


src/concrete/ml/sklearn/neighbors.py
```diff
@@ -0,0 +1,145 @@
...
+        obj.weights = metadata["weights"]
+        obj.algorithm = metadata["algorithm"]
+        obj.p = metadata["p"]
+        obj.metric = metadata["metric"]
+        obj.metric_params = metadata["metric_params"]
+        obj.n_jobs = metadata["n_jobs"]
+        return obj
+
+
+class _KNeighborsRegressor:
```
andrei-stoian-zama, line 119: you can remove these next classes

andrei-stoian-zama (changes requested):

src/concrete/ml/sklearn/neighbors.py
kcelia: alright

jfrery (commented):
Great first draft well done! Could you explain the choices about what is done in FHE and what is done in clear on the client side?

It looks like we only partly compute the distances in FHE. Why isn't the sqrt done in FHE? Is it just prohibitively expensive to compute the topk and majority vote in FHE?

Euclidean distance might not be the most effective distance to compute in FHE due to the square (ciphertext multiplication) and then the sqrt. Would something like the manhattan distance be better suited to start with?

src/concrete/ml/sklearn/base.py
```diff
@@ -1688,3 +1692,264 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+            numpy.ndarray: The quantized predicted values.
+        """
+        assert self._q_X_fit_quantizer is not None, self._is_not_fitted_error_message()
+
+        # Pairwise euclidean distance
+        # dist(x, y) = sqrt(dot(x, x) - 2 * dot(x, y) + dot(y, y))
+        distance_matrix = (
+            numpy.sum(q_X**2, axis=1, keepdims=True)
+            - 2 * q_X @ self._q_X_fit.T
+            + numpy.expand_dims(numpy.sum(self._q_X_fit**2, axis=1), 0)
```
jfrery, line 1923: This can be done at training time I suppose

fd0r:
Seems like this PR adds a lot of time to the CI.

src/concrete/ml/sklearn/base.py
```diff
@@ -1913,29 +1915,97 @@ def _inference(self, q_X: numpy.ndarray) -> numpy.ndarray:
...
+                return enc_arr
+
+            def scatter1d(x, v, indices):
+                for idx, i in enumerate(indices):
+                    x[i] = v[idx]
+                return x
+
+            def mul_tlu(a, b):
+                # (a - b)^2 - (a + b)^2 = -4ab => ab = ((a + b)^2 - (a - b)^2) / 4
+                return (((a + b) ** 2 - (a - b) ** 2) / 4).astype(numpy.int64)
```
jfrery, line 1948: We can do the a * b instead as it's supported by concrete.

src/concrete/ml/sklearn/base.py
```diff
@@ -1688,3 +1692,332 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+                - 2 * q_X @ self._q_X_fit.T
+                + numpy.expand_dims(numpy.sum(self._q_X_fit**2, axis=1), 0)
+            )
+
+        distance_matrix = pairwise_euclidean_distance(q_X)
+
+        # sqr not done
+
+        # 2. Sorting
+        def topk_sorting(x):
```
jfrery, line 1932: Can we ask concrete to implement argsort? cc @bcm-at-zama @andrei-stoian-zama

bcm-at-zama: if it is doable in FHE and if it is a priority: yes, it can be in the possible future-KR list for Q4 and later

tests/common/test_skearn_model_lists.py
```diff
@@ -69,12 +76,18 @@ def test_get_sklearn_models():
         Ridge,
         TweedieRegressor,
     ]
+
+    expected_neighbors_list = [KNeighborsClassifier]
```
bcm-at-zama, line 80: excellent! all generic tests work directly for KNN, without changes on the test side? there is nothing really special in the KNN, as compared to the previously existing sklearn models, in terms of how / what we test?

bcm-at-zama: lol my bad, it's tests/common/test_skearn_model_lists.py, it's not the generic test file

src/concrete/ml/pytest/utils.py
```diff
@@ -95,9 +97,25 @@
         id=get_model_name(model),
     )
     for model in _classifier_models
+    if get_model_name(model) != "KNeighborsClassifier"
     for n_classes in [2, 4]
+] + [
+    pytest.param(
```
andrei-stoian-zama, line 103: so, in all tests we check the KNN on a 3-dimensional, 10-sample dataset with 2 classes, is that correct ? 

Is 10 examples not too big for the CI ? We could do 2 examples no ?

src/concrete/ml/search_parameters/p_error_search.py
```diff
@@ -104,6 +110,16 @@ def compile_and_simulated_fhe_inference(
     """
 
     compile_params: Dict = {}
+
+    default_configuration = Configuration(
+        dump_artifacts_on_unexpected_failures=False,
+        enable_unsafe_features=True,
+        use_insecure_key_cache=True,
+        insecure_key_cache_location="ConcreteNumpyKeyCache",
+        parameter_selection_strategy=ParameterSelectionStrategy.MONO
```
andrei-stoian-zama, line 119: we can merge this way and wait for the new CP that fixes this but we should add a TODO comment and an issue

src/concrete/ml/sklearn/base.py
```diff
@@ -657,7 +657,10 @@ def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.
                 # Execute the inference in FHE or with simulation
                 q_y_pred_i = predict_method(q_X_i)
 
-                q_y_pred_list.append(q_y_pred_i[0])
+                if self.__class__.__name__ == "KNeighborsClassifier":
+                    q_y_pred_list.append(q_y_pred_i)
+                else:
+                    q_y_pred_list.append(q_y_pred_i[0])
```
RomanBredehoft, line 663: this does not look right, the output of a circuit is a tuple of one element (since we only pass batches of 1 as inputs), this is why we have this `[0]`. I can't see why a specific model should do that differently

tests/deployment/test_client_server.py
```diff
@@ -95,10 +98,21 @@ def test_client_server_sklearn(
...
+            insecure_key_cache_location="ConcreteNumpyKeyCache",
+            parameter_selection_strategy=fhe.ParameterSelectionStrategy.MONO,
+            single_precision=True,
+        )
+
     # Running the simulation using a model that is not compiled should not be possible
     with pytest.raises(AttributeError, match=".* model is not compiled.*"):
         client_server_simulation(x_train, x_test, model, default_configuration)
-
+    # With n_bits = 3, KNN is not compilable
```
andrei-stoian-zama, line 115: but don't we want to leave the n_bits setting as it was for the other models ? Is it doable ? 

if it doesn't matter that we reduce n_bits to 2 for all modesl and it's faster this way (are we doing FHE computation herE), then we can set it to 2 all the time and gain some CI time

andrei-stoian-zama (changes requested):

src/concrete/ml/sklearn/base.py
RomanBredehoft: @kcelia

src/concrete/ml/search_parameters/p_error_search.py
kcelia: I created an issue to remove the mono execution: https://github.com/zama-ai/concrete-ml-internal/issues/3978
I also changed the default_configuration in the conftest.py, to make it more clear.

tests/deployment/test_client_server.py
kcelia: Yes, we are doing FHE computation here, but it doesn't impact the other models and we gain mere seconds in the CI.
So, I thought we can keep 2 for all models.

src/concrete/ml/sklearn/base.py
kcelia: I'm addressing all your comments.
I said in https://zama-ai.slack.com/archives/C0455QDGT6C/p1694684283172319?thread_ts=1694590559.480319&cid=C0455QDGT6C, I agree with you.
I was a workaround.

src/concrete/ml/search_parameters/p_error_search.py
RomanBredehoft: thanks for the issue (maybe could you explain a bit more what we would need to do in order to fix the issue ? I mean in the issue), as Andrei said could you put a FIXME here so that we know that there's actually an issue here ?

src/concrete/ml/sklearn/base.py
RomanBredehoft: ah yes I'm not following this discussion ! thanks

conftest.py
```diff
@@ -146,19 +147,32 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus):  # pylint: disabl
 def default_configuration():
     """Return the default test compilation configuration."""
 
-    # Remove parameter_selection_strategy once it is set to multi-parameter in Concrete Python
```
RomanBredehoft, line 149: please don't remove these comments as they are pretty important to track down known issues

conftest.py
```diff
@@ -146,19 +147,32 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus):  # pylint: disabl
...
+    def default_configuration_impl(mode: Optional[str] = "MULTI"):
+        """Return the default test compilation configuration for a given strategy.
+
+        Args:
+            mode (Optional[str]): Whether we execute in a MONO or MULTI precision settings.
+
+        Returns:
+            The configuration.
+        """
+        if mode == "MULTI":
```
RomanBredehoft, line 159: really not a fan of this as it brings many changes in tests that I don't thing they are necessary. If it is absolutely necessary, please explain why and add a link to an issue here !

conftest.py
```diff
@@ -492,8 +511,12 @@ def check_is_good_execution_for_cml_vs_circuit_impl(
                     # tests), especially since these results are tested in other tests such as the
                     # `check_subfunctions_in_fhe`
                     if is_classifier_or_partial_classifier(model):
-                        results_cnp_circuit = model.predict_proba(*inputs, fhe=fhe_mode)
-                        results_model = model.predict_proba(*inputs, fhe="disable")
+                        if isinstance(model, SklearnKNeighborsMixin):
```
RomanBredehoft, line 514: could you link an issue here if KNN should support predict-proba at one point !

RomanBredehoft (changes requested):
Many changes in our tests related to this "defaut_configuration" fixture and the fact that KNN does not seem to support MULTI strategy yet. I think you should be able to avoid these changes by simply force MONO in their compile method as we do with linear models ! Also, don't forget to link issues when there are some, as we might forget to track them properly later on

src/concrete/ml/search_parameters/p_error_search.py
RomanBredehoft: also I think this change is not necessary (at least for the strategy part) if you directly force KNN to MONO strategy in their compilation step as done for linear models

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1697,358 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+                    - op_inputs : number of bits to quantize the input values
+                    - op_weights: number of bits to quantize the learned parameters
+                Default to 3.
+        """
+        self.n_bits: Union[int, Dict[str, int]] = n_bits
+
+        #: The quantizer to use for quantizing the model's weights
+        self._weight_quantizer: Optional[UniformQuantizer] = None
+        self._q_X_fit_quantizer: Optional[UniformQuantizer] = None
+        self._q_X_fit: numpy.ndarray
```
RomanBredehoft, line 1734: could you add a small comment explaining what are these attributes / why are they necessary ?

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1697,358 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+                    diff = a - b
+                    sign = diff < 0
+
+                    max_x = a + numpy.maximum(0, b - a)
+                    x = scatter1d(x, a + b - max_x, range_i)  # x[range_i] = a + b - max_x
+                    x = scatter1d(x, max_x, range_i + d)  # x[range_i + d] = max_x
+
+                    max_idx = a_i + mul_tlu((b_i - a_i), sign)
+
+                    # idx[range_i] = a_i + b_i - max_idx
```
RomanBredehoft, line 2001: is this comment supposed to be removed or is it supposed to be next to the following line ?

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1697,358 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+            x = fhe_array(x)
+
+            assert x.shape[0] == self.n_neighbors
+
+            return x
+
+        # 1. Pairwise_euclidiean distance
+        distance_matrix = pairwise_euclidean_distance(q_X)
+
+        # sqr not done
```
RomanBredehoft, line 2022: what is this comment ?

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1697,358 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+
+        X = check_array_and_assert(X)
+
+        y_preds = []
+        for query in X:
+            # Argsort
+            arg_sort = super().predict(query[None], fhe)
+            assert arg_sort.size == self.n_neighbors
+            # Majority vote
+            # pylint: disable=protected-access
```
RomanBredehoft, line 2039: `disable-next` instead of `disable`, else I think it disables for the whole file !

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1697,358 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+        # Check that the underlying sklearn model has been set and fit
+        assert self.sklearn_model is not None, self._sklearn_model_is_not_fitted_error_message()
+
+        self.onnx_model_ = hb_convert(
+            self.sklearn_model,
+            backend="onnx",
+            test_input=test_input,
+            extra_config={
+                "onnx_target_opset": OPSET_VERSION_FOR_ONNX_EXPORT,
+                # pylint: disable=protected-access, no-member
```
RomanBredehoft, line 1753: `disable-next`, same reasons !

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1697,358 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
         y_logits = self.decision_function(X, fhe=fhe)
         y_proba = self.post_processing(y_logits)
         return y_proba
+
+
+# pylint: disable=invalid-name,too-many-instance-attributes
```
RomanBredehoft, line 1702: `disable-next` is better, else it disables for the whole file ! If it does not work, then it's better to re-enable it after the class' definition !

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -0,0 +1,119 @@
+"""Implement sklearn linear model."""
+from typing import Any, Dict
+
+import sklearn.linear_model
+
+from .base import SklearnKNeighborsClassifierMixin
+
+
+# pylint: disable=invalid-name,too-many-instance-attributes
```
RomanBredehoft, line 9: `disable-next`, same reasons !

conftest.py
kcelia: Correct ! I removed them by mistake

conftest.py
kcelia: Great catch, I forgot to add it, here.

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: I think the comment explains the line by giving an alternative explanation. It could be re-written as :+1: 

`# The following is equivalent to:  idx[range_i] = a_i + b_i - max_idx`

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1697,358 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+                Args:
+                    a (numpy.ndarray): An encrypted array
+                    b (numpy.ndarray): An encrypted array
+
+                Returns:
+                    numpy.ndarray: The result of a * b
+                """
+                return a * b
+
+            comparisons = numpy.zeros(x.shape)
```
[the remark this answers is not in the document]
jfrery: Agree with Roman here. Maybe creating more function could help? It's pretty hard to follow

src/concrete/ml/sklearn/base.py
kcelia: fixed

conftest.py
kcelia: Fixed

src/concrete/ml/search_parameters/p_error_search.py
kcelia: Following your advice, I forced the KNN to run in MONO strategy

src/concrete/ml/sklearn/base.py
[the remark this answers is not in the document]
kcelia: I added some comments, let me know if it's more clear.

src/concrete/ml/sklearn/base.py
kcelia: with KNN, you don’t need a training phase at all, we do not talk a bout weights. Instead we we keep in memory the train base to calculate distances and similarities, commonly called _X_fit (from scikit learn).

Added a comment

conftest.py
RomanBredehoft: this was resolved but I don't see any issue here !

src/concrete/ml/sklearn/base.py
RomanBredehoft: yes I meant in the code so thanks ! however, it's better to write comments above each attributes starting with "x:" (for doc visualization) , for both `_q_X_fit_quantizer` and `_q_X_fit`

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1695,388 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+        self.output_quant_params = UniformQuantizationParameters(
+            scale=self._q_X_fit_quantizer.scale,
+            zero_point=self._q_X_fit_quantizer.zero_point,
+            offset=0,
+        )
+
+        output_quantizer = UniformQuantizer(params=self.output_quant_params, no_clipping=True)
+
+        # Since the matmul and the bias both use the same scale and zero-points, we obtain that
+        # y = S*(q_y - 2*Z) when de-quantizing the values. We therefore need to multiply the initial
```
RomanBredehoft, line 1829: this comment is not relevant anymoe, there is not bias term here !

kcelia: You are right, I forgot to remove it.
Good catch, thanks

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1695,385 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+        assert q_X.dtype == numpy.int64, "Inputs were not quantized to int64 values"
+        return q_X
+
+    def dequantize_output(self, q_y_preds: numpy.ndarray) -> numpy.ndarray:
+        self.check_model_is_fitted()
+
+        # We compute the sorted argmax in FHE, which are integers.
+        # No need to de-quantize the output values
+
+        assert q_y_preds[0].shape[-1] == self.n_neighbors, (
```
RomanBredehoft, line 1851: better to check `q_y_preds.n_dim == self.n_neighbors` right ? with `q_y_preds[0].shape[-1]` , having `q_y_preds[0]` empty will fail with wrong message and a shape of `(1, 1, 1, self.n_neighbors)` will wrongly succeed

src/concrete/ml/sklearn/base.py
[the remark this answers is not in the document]
RomanBredehoft: great thanks, but I think you can still you could add a bit more, like what does the complete loop `for t in range(ln2n - 1, -1, -1):` does, same for `for bq in range(ln2n - 1, t - 1, -1):` ! I think it sill lacks a bit of more generic explanations ! Feel free to add references or similar things to make it clearer

tests/sklearn/test_sklearn_models.py
```diff
@@ -273,10 +279,18 @@ def check_double_fit(model_class, n_bits, x_1, x_2, y_1, y_2):
...
-            for (quantizer_1, quantizer_2) in zip(quantizers_1, quantizers_2)
-        )
+
+        # For now, in KNN, we compute the pairwise Euclidean distance between the encrypted
+        # X and each element in the database.
+        # Then, we return the indices of the k closest distances to this point.
+        # The exact precision of computation of the quantization and dequantization parameters
+        # is not relevant in this case. That's why the assertion test is being ignored
+        # for now in the context of the KNN algorithm.
+        if get_model_name(model) != "KNeighborsClassifier":
```
RomanBredehoft, line 289: even with the added comment, I'm not sure I get why two seeded fits of the same KNNs with the same dabatase could lead to different quantizers

also you mention "for now" in the comment, does that mean we'll need to do something in the future ? If so, could you create and add an issue here ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1406,6 +1455,9 @@ def test_predict_correctness(
             f"number_of_tests_in_non_fhe = {number_of_tests_in_non_fhe})"
         )
 
+    if n_bits > 5 and get_model_name(model) == "KNeighborsClassifier":
```
RomanBredehoft, line 1458: why is that ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -852,6 +899,8 @@ def check_fitted_compiled_error_raises(model_class, n_bits, x, y):
             model.predict(x)
 
     if is_classifier_or_partial_classifier(model_class):
+        if get_model_name(model) == "KNeighborsClassifier":
```
RomanBredehoft, line 902: you forgot to add a fixme here as well !

src/concrete/ml/sklearn/base.py
kcelia: This configuration is not possible because the input must be 2D.
q_y_preds is tuple of size 2, so q_y_preds.ndim returns 2.

RomanBredehoft: yes indeed what I said is wrong, but then also, your print message indicates that the predictions must have a shape of `({self.n_neighbors},)` so it's 1d or 2d ? and i any case are you sure it's always going to be 2d ? with neighbors the first dim  ? in any case maybe the assert is not useful enough (you could have let `dequantize_output` just return q_y_preds) !

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -0,0 +1,117 @@
...
+class KNeighborsClassifier(SklearnKNeighborsClassifierMixin):
+    """A k-nearest classifier model with FHE.
+
+    Parameters:
+        n_bits (int, Dict[str, int]): Number of bits to quantize the model. If an int is passed
+            for n_bits, the value will be used for quantizing inputs and weights. If a dict is
+            passed, then it should contain "op_inputs" and "op_weights" as keys with
+            corresponding number of quantization bits so that:
+            - op_inputs : number of bits to quantize the input values
+            - op_weights: number of bits to quantize the learned parameters
```
jfrery, line 19: I am not sure it's clear what you are referring to here with `op_weights` and `op_inputs`.

tests/sklearn/test_sklearn_models.py
jfrery: I am also not sure here. Here we expect that .fit() to change when we re-fit. Why wouldn't this be the case on KNN?

tests/sklearn/test_sklearn_models.py
```diff
@@ -663,9 +694,20 @@ def check_grid_search(model_class, x, y, scoring):
...
+
+        if get_model_name(model_class) == "KNeighborsClassifier" and scoring in [
+            "roc_auc",
+            "average_precision",
+        ]:
+            # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3962
+            pytest.skip("Skipping predict_proba for KNN, doesn't work for now")
+
+        # pylint: disable=invalid-name
+        cv = 2 if get_model_name(model_class) == "KNeighborsClassifier" else 5
```
jfrery, line 707: I think cv is not that important. If 5 is too large let's just pick 2 or 3. We just want gridsearch to work with our models. So you can probably remove this condition

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1695,385 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+    def __init__(self, n_bits: Union[int, Dict[str, int]] = 3):
+        """Initialize the FHE knn model.
+
+        Args:
+            n_bits (int, Dict[str, int]): Number of bits to quantize the model. If an int is passed
+                for n_bits, the value will be used for quantizing inputs and weights. If a dict is
+                passed, then it should contain "op_inputs" and "op_weights" as keys with
+                corresponding number of quantization bits so that:
+                    - op_inputs : number of bits to quantize the input values
+                    - op_weights: number of bits to quantize the learned parameters
```
jfrery, line 1724: Same here. It's not clear what these values refer to

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1695,385 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+    def _set_onnx_model(self, test_input: numpy.ndarray) -> None:
+        """Retrieve the model's ONNX graph using Hummingbird conversion.
+
+        Args:
+            test_input (numpy.ndarray): An input data used to trace the model execution.
+        """
+        # Check that the underlying sklearn model has been set and fit
+        assert self.sklearn_model is not None, self._sklearn_model_is_not_fitted_error_message()
+
+        self.onnx_model_ = hb_convert(
```
jfrery, line 1749: hmm do we use this? I think we coded everything from scratch so I don't get why we extract the onnx here.

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1695,385 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+        input_n_bits = n_bits["op_inputs"]
+        input_options = QuantizationOptions(n_bits=input_n_bits, is_signed=True)
+
+        # Quantize the inputs and store the associated quantizer
+        q_inputs = QuantizedArray(n_bits=input_n_bits, values=X, options=input_options)
+        input_quantizer = q_inputs.quantizer
+        self.input_quantizers.append(input_quantizer)
+
+        weights_n_bits = n_bits["op_weights"]
+        weight_options = QuantizationOptions(n_bits=weights_n_bits, is_signed=True)
```
jfrery, line 1799: The namings are confusing here. Why are we talking about weights?

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1695,385 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+            # 1. Pairwise euclidean distance
+            # dist(x, y) = sqrt(dot(x, x) - 2 * dot(x, y) + dot(y, y))
+            return (
+                numpy.sum(q_X**2, axis=1, keepdims=True)
+                - 2 * q_X @ self._q_X_fit.T
+                + numpy.expand_dims(numpy.sum(self._q_X_fit**2, axis=1), 0)
+            )
+
+        def topk_sorting(x):
+            """Argsort in FHE.
```
jfrery, line 1913: Can you add the complexity of this function here please?

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1695,385 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+
+        X = check_array_and_assert(X)
+
+        y_preds = []
+        for query in X:
+            # Argsort
+            arg_sort = super().predict(query[None], fhe)
+            # Majority vote
+            # pylint: disable-next=protected-access
+            label_indices = self._y[arg_sort.flatten()]
```
jfrery, line 2065: Are the training data point used for the fit shared with the client? If this self._y exists here I assume it's because it's on the client machine?

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(
+        self,
+        n_bits=2,
+        n_neighbors=3,
+        *,
+        weights="uniform",
+        algorithm="auto",
+        leaf_size=30,
+        p=2,
+        metric="minkowski",
```
andrei-stoian-zama, line 38: L2 is minkowski with p=2 right ? 

Shouldn't we assert if something else is given ?

jfrery (changes requested):
Thanks @kcelia ! I have made some comments. Looks like adding a model in the test is quite intense. Do you feel some things could be made simpler? If so feel free to open an issue.

tests/sklearn/test_sklearn_models.py
andrei-stoian-zama: I think you should treat KNN like trees: see the condition before that only takes the input quantizers for trees. You should do the same here: KNN has only input quantizers. Then you can keep the assert in all cases

andrei-stoian-zama (changes requested):
Looks good! Some minor changes please!

tests/sklearn/test_sklearn_models.py
kcelia: ok

src/concrete/ml/sklearn/base.py
kcelia: It's not big test, I added it when I had a mismatch dimension due to the extra dim with tuple.
You  even mentioned it in `this does not look right, the output of a circuit is a tuple of one element (since we only pass batches of 1 as inputs), this is why we have this [0]. I can't see why a specific model should do that differently`.

I'll remove it.

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -0,0 +1,117 @@
...
+        metric_params=None,
+        n_jobs=None,
+    ):
+        # Call SklearnKNeighborsClassifierMixin's __init__ method
+        super().__init__(n_bits=n_bits)
+
+        self.n_neighbors = n_neighbors
+        self.algorithm = algorithm
```
andrei-stoian-zama, line 46: also need to assert algorithm is "brute force" or "auto". the other ones (k-d tree, ball tree) can not be done in FHE. see https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html

src/concrete/ml/sklearn/neighbors.py
andrei-stoian-zama: also note we don't support metric= a callable function.

src/concrete/ml/sklearn/neighbors.py
kcelia: It was initially a copay past from the classifiermixin, I forgot to remove it. 
we don't talk about weights in KNN, but about X_fit
Thanks

src/concrete/ml/sklearn/base.py
kcelia: I'll close the comment according to this discussion: https://zama-ai.slack.com/archives/C0455QDGT6C/p1695131751281249

src/concrete/ml/sklearn/base.py
kcelia: And now ?

src/concrete/ml/sklearn/base.py
kcelia: I changed it to: 
``` 
        # The square root in the Euclidean distance calculation is not applied to speed up FHE
        # computations.
        # Being a monotonic function, it does not affect the logic of the calculation, notably for
        # the argsort.

```

conftest.py
kcelia: closed by mistake, it has been added.

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1694,371 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+    def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
+
+        X = check_array_and_assert(X)
+
+        y_preds = []
+        for query in X:
+            # Argsort
+            arg_sort = super().predict(query[None], fhe)
+            # Majority vote
+            # pylint: disable-next=protected-access
```
RomanBredehoft, line 2049: are you sure you need this pylint ? we are still in the class so I don't think it should complain that you are accessing `self._y` !

tests/sklearn/test_sklearn_models.py
```diff
@@ -626,8 +642,8 @@ def check_pipeline(model_class, x, y):
         param_grid = {
             "model__n_bits": [2, 3],
         }
-
-    grid_search = GridSearchCV(pipe_cv, param_grid, error_score="raise", cv=3)
+    # Since the data-set is really small for KNN, we have to decrease the number of splits
```
RomanBredehoft, line 645: this comment is not necessary I think, you could remove it (or instead say "we need a small number of splits")

RomanBredehoft (approved):
thanks a lot, looks good to me ! I only have two remaining small comments

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -0,0 +1,124 @@
...
+        n_jobs=None,
+    ):
+        # Call SklearnKNeighborsClassifierMixin's __init__ method
+        super().__init__(n_bits=n_bits)
+
+        assert algorithm in ["brute", "auto"], f"Algorithm = `{algorithm}` is supported in FHE."
+        assert not callable(metric), "`metric` should not be a callable object"
+        assert (
+            p == 2 and metric == "minkowski"
+        ), "Only `L2` norm is supported with `p=2` and `metric = 'minkowski'`"
```
andrei-stoian-zama, line 50: use assert_true please but the message here is good !

andrei-stoian-zama (dismissed):
Looks good but please switch to assert_true for the asserts you added. You can do it in a second PR if you want

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -0,0 +1,124 @@
...
+        p=2,
+        metric="minkowski",
+        metric_params=None,
+        n_jobs=None,
+    ):
+        # Call SklearnKNeighborsClassifierMixin's __init__ method
+        super().__init__(n_bits=n_bits)
+
+        assert algorithm in ["brute", "auto"], f"Algorithm = `{algorithm}` is supported in FHE."
+        assert not callable(metric), "`metric` should not be a callable object"
```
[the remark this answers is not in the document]
kcelia: OK

tests/sklearn/test_sklearn_models.py
```diff
@@ -1526,7 +1567,11 @@ def test_p_error_global_p_error_simulation(
     def check_for_divergent_predictions(x, model, fhe, max_iterations=N_ALLOWED_FHE_RUN):
         """Detect divergence between simulated/FHE execution and clear run."""
         predict_function = (
-            model.predict_proba if is_classifier_or_partial_classifier(model) else model.predict
+            model.predict_proba
+            if is_classifier_or_partial_classifier(model)
+            # predict_prob not implemented yet for KNeighborsClassifier
```
RomanBredehoft, line 1572: worth a fixme here !

tests/deployment/test_client_server.py
```diff
@@ -99,6 +99,7 @@ def test_client_server_sklearn(
     with pytest.raises(AttributeError, match=".* model is not compiled.*"):
         client_server_simulation(x_train, x_test, model, default_configuration)
 
+    # With n_bits = 3, KNN is not compilable
     fhe_circuit = model.compile(
```
andrei-stoian-zama, line 103: We discussed adding a test here for KNeighborsClassifier that checks that `predict` in the clear is the same as `post_processing` of the results of the server.

andrei-stoian-zama (approved):
Looks good !

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1694,382 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+            n_bits (int): Number of bits to quantize the model. IThe value will be used for
+                quantizing inputs and X_fit. Default to 3.
+        """
+        self.n_bits: int = n_bits
+        # _q_fit_X: In distance metric algorithms, `_q_fit_X` stores the training set to compute
+        # the similarity or distance measures. There is no `weights` attribute because there isn't
+        # a training phase
+        self._q_fit_X: numpy.ndarray
+        # _y: Labels of `_q_fit_X`
+        self._y: numpy.ndarray
```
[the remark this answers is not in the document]
andrei-stoian-zama: yes, we can keep it but we need to make sure it doesn't exist in the model exported in the client (in client / server)

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: it's a constant no? it will be precomputed by CP

src/concrete/ml/sklearn/base.py
```diff
@@ -1690,3 +1694,382 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+                    max_x = a + numpy.maximum(0, b - a)
+
+                    # Swap if a > b
+                    # x[range_i] = max_x(a, b): First bitonic sequence gets min(a, b)
+                    x = scatter1d(x, a + b - max_x, range_i)
+                    # x[range_i + d] = min(a, b): Second bitonic sequence gets max(a, b)
+                    x = scatter1d(x, max_x, range_i + d)
+
+                    # Max index selection
+                    sign = diff <= 0
```
jfrery, line 1986: This is not the sign but a boolean I guess. Why isn't this done right after computing diff?

Have you tried the CP comparison optimization by replacing
```python
diff = a - b
sign = diff <= 0
```
to
```python
sign = a <= b
```
?

jfrery (commented):
Looks good just a few comments.

tests/sklearn/test_sklearn_models.py
kcelia: Added

src/concrete/ml/sklearn/base.py
kcelia: It's indeed a boolean that tells us if a is greater than b.
I'll change the naming.

src/concrete/ml/sklearn/base.py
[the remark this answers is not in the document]
RomanBredehoft: ah yes ok so basically it's for the `predict` but not for the post_processing right got it

tests/deployment/test_client_server.py
RomanBredehoft: I think the issue is broader than KNN right ? We want to add test for all regressors (predict vs server result) and classifiers (predict_proba vs server result, except for KNN which does not have predict_proba yet) ! 
-> https://github.com/zama-ai/concrete-ml-internal/issues/3996

src/concrete/ml/sklearn/base.py
jfrery: Maybe you can just check on a simple example if you have any time improvement. Otherwise you can leave it.

RomanBredehoft: if you don't have the time right now worth an issue then I think because this might be a good thing to try anyway

kcelia: I'll do it in a separate PR.
Some testings are needed.

jfrery: Great can you create an issue for this?

jfrery (approved):
Looks good to me! Thanks. Please just create an issue to check the CP comparison (see [comment](https://github.com/zama-ai/concrete-ml/pull/217#discussion_r1332608649))

andrei-stoian-zama (approved):

PR #222, opened 2023-09-04
chore: remove GPG import workaround in release workflow

RomanBredehoft:
The issue is already fixed : https://github.com/crazy-max/ghaction-import-gpg/issues/176

Also had to put back the `sudo apt-mark hold docker.io` even though we don't need it in regular CIs (still unsure why but it blocked the release process' deps installation)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -134,6 +134,10 @@ jobs:
 
       - name: Install dependencies
         run: |
+          # We need to freeze docker.io because its update requires user input
+          sudo apt update
+          sudo apt-mark hold docker.io
+
           ./script/make_utils/setup_os_deps.sh  --linux-install-python
           make setup_env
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -134,6 +134,10 @@ jobs:
 
       - name: Install dependencies
         run: |
+          # We need to freeze docker.io because its update requires user input
+          sudo apt update
+          sudo apt-mark hold docker.io
+
```
fd0r, line 140: We already encountered this IIRC, there is a way to automate the user inputs also IIRC

But this works for now.

RomanBredehoft: yeah but why don't we need i in regular CIs ?

fd0r: Ho actually the thing I remember is that when this happened the quick solution that we took was to update the AMI with the newest version of `docker.io`

RomanBredehoft: interesting, but still unsure why I encounter the issue in the release CI only

PR #223, opened 2023-09-04
chore: fix syntax error in release workflow

RomanBredehoft:
Forgot to fix a syntax error last time, I tested the script locally and it should now work properly. I also made the error messages a bit more clear

### DIFF
--- a/.github/workflows/prepare_release.yaml
+++ b/.github/workflows/prepare_release.yaml
@@ -70,11 +70,11 @@ jobs:
         run: |
           IS_PATCH="$(poetry run python ./script/make_utils/version_utils.py is_patch_release --version "$VERSION")"
 
-          if [[ "${IS_PATCH}" == 'true']]; then
+          if [[ "${IS_PATCH}" == "false" ]]; then
 
             if [[ "${BRANCH_IS_RELEASE}" == "false" ]]; then
               echo "Patch release cannot be done: target branch is not a release branch"
-              echo "Got branch ${BRANCH_NAME}"
+              echo "Got branch '${BRANCH_NAME}'"
               exit 1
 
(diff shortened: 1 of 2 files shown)


.github/workflows/prepare_release.yaml
```diff
@@ -70,11 +70,11 @@ jobs:
         run: |
           IS_PATCH="$(poetry run python ./script/make_utils/version_utils.py is_patch_release --version "$VERSION")"
 
-          if [[ "${IS_PATCH}" == 'true']]; then
+          if [[ "${IS_PATCH}" == "false" ]]; then
```
RomanBredehoft, line 73: the error was the missing empty space here

.github/workflows/release.yaml
```diff
@@ -181,22 +181,23 @@ jobs:
       # release version, the release preparation process is stopped as well
       - name: Check release preparation is from right branch
         run: |
-          if [[ "${{ env.IS_PATCH }}" == 'true']]; then
+          if [[ "${IS_PATCH}" == "true" ]]; then
```
RomanBredehoft, line 184: same for here, missing empty space after `"true"`

.github/workflows/release.yaml
```diff
@@ -181,22 +181,23 @@ jobs:
       # release version, the release preparation process is stopped as well
       - name: Check release preparation is from right branch
         run: |
-          if [[ "${{ env.IS_PATCH }}" == 'true']]; then
+          if [[ "${{ env.IS_PATCH }}" == "true" ]]; then
```
RomanBredehoft, line 184: same fore here

fd0r (approved):

PR #228, opened 2023-09-05
fix: fix issues with OMP library

bcm-at-zama:
there are path problems between torch and concrete omp libraries

closes #[3951](https://github.com/zama-ai/concrete-ml-internal/issues/3951)


bcm-at-zama:
With this, conv works again, as well as pytest, on my Mac intel

Makefile
```diff
@@ -56,6 +57,17 @@ sync_env: check_poetry_version
...
 	"$(MAKE)" setup_env
 
+.PHONY: fix_omp_issues_for_intel_mac # Fix OMP issues for macOS Intel, https://github.com/zama-ai/concrete-ml-internal/issues/3951
+fix_omp_issues_for_intel_mac:
+	if [[ $$(uname) == "Darwin" ]]; then \
+		WHICH_VENV=$$(which python | sed -e "s@bin/python@@"); \
+		rm $${WHICH_VENV}/lib/python3.9/site-packages/torch/lib/libiomp5.dylib; \
+		ln -s $${WHICH_VENV}/lib/python3.9/site-packages/concrete/.dylibs/libomp.dylib $${WHICH_VENV}/lib/python3.9/site-packages/torch/lib/libiomp5.dylib; \
+		rm $${WHICH_VENV}/lib/python3.9/site-packages/functorch/.dylibs/libiomp5.dylib; \
+		ln -s $${WHICH_VENV}/lib/python3.9/site-packages/concrete/.dylibs/libomp.dylib $${WHICH_VENV}lib/python3.9/site-packages/functorch/.dylibs/libiomp5.dylib; \
```
fd0r, line 67: Hum not sure about having python3.9 hard-coded here

bcm-at-zama: yes you're right

bcm-at-zama: what would you do, instead? I can try to find which python is available, but that's not obvious to me what to do if there is more than a single one. Maybe I can do it for all of them

bcm-at-zama:
By the way, I don't know why it works already in our CI for Mac, without these fixes. @youben11 told me that Concrete has these fixes in their CI

Makefile
bcm-at-zama: @youben11 : maybe you could guide me to which piece of code, on the Concrete CI, does that? I could see if I could borrow your ideas, if they apply here

youben11: In our CI, we already setup a specific version of python so we don't really have a problem. For your case, you can run the following command `python -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")'`

bcm-at-zama: Thx @youben11

RomanBredehoft:
> By the way, I don't know why it works already in our CI for Mac, without these fixes. @youben11 told me that Concrete has these fixes in their CI

we never run the mac intel CI right ? or maybe in the weekly (and in that case we never check) ?

bcm-at-zama:
We used to run Mac in weekly builds, and I would have imagined it was intel since at the beginning, Concrete did not work on ARM

bcm-at-zama:
The issue was there https://github.com/zama-ai/concrete-ml-internal/actions/runs/6066968202/job/16458138665, but no-one noticed. It would be important to keep an eye on weekly builds please, they find issues that are not seen in regular builds, CC @andrei-stoian-zama

fd0r:
Maybe we could have a specific point during our weekly meeting to check why the weekly build failed if it did.

RomanBredehoft:
> Maybe we could have a specific point during our weekly meeting to check why the weekly build failed if it did.

good idea yes

Makefile
bcm-at-zama: should be fixed now

deps_licenses/licenses_mac_intel_user.txt
```diff
@@ -41,7 +42,7 @@ protobuf, 3.20.3, BSD-3-Clause
 psutil, 5.9.5, BSD License
 pycparser, 2.21, BSD License
 pydantic, 1.10.12, MIT License
-pytest, 7.2.0, MIT License
+pytest, 7.4.1, MIT License
```
bcm-at-zama, line 45: by the way, I updated Mac intel license files. Would there be a way to do it automatically in your action, @RomanBredehoft ?

RomanBredehoft: not sure I understand the question ? The action (which was made by @fd0r , credits to him !) updates automatically linux and mac intel licenses once triggered) , but you've already done it for mac here (are you talking about linux then ?)

bcm-at-zama: and ok, the question is for @fd0r then, sorry for mis direction

bcm-at-zama:
Pushing a new version which should work with `make shell_lint` (which was already ok on my Mac).

bcm-at-zama:
The build is green

deps_licenses/licenses_mac_intel_user.txt
bcm-at-zama: OK I see: the action to refresh is not launched automatically, is it? and, as we check licenses only on linux in our builds, we don't see it's not uptodate. But at the same time, when the last person launched the refresh for linux, the refresh should also have been done for Mac intel, no? 

Let's wait @fd0r maybe

RomanBredehoft: no we have to trigger it by targeting our branch, not sure we want to run a make license and add a commit each time we push something (if it's even doable) ..

if you trigger the action, it will do both linux and mac intel yes. You can try it out, it's just here : https://github.com/zama-ai/concrete-ml/blob/main/.github/workflows/update_licenses.yaml

The only thing @fd0r wanted to improve is that for now, it will open 2 PRs targetting your branch (no need to wait for CIs as they don't target main) that you'll need to merge, but ultimately we'll make them directly commit in your branch (as notebook refreshed are done)

bcm-at-zama: >  not sure we want to run a make license and add a commit each time we push something

we would only commit if something changed. But ok keep it the way it is, if it's convenient for you

RomanBredehoft: I'm still not sure we can commit within a ci without having the recursive error you faced long ago anyway !

RomanBredehoft (approved):
thanks !

RomanBredehoft:
CI is running in : https://github.com/zama-ai/concrete-ml/actions/runs/6144015302

RomanBredehoft (changes requested):
The MacOS CI failed with something related to OMP : https://github.com/zama-ai/concrete-ml/actions/runs/6144015302/job/16668575489

RomanBredehoft (approved):
thanks a lot !

fd0r (approved):

PR #229, opened 2023-09-05
fix: fix bad top1 accuracy in cifar_brevitas_training use case

RomanBredehoft:
We observed a small (~1.5% with 6 bits rounding) drop in accuracy when using a better (more representative) compilation inputset, which is currently being investigated : https://github.com/zama-ai/concrete-ml-internal/issues/3957

In the mean time, this should solve the very low top1 accuracy we currently face (17% -> 86% top1 accuracy with 6 bits rounding)  

closes https://github.com/zama-ai/concrete-ml-internal/issues/3939

### DIFF
--- a/use_case_examples/cifar_brevitas_training/bnn_pynq_train.py
+++ b/use_case_examples/cifar_brevitas_training/bnn_pynq_train.py
@@ -71,7 +71,7 @@ def parse_args(args):
     )
     add_bool_arg(parser, "detect_nan", default=False)
     # Compute resources
-    parser.add_argument("--num_workers", default=4, type=int, help="Number of workers")
+    parser.add_argument("--num_workers", default=0, type=int, help="Number of workers")
     parser.add_argument("--gpus", type=none_or_str, default=None, help="Comma separated GPUs")
     # Optimizer hyper-parameters
     parser.add_argument("--batch_size", default=100, type=int, help="batch size")
--- a/use_case_examples/cifar_brevitas_training/evaluate_torch_cml.py
+++ b/use_case_examples/cifar_brevitas_training/evaluate_torch_cml.py
@@ -6,27 +6,17 @@ import torch
 from concrete.fhe import Configuration, ParameterSelectionStrategy
 from models import cnv_2w2a
 from torch.utils.data import DataLoader
-from torchvision import transforms
-from torchvision.datasets import CIFAR10
 from tqdm import tqdm
-from trainer import accuracy
+from trainer import accuracy, get_test_set, get_train_set
 
 from concrete.ml.torch.compile import compile_brevitas_qat_model
 
 
-def evaluate(torch_model, cml_model, device):
-
-    # Transform pipeline that normalizes the data between -1 and 1
-    transform_to_tensor = transforms.Compose(
-        [
-            transforms.ToTensor(),
-            transforms.Lambda(lambda x: 2.0 * x - 1.0),
-        ]
-    )
+def evaluate(torch_model, cml_model, device, num_workers):
 
     # Import and load the CIFAR test dataset (following bnn_pynq_train.py)
-    test_set = CIFAR10(root=".datasets/", train=False, download=True, transform=transform_to_tensor)
-    test_loader = DataLoader(test_set, batch_size=128, shuffle=False, num_workers=0)
+    test_set = get_test_set(dataset="CIFAR10", datadir=".datasets/")
+    test_loader = DataLoader(test_set, batch_size=128, shuffle=False, num_workers=num_workers)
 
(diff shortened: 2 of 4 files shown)


use_case_examples/cifar_brevitas_training/evaluate_torch_cml.py
```diff
@@ -97,9 +87,15 @@ def main(args):
     # Load weights
     model.load_state_dict(checkpoint["state_dict"], strict=False)
 
-    # Create a representative input-set that will be used used for both computing quantization
-    # parameters and compiling the model
-    input_set = torch.randn(1, 3, 32, 32)
```
RomanBredehoft, line 102: this is the main change basically

use_case_examples/cifar_brevitas_training/bnn_pynq_train.py
```diff
@@ -71,7 +71,7 @@ def parse_args(args):
     )
     add_bool_arg(parser, "detect_nan", default=False)
     # Compute resources
-    parser.add_argument("--num_workers", default=4, type=int, help="Number of workers")
+    parser.add_argument("--num_workers", default=0, type=int, help="Number of workers")
```
fd0r, line 74: 🙏🏼

fd0r (approved):

use_case_examples/cifar_brevitas_training/evaluate_torch_cml.py
```diff
@@ -97,9 +87,15 @@ def main(args):
...
+    # Load the training set
+    train_set = get_train_set(dataset="CIFAR10", datadir=".datasets/")
+
+    # Create a representative input-set from the training set that will be used used for both
+    # computing quantization parameters and compiling the model
+    num_samples = 100
+    input_set = torch.stack(
+        [train_set[index][0] for index in range(min(num_samples, len(train_set)))]
+    )
```
jfrery, line 98: Is the dataset seeded? Let's make sure we have the same calibration set so that we don't include randomness in the calibration.

use_case_examples/cifar_brevitas_training/bnn_pynq_train.py
jfrery: Have you tried this on a linux machine? I remember this failed on mine but it's probably fixed now? I can try if you want.

jfrery (commented):
Thanks! looks great to me.

use_case_examples/cifar_brevitas_training/bnn_pynq_train.py
RomanBredehoft: yes if you could try that would be great, it fails on Mac if it's not set to 0

use_case_examples/cifar_brevitas_training/evaluate_torch_cml.py
RomanBredehoft: it has never been seeded in any use case if I recall, not sure we want to start to do that, especially since this inputset is probably less "random" than using the initial `torch.randn(1, 3, 32, 32)` 😅

use_case_examples/cifar_brevitas_training/trainer.py
```diff
@@ -90,6 +91,61 @@ def accuracy(
         return list_topk_accs
 
 
+def get_train_set(dataset: str, datadir: Union[str, Path]) -> Dataset:
```
kcelia, line 94: It doesn't concern this PR, but we have a lot of redundancy between cifar_training and cifar_fietuning. Such as `get_train/test_set` in equivalent to `get_torchvision_dataset` in cifar_finetuning.

We should discuss the refactoring in the next meeting.

use_case_examples/cifar_brevitas_training/bnn_pynq_train.py
jfrery: Looks like it's working for me

use_case_examples/cifar_brevitas_training/evaluate_torch_cml.py
jfrery: Yes that torch.randn was not great but the calibration data point was not having any impact on accuracy. Now it does so maybe we want to stick with one accuracy to be able to see any other causes of drop (else than random).

RomanBredehoft: I don't think it matters a lot, the accuracy is stable with a 100 samples normalized between -1 and 1. Besides, we already have an issue for seeding notebooks : https://github.com/zama-ai/concrete-ml-internal/issues/1713

kcelia (approved):

PR #230, opened 2023-09-06
chore: remove use of MPS device

RomanBredehoft:
Basically I observed a drop in torch's accuracy by just using MPS devices, and I could not find any reason for it (explanations in the issue). I suggest that we disable it for now until we properly test it / the bug is fixed  

refs https://github.com/zama-ai/concrete-ml-internal/issues/3953


fd0r (approved):

fd0r:
Maybe: https://github.com/pytorch/pytorch/issues/83015 ?

RomanBredehoft:
> Maybe: https://github.com/pytorch/pytorch/issues/83015 ?

good finding, I will first check if setting `non_blocking=False`  fixes the issue

RomanBredehoft:
I've just checked and disabling `non_blocking`did not solve anything, i therefore merge this PR until we know more about it

PR #231, opened 2023-09-06
chore: moving towards autorounding

fd0r:
Experimental commit.

I’m in the arena trying stuff. Some will work, some won’t. But always learning.

### DIFF
--- a/src/concrete/ml/quantization/base_quantized_op.py
+++ b/src/concrete/ml/quantization/base_quantized_op.py
@@ -5,12 +5,12 @@ from inspect import Parameter, _empty, signature
 from typing import Any, Callable, Dict, List, Optional, Set, TextIO, Tuple, Type, Union, cast
 
 import numpy
+from concrete.fhe.tracing import Tracer
 
 from concrete import fhe
 
 from ..common.debugging import assert_false, assert_true
 from ..common.serialization.dumpers import dump, dumps
-from ..common.utils import compute_bits_precision
 from ..onnx.onnx_utils import ONNX_OPS_TO_NUMPY_IMPL
 from ..onnx.ops_impl import ONNXMixedFunction, RawOpOutput
 from .quantizers import (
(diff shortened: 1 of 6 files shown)


fd0r:
No review needed at this point, I just wanted to see how does this behaves in the CI.

@jfrery we don't need this calibration step if auto-round calibration is done at compile time right?

fd0r:
Using this auto-round features fixes the issue mentioned in:
https://github.com/zama-ai/concrete-ml-internal/issues/3429#issuecomment-1660372429

jfrery:
> @jfrery we don't need this calibration step if auto-round calibration is done at compile time right?

So the problem is that we need the rounding to work when we calibrate the quantized ops. Otherwise we will have different values between concrete and cml. I think the whole complexity is there. We need to compile first to have the fhe.rounding_pattern to work.

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -945,27 +950,6 @@ def cnp_round(
...
-
-            # mypy
-            assert self.lsbs_to_remove is not None
-
-            # Apply rounding if needed
-            if self.lsbs_to_remove > 0:
-                x = fhe.round_bit_pattern(x, lsbs_to_remove=self.lsbs_to_remove)
-
+        if self.rounder is not None and not calibrate_rounding:
+            x = fhe.round_bit_pattern(x, lsbs_to_remove=self.rounder)
```
jfrery, line 954: So this won't work because you have not compiled at this point. Compilation isn't part of the calibration it happens afterward.

fd0r:
Fixing conformance and checking what breaks.

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Actually this is not executed at calibration time so no failures.

jfrery (commented):
Looks great to me! So the trick here is that rounding is not used during cml calibration. Maybe we can add correctness test to make sure clear quantized == simulation with rounding.

jfrery:
What do we do here?

fd0r:
🤷🏼 

I guess we don't want this in the end right @andrei-stoian-zama ?

PR #233, opened 2023-09-07
chore: fix pre processing cifar fhe

jfrery:
The fhe inference has both cifar10 and cifar100 preprocessing values wrong as well.


jfrery:
My bad this time it's the std that comes first and not the mean. I read too fast.

PR #236, opened 2023-09-08
fix: make sure structured pruning and unstructured pruning work well together

andrei-stoian-zama:
Structured pruning and unstructured pruning were causing some bad interference. 

Having both on the same layer caused pruning to be removed twice from that layer which causes a torch assert.


RomanBredehoft (approved):
indeed, thanks a lot !

PR #237, opened 2023-09-08
chore: clean and refactor pytest commands in Makefile

RomanBredehoft:
there was a lot of redundancy , now it should be cleaner

closes https://github.com/zama-ai/concrete-ml-internal/issues/3892


RomanBredehoft:
triggered the CI manually as I touched pytest : https://github.com/zama-ai/concrete-ml/actions/runs/6123285209

Makefile
```diff
@@ -190,125 +192,112 @@ SPCC_DEPS := check_python_format pylint_src pylint_tests mypy mypy_test pydocsty
...
 	--randomly-dont-reset-seed \
+	--capture=tee-sys \
 	--cache-clear \
 	${PYTEST_OPTIONS}
 
-.PHONY: pytest_one # Run pytest on a single file or directory (TEST) a certain number of times (COUNT)
-pytest_one:
-	poetry run pytest \
-	-svv \
-	-n $$(./script/make_utils/ncpus.sh) \
```
jfrery, line 233: So we completely remove the command?

Makefile
```diff
@@ -6,6 +6,8 @@ DEV_CONTAINER_VENV_VOLUME:=concrete-ml-venv
 DEV_CONTAINER_CACHE_VOLUME:=concrete-ml-cache
 DOCKER_VENV_PATH:="$${HOME}"/dev_venv/
 SRC_DIR:=src
+TEST:=tests
+N_CPU:=4
```
jfrery, line 10: What is this used for? pytest? should we really hardcode this?

RomanBredehoft:
tests passed fine

jfrery (commented):
Thanks for the cleanup + clarification in the comments. Just a few comments.

Makefile
RomanBredehoft: no it's still there, I just moved it below in the file

jfrery: Oh alright got it.

Makefile
jfrery: Alright then. It's just that this 4 looks really arbitrary but if we were doing this before then fine.

jfrery (dismissed):
Thanks a lot!

Makefile
RomanBredehoft: yes it is, but there's a coment in the pytest command that sayd that we need to set a low number (lower than the machine's number of CPUs), otherwise it fails : 
> because most tests include parallel execution using all CPUs and too many parallel tests would lead to contention. Thus N is set to something lower than the overall number of CPUs

RomanBredehoft:
actually tests did not pass in last CI (which should be fixed by https://github.com/zama-ai/concrete-ml/pull/244), I need to check again

kcelia (approved):
Thanks

PR #238, opened 2023-09-08
chore: trigger CI if Makefile changes

RomanBredehoft:
Not a huge fan of triggering the CI if any changes has been found in the makefile as not all commands are impacting the CI, but since there are still a few of them I can't find a better solution than that (cc @bcm-at-zama )

closes https://github.com/zama-ai/concrete-ml-internal/issues/3955


bcm-at-zama:
Tx @RomanBredehoft

PR #244, opened 2023-09-11
chore: run all CI steps if triggered manually

RomanBredehoft:
By default, several steps would be skipped if the CI was triggered manually through GitHub's Action interface (ex: https://github.com/zama-ai/concrete-ml/actions/runs/6123285209/job/16620945850)

I fixed that by skipping the "check for changes" step, which should prevent other steps from being skipped

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -32,7 +32,7 @@ on:
         type: string
         default: "3.8 3.9"
       manual_call:
-        description: 'To distinguish workflow_call from workflow_dispatch'
+        description: "Do not uncheck this"
         type: boolean
         required: false
         default: true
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -32,7 +32,7 @@ on:
         type: string
         default: "3.8 3.9"
       manual_call:
-        description: 'To distinguish workflow_call from workflow_dispatch'
+        description: "Do not uncheck this"
```
RomanBredehoft, line 35: For clarity when triggering the step manually. See this comment for context : 
  

> Workflow call refers to the release process (it enables the current CI workflow to be called by
>  another workflow from the same repository, in this case the release one)
>  The "release" event is not put by default in order to avoid running the following CI without
>  explicitly indicating it in the caller workflow
>  Besides, GitHub actions are not able to differentiate 'workflow_dispatch' from 'workflow_call'
>  based on 'github.event_name' and both are set to 'workflow_dispatch'. Therefore, an optional 
>  input 'manual_call' with proper default values is added to both as a workaround, following one 
>  user's suggestion : https://github.com/actions/runner/discussions/1884
>  FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3930

kcelia (approved):
Thanks

PR #246, opened 2023-09-13
chore: implement hybrid model demo with GPT-2

fd0r:
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3842 
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3869
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3875
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3855
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3852

### DIFF
--- a/src/concrete/ml/deployment/fhe_client_server.py
+++ b/src/concrete/ml/deployment/fhe_client_server.py
@@ -54,11 +54,14 @@ class FHEModelServer:
                 versions = json.load(file)
 
         errors = []
-        packages_to_check = {"concrete-python"}
+        packages_to_check = {"concrete-python", "concrete-ml"}
         for package_name, package_version in versions.items():
             if package_name not in packages_to_check:
                 continue
-            current_version = version(package_name)
+            if package_name == "concrete-ml":
+                current_version = CML_VERSION
+            else:
+                current_version = version(package_name)
             if package_version != current_version:  # pragma: no cover
                 errors.append((package_name, package_version, current_version))
         if errors:  # pragma: no cover
@@ -190,13 +193,10 @@ class FHEModelDev:
         # Add versions
         versions_path = Path(self.path_dir).joinpath("versions.json")
         versions = {
-            package_name: version(package_name)
-            for package_name in ["concrete-ml", "concrete-python"]
+            "concrete-python": version("concrete-python"),
+            "concrete-ml": CML_VERSION,
+            "python": f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}",
         }
-        versions[
-            "python"
-        ] = f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
-
         with open(versions_path, "w", encoding="utf-8") as file:
             json.dump(fp=file, obj=versions)
 
@@ -251,11 +251,14 @@ class FHEModelClient:
                 versions = json.load(file)
 
         errors = []
-        packages_to_check = {"concrete-python"}
+        packages_to_check = {"concrete-python", "concrete-ml"}
         for package_name, package_version in versions.items():
             if package_name not in packages_to_check:
                 continue
-            current_version = version(package_name)
+            if package_name == "concrete-ml":
+                current_version = CML_VERSION
+            else:
+                current_version = version(package_name)
             if package_version != current_version:  # pragma: no cover
                 errors.append((package_name, package_version, current_version))
         if errors:  # pragma: no cover
--- a/src/concrete/ml/onnx/convert.py
+++ b/src/concrete/ml/onnx/convert.py
@@ -2,23 +2,116 @@
 
 import tempfile
 from pathlib import Path
-from typing import Callable, Optional, Tuple, Union
+from typing import Callable, Tuple, Union
 
 import numpy
 import onnx
+import onnxoptimizer
 import torch
-from onnx import checker
+from onnx import checker, helper
 
 from .onnx_utils import IMPLEMENTED_ONNX_OPS, execute_onnx_with_numpy, get_op_type
 
 OPSET_VERSION_FOR_ONNX_EXPORT = 14
 
 
-def get_equivalent_numpy_forward_and_onnx_model(
+# pylint: disable=too-many-branches
+def fuse_matmul_bias_to_gemm(onnx_model: onnx.ModelProto):
+    """Fuse sequence of matmul -> add into a gemm node.
+
+    Args:
+        onnx_model (onnx.ModelProto): A onnx model to optimize using Mat-Mult + Add -> Gemm
+
+    Returns:
+        onnx.ModelProto: the optimized onnx model
+
+    """
+    # Convert nodes to list to avoid modifying iterable during iteration
+    nodes_list = list(onnx_model.graph.node)
+
+    # Iterate through graph nodes
+    for matmul_node in nodes_list:
+        # Run only if the node is a MatMul node
+        if matmul_node.op_type != "MatMul":
+            continue
+        # Store MatMul node output name
+        matmul_node_output_name = matmul_node.output[0]
+        assert len(matmul_node.output) == 1
+
+        # Make sure that only one node uses the output of the mat-mult
+        mat_mult_output_use_node = []
+        for other_node in onnx_model.graph.node:
+            if other_node is matmul_node:
+                continue
+            if matmul_node_output_name in other_node.input:
+                mat_mult_output_use_node.append(other_node)
+        if len(mat_mult_output_use_node) != 1:
+            continue
+
+        # Check that following node is Add
+        add_node = mat_mult_output_use_node[0]
+        if add_node.op_type != "Add":
+            continue
+        assert len(add_node.output) == 1
+
+        # Find other Add input
+        bias_other_input_node_name = None
+        for input_name in add_node.input:
+            if input_name != matmul_node_output_name:
+                bias_other_input_node_name = input_name
+        assert bias_other_input_node_name is not None
+
+        # Only merge if the input of the add node is an initializer
+        # otherwise there might be some scaling issues
+        initializer_names = [elt.name for elt in onnx_model.graph.initializer]
+        if bias_other_input_node_name not in initializer_names:
+            continue
+
+        # Create a GEMM node which combines the MatMul and Add operations
+        gemm_node = helper.make_node(
+            "Gemm",  # op_type
+            [matmul_node.input[0], matmul_node.input[1], bias_other_input_node_name],  # inputs
+            [add_node.output[0]],  # outputs
+            name="Gemm_Node",
+            alpha=1.0,
+            beta=1.0,  # attributes
+        )
+        assert len(gemm_node.output) == 1
+
+        # Replace the MatMul and Add nodes with the GEMM node
+        # The graph needs to keep being topologically sorted
+        mat_mult_node_index = list(onnx_model.graph.node).index(matmul_node)
+        add_node_index = list(onnx_model.graph.node).index(add_node)
+        gemm_node_index = max(mat_mult_node_index, add_node_index)
+
+        onnx_model.graph.node.insert(gemm_node_index, gemm_node)
+        onnx_model.graph.node.remove(add_node)
+        onnx_model.graph.node.remove(matmul_node)
+
+        # Update connections in the graph
+        for potential_next_node in onnx_model.graph.node:
+            # Check if this node was connected to the add_node
+            if add_node.output[0] not in potential_next_node.input:
+                continue
+
+            # replace the reference to the old add_node output with the gemm_node output
+            for idx, potential_next_node_input in enumerate(potential_next_node.input):
+                if potential_next_node_input == add_node.output[0]:
+                    potential_next_node.input[idx] = gemm_node.output[0]
+
+        # Update the model's output if necessary
+        for model_output in onnx_model.graph.output:
+            if model_output.name == add_node.output[0]:
+                model_output.name = gemm_node.output[0]
+
+    return onnx_model
+
+
+def get_equivalent_numpy_forward_from_torch(
     torch_module: torch.nn.Module,
     dummy_input: Union[torch.Tensor, Tuple[torch.Tensor, ...]],
-    output_onnx_file: Optional[Union[Path, str]] = None,
-) -> Tuple[Callable[..., Tuple[numpy.ndarray, ...]], onnx.GraphProto]:
+    output_onnx_file: Union[None, Path, str] = None,
+) -> Tuple[Callable[..., Tuple[numpy.ndarray, ...]], onnx.ModelProto]:
     """Get the numpy equivalent forward of the provided torch Module.
 
     Args:
@@ -34,12 +127,12 @@ def get_equivalent_numpy_forward_and_onnx_model(
             execute the equivalent numpy code to the passed torch_module and the generated ONNX
             model.
     """
-
     output_onnx_file_path = Path(
         tempfile.mkstemp(suffix=".onnx")[1] if output_onnx_file is None else output_onnx_file
     )
     use_tempfile: bool = output_onnx_file is None
 
+    # Export to ONNX
     torch.onnx.export(
         torch_module,
         dummy_input,
@@ -47,17 +140,15 @@ def get_equivalent_numpy_forward_and_onnx_model(
         opset_version=OPSET_VERSION_FOR_ONNX_EXPORT,
     )
     equivalent_onnx_model = onnx.load_model(str(output_onnx_file_path))
-
-    checker.check_model(equivalent_onnx_model)
-
     # Remove the tempfile if we used one
     if use_tempfile:
         output_onnx_file_path.unlink()
 
-    # The model was checked just above
-    equivalent_numpy_forward = get_equivalent_numpy_forward(
-        equivalent_onnx_model, check_model=False
+    equivalent_numpy_forward, equivalent_onnx_model = get_equivalent_numpy_forward_from_onnx(
+        equivalent_onnx_model, check_model=True
     )
+    with output_onnx_file_path.open("wb") as file:
+        file.write(equivalent_onnx_model.SerializeToString())
 
     return (
         equivalent_numpy_forward,
@@ -65,10 +156,10 @@ def get_equivalent_numpy_forward_and_onnx_model(
     )
 
 
-def get_equivalent_numpy_forward(
+def get_equivalent_numpy_forward_from_onnx(
     onnx_model: onnx.ModelProto,
     check_model: bool = True,
-) -> Callable[..., Tuple[numpy.ndarray, ...]]:
+) -> Tuple[Callable[..., Tuple[numpy.ndarray, ...]], onnx.ModelProto]:
     """Get the numpy equivalent forward of the provided ONNX model.
 
     Args:
@@ -87,9 +178,29 @@ def get_equivalent_numpy_forward(
     """
     if check_model:
         checker.check_model(onnx_model)
-    required_onnx_operators = set(get_op_type(node) for node in onnx_model.graph.node)
-    unsupported_operators = required_onnx_operators - IMPLEMENTED_ONNX_OPS
+    checker.check_model(onnx_model)
 
+    # Optimize ONNX graph
+    # List of all currently supported onnx optimizer passes
+    # From https://github.com/onnx/optimizer/blob/master/onnxoptimizer/pass_registry.h
+    onnx_passes = [
+        "fuse_matmul_add_bias_into_gemm",
+        "eliminate_nop_pad",
+        "fuse_pad_into_conv",
+        "extract_constant_to_initializer",
+        "eliminate_unused_initializer",
+    ]
+    equivalent_onnx_model = onnxoptimizer.optimize(onnx_model, onnx_passes)
+    checker.check_model(equivalent_onnx_model)
+    # Custom optimization
+    # ONNX optimizer does not optimize Mat-Mult + Bias pattern into GEMM if the input isn't a matrix
+    # We manually do the optimization for this case
+    equivalent_onnx_model = fuse_matmul_bias_to_gemm(equivalent_onnx_model)
+    checker.check_model(equivalent_onnx_model)
+
+    # Check supported operators
+    required_onnx_operators = set(get_op_type(node) for node in equivalent_onnx_model.graph.node)
+    unsupported_operators = required_onnx_operators - IMPLEMENTED_ONNX_OPS
     if len(unsupported_operators) > 0:
         raise ValueError(
             "The following ONNX operators are required to convert the torch model to numpy but are "
@@ -97,4 +208,7 @@ def get_equivalent_numpy_forward(
             f"Available ONNX operators: {', '.join(sorted(IMPLEMENTED_ONNX_OPS))}"
         )
 
-    return lambda *args: execute_onnx_with_numpy(onnx_model.graph, *args)
+    # Return lambda of numpy equivalent of onnx execution
+    return (
+        lambda *args: execute_onnx_with_numpy(equivalent_onnx_model.graph, *args)
+    ), equivalent_onnx_model
--- a/src/concrete/ml/torch/compile.py
+++ b/src/concrete/ml/torch/compile.py
@@ -3,7 +3,7 @@
 import tempfile
 import warnings
 from pathlib import Path
-from typing import Optional, Tuple, Union
+from typing import Dict, Optional, Tuple, Union
 
 import numpy
 import onnx
@@ -55,7 +55,7 @@ def build_quantized_module(
     model: Union[torch.nn.Module, onnx.ModelProto],
     torch_inputset: Dataset,
     import_qat: bool = False,
-    n_bits=MAX_BITWIDTH_BACKWARD_COMPATIBLE,
+    n_bits: Union[int, Dict[str, int]] = MAX_BITWIDTH_BACKWARD_COMPATIBLE,
     rounding_threshold_bits: Optional[int] = None,
 ) -> QuantizedModule:
     """Build a quantized module from a Torch or ONNX model.
@@ -113,7 +113,7 @@ def _compile_torch_or_onnx_model(
     configuration: Optional[Configuration] = None,
     artifacts: Optional[DebugArtifacts] = None,
     show_mlir: bool = False,
-    n_bits=MAX_BITWIDTH_BACKWARD_COMPATIBLE,
+    n_bits: Union[int, Dict[str, int]] = MAX_BITWIDTH_BACKWARD_COMPATIBLE,
     rounding_threshold_bits: Optional[int] = None,
     p_error: Optional[float] = None,
     global_p_error: Optional[float] = None,
@@ -272,7 +272,7 @@ def compile_onnx_model(
     configuration: Optional[Configuration] = None,
     artifacts: Optional[DebugArtifacts] = None,
     show_mlir: bool = False,
-    n_bits=MAX_BITWIDTH_BACKWARD_COMPATIBLE,
+    n_bits: Union[int, Dict] = MAX_BITWIDTH_BACKWARD_COMPATIBLE,
     rounding_threshold_bits: Optional[int] = None,
     p_error: Optional[float] = None,
     global_p_error: Optional[float] = None,
@@ -340,7 +340,7 @@ def compile_brevitas_qat_model(
     rounding_threshold_bits: Optional[int] = None,
     p_error: Optional[float] = None,
     global_p_error: Optional[float] = None,
-    output_onnx_file: Union[Path, str] = None,
+    output_onnx_file: Union[None, Path, str] = None,
     verbose: bool = False,
 ) -> QuantizedModule:
     """Compile a Brevitas Quantization Aware Training model.
@@ -378,7 +378,6 @@ def compile_brevitas_qat_model(
     Returns:
         QuantizedModule: The resulting compiled QuantizedModule.
     """
-
     inputset_as_numpy_tuple = tuple(
         convert_torch_tensor_or_numpy_array_to_numpy_array(val) for val in to_tuple(torch_inputset)
     )
@@ -418,8 +417,11 @@ def compile_brevitas_qat_model(
     # https://github.com/onnx/optimizer/blob/master/onnxoptimizer/pass_registry.h
     # In the export function, the `args` parameter is used instead of the `input_shape` one in
     # order to be able to handle multi-inputs models
-    exporter.onnx_passes.append("eliminate_nop_pad")
-    exporter.onnx_passes.append("fuse_pad_into_conv")
+    exporter.onnx_passes += [
+        "eliminate_nop_pad",
+        "fuse_pad_into_conv",
+        "fuse_matmul_add_bias_into_gemm",
+    ]
     onnx_model = exporter.export(
         torch_model,
         args=dummy_input_for_tracing,
--- a/src/concrete/ml/torch/hybrid_model.py
+++ b/src/concrete/ml/torch/hybrid_model.py
@@ -1,16 +1,45 @@
 """Implement the conversion of a torch model to a hybrid fhe/torch inference."""
 
+import ast
+import enum
+import io
+import sys
+import time
 import uuid
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
+import numpy
+import requests
 import torch
 from concrete.fhe import Configuration
 from torch import nn
 from transformers import Conv1D
 
+from ..common.utils import MAX_BITWIDTH_BACKWARD_COMPATIBLE
 from ..deployment.fhe_client_server import FHEModelClient, FHEModelDev
-from .compile import compile_torch_model
+from .compile import QuantizedModule, compile_torch_model
+
+
+class HybridFHEMode(enum.Enum):
+    """Simple enum for different modes of execution of HybridModel."""
+
+    DISABLE = "disable"  # Use torch weights
+    REMOTE = "remote"  # Use remote FHE server
+    SIMULATE = "simulate"  # Use FHE simulation
+    CALIBRATE = "calibrate"  # Use calibration (to run before FHE compilation)
+
+
+def tuple_to_underscore_str(tup: Tuple) -> str:
+    """Convert a tuple to a string representation.
+
+    Args:
+        tup (Tuple): a tuple to change into string representation
+
+    Returns:
+        str: a string representing the tuple
+    """
+    return repr(tup).replace("(", "p_").replace(")", "_p").replace(", ", "_")
 
 
 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3858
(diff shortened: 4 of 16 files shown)


andrei-stoian-zama:
Closes 
- https://github.com/zama-ai/concrete-ml-internal/issues/3842 ?
- https://github.com/zama-ai/concrete-ml-internal/issues/3869 ?

fd0r:
Yes I think so.
Currently fixing the batch-size/gemm optimization issue.

src/concrete/ml/torch/compile.py
```diff
@@ -418,8 +414,11 @@ def compile_brevitas_qat_model(
     # https://github.com/onnx/optimizer/blob/master/onnxoptimizer/pass_registry.h
     # In the export function, the `args` parameter is used instead of the `input_shape` one in
     # order to be able to handle multi-inputs models
-    exporter.onnx_passes.append("eliminate_nop_pad")
-    exporter.onnx_passes.append("fuse_pad_into_conv")
+    exporter.onnx_passes += [
+        "eliminate_nop_pad",
+        "fuse_pad_into_conv",
+        "fuse_matmul_add_bias_into_gemm",
```
jfrery, line 420: This doesn't work, right?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -188,28 +366,36 @@ def _get_module_by_name(model: nn.Module, name: str) -> Union[RemoteModule, nn.M
...
 
     def compile_model(
         self,
         x: torch.Tensor,
         n_bits: int = 8,
-        rounding_threshold_bits: int = 8,
-        p_error=0.01,
-        configuration: Configuration = None,
+        rounding_threshold_bits: Optional[int] = 8,
+        p_error: float = 0.01,
```
jfrery, line 397: Default should follow the one from CML

tests/torch/test_compile_torch.py
```diff
@@ -1056,7 +1056,7 @@ def test_net_has_no_tlu(
     if num_inputs > 1 and use_qat:
         return
 
-    use_conv = isinstance(input_shape, tuple) and len(input_shape) > 1
+    use_conv = isinstance(input_shape, tuple) and len(input_shape) == 3
```
jfrery, line 1059: What's happening?

use_case_examples/hybrid_model/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+This use case example showcases how to partially run layers in FHE.
+
+In this case we apply a fully connected layer of a GPT-2 model in FHE.
+
+## How to run this use-case
+
+0. Install additional requirements using `python -m pip install -r requirements.txt`
+1. Compile GPT-2 model using `bash compile.sh` script
+1. Run FHE server using `bash serve.sh`
```
jfrery, line 11: server.sh ?

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,140 @@
...
+        single_precision=False,
+        parameter_selection_strategy=ParameterSelectionStrategy.MULTI,
+    )
+
+    # Create a hybrid model
+    hybrid_model = HybridFHEModel(model, module_names)
+    hybrid_model.compile_model(
+        inputs,
+        n_bits=8,
+        # setting it to None is not enough -> weird
```
jfrery, line 37: Can we remove? What did you mean here?

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,140 @@
...
+        rounding_threshold_bits=None,
+        configuration=configuration,
+    )
+
+    # Sanity checks
+    logits_simulate = hybrid_model(inputs, fhe="simulate").logits
+    logits_disable = hybrid_model(inputs, fhe="disable").logits
+    logits_original = model(inputs).logits
+    # Ensure logits_disable and logits_original return the same output for the logits
```
jfrery, line 46: Could you add a space before the comments?

use_case_examples/hybrid_model/load_and_analyze_data.py
jfrery, line 1: What is this file? Doesn't look like to be defined in the readme?

use_case_examples/llm/QGPT2Evaluate.ipynb
```diff
@@ -695,5 +695,5 @@
   }
  },
  "nbformat": 4,
- "nbformat_minor": 2
+ "nbformat_minor": 4
```
jfrery, line 698: Is this expected?

jfrery (changes requested):
Nicely done! I have a few comments as well as question for my understanding.

use_case_examples/llm/QGPT2Evaluate.ipynb
fd0r: No idea actually, `nbqa` probably? 🤷🏼

src/concrete/ml/torch/compile.py
fd0r: They work, but depending on the shape of the input.

tests/torch/test_compile_torch.py
fd0r: I can change that back but I found it confusing when I was debugging

use_case_examples/hybrid_model/README.md
fd0r: No no, `serve.sh` as in to serve a FHE model.

src/concrete/ml/torch/hybrid_model.py
fd0r: alright yeah I can change that, it was like that before so I didn't consider changing it

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Hum not sure on that, I had a weird behavior if not set, I'll try removing it

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Do you mean a blank line?

src/concrete/ml/torch/hybrid_model.py
jfrery, line 1: Could you add the file to the __init__.py so that we can do `from concrete.ml.torch import HybridFHEModel` ?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -53,35 +81,107 @@ def convert_conv1d_to_linear(layer_or_module):
...
 
     def __init__(
         self,
-        module=None,
-        server_remote_address=None,
+        module: Optional[nn.Module] = None,
+        server_remote_address: Optional[str] = None,
+        module_name: Optional[str] = None,
+        model_name: Optional[str] = None,
+        verbose: int = 0,
```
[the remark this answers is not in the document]
andrei-stoian-zama: so we should have an Args section in this docstring

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -188,28 +366,36 @@ def _get_module_by_name(model: nn.Module, name: str) -> Union[RemoteModule, nn.M
...
+            module.init_fhe_client(path_to_client=path_to_client, path_to_keys=path_to_keys)
 
     def compile_model(
         self,
         x: torch.Tensor,
         n_bits: int = 8,
-        rounding_threshold_bits: int = 8,
-        p_error=0.01,
-        configuration: Configuration = None,
+        rounding_threshold_bits: Optional[int] = 8,
```
andrei-stoian-zama, line 396: this seems like a weird default to me.. why is 8 a good value ? do we know it's a good value for LLMs or for any NN in general ?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -1,16 +1,44 @@
...
+import numpy
+import requests
 import torch
 from concrete.fhe import Configuration
 from torch import nn
 from transformers import Conv1D
 
 from ..deployment.fhe_client_server import FHEModelClient, FHEModelDev
-from .compile import compile_torch_model
+from .compile import QuantizedModule, compile_torch_model
```
jfrery, line 20: I realize we have a problem here. It won't work for QAT models from brevitas.

tests/torch/test_compile_torch.py
andrei-stoian-zama: a comment could be useful.

alternatively you can put the `use_conv` as a constant in the pytest parameter list just like `use_qat`

src/concrete/ml/torch/compile.py
jfrery: I see. But don't we have a function that does exactly that as well in CML? Should we keep both?

tests/torch/test_compile_torch.py
jfrery: I was just curious it's weird to see hard coded value change like this. Thanks

use_case_examples/hybrid_model/README.md
jfrery: Got it thanks.

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
```diff
@@ -0,0 +1,89 @@
...
+
+    device = "cpu"
+    if torch.cuda.is_available():
+        device = "cuda"
+    if mps.is_available():
+        device = "mps"
+    print(f"Using device: {device}")
+
+    # Get GPT2 from Huggingface
+    # TODO: migrate to auto-model with model_name
```
andrei-stoian-zama, line 29: can you make an issue on this ?

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: Are you talking about rounding_threshold_bits? Isn't it default to None anyway?

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: Yes blank line ^^

use_case_examples/llm/QGPT2Evaluate.ipynb
jfrery: Not sure. It's not important I guess.

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
```diff
@@ -0,0 +1,89 @@
...
+from pathlib import Path
+
+import torch
+from torch.backends import mps
+from transformers import AutoModelForCausalLM, AutoTokenizer, PreTrainedTokenizer, TextStreamer
+
+from concrete.ml.torch.hybrid_model import FHEMode, HybridFHEModel
+
+if __name__ == "__main__":
+    configs = [
```
andrei-stoian-zama, line 12: can you explain what is in this config structure ?

use_case_examples/hybrid_model/load_and_analyze_data.py
andrei-stoian-zama: yes, do we need it ?

andrei-stoian-zama (changes requested):
Looks good! Some minor changes especially to docs: ex: better docstrings for the arguments to the Hybrid class

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -251,11 +251,14 @@ def load(self):  # pylint: disable=no-value-for-parameter
                 versions = json.load(file)
 
         errors = []
-        packages_to_check = {"concrete-python"}
+        packages_to_check = {"concrete-python", "concrete-ml"}
```
RomanBredehoft, line 254: actually could this section be refactored into a function ? I see that you seem to use the same lines for both the server and client classes

src/concrete/ml/onnx/convert.py
```diff
@@ -34,41 +127,39 @@ def get_equivalent_numpy_forward_and_onnx_model(
...
+    # Export to ONNX
     torch.onnx.export(
         torch_module,
         dummy_input,
         str(output_onnx_file_path),
         opset_version=OPSET_VERSION_FOR_ONNX_EXPORT,
     )
     equivalent_onnx_model = onnx.load_model(str(output_onnx_file_path))
-
-    checker.check_model(equivalent_onnx_model)
```
RomanBredehoft, line 51: why was this removed ?

src/concrete/ml/onnx/convert.py
```diff
@@ -87,14 +178,86 @@ def get_equivalent_numpy_forward(
...
     if check_model:
         checker.check_model(onnx_model)
-    required_onnx_operators = set(get_op_type(node) for node in onnx_model.graph.node)
-    unsupported_operators = required_onnx_operators - IMPLEMENTED_ONNX_OPS
+    checker.check_model(onnx_model)
 
+    # Optimize ONNX graph
+    # List of all currently supported onnx optimizer passes
+    # From https://github.com/onnx/optimizer/blob/master/onnxoptimizer/pass_registry.h
+    # onnx_passes = [
```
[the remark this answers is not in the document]
RomanBredehoft: yes maybe it's not necessary to keep them all here since the relevant link is given

src/concrete/ml/onnx/convert.py
```diff
@@ -87,14 +178,86 @@ def get_equivalent_numpy_forward(
...
+    ]
+    equivalent_onnx_model = onnxoptimizer.optimize(onnx_model, onnx_passes)
+    checker.check_model(equivalent_onnx_model)
+    # Custom optimization
+    # ONNX optimizer does not optimize Mat-Mult + Bias pattern into GEMM if the input isn't a matrix
+    # We manually do the optimization for this case
+    equivalent_onnx_model = fuse_matmul_bias_to_gemm(equivalent_onnx_model)
+    with open("debug.onnx", "wb") as file:
+        file.write(equivalent_onnx_model.SerializeToString())
+    # checker.check_model(equivalent_onnx_model)
```
RomanBredehoft, line 248: to remove ?

src/concrete/ml/quantization/post_training.py
```diff
@@ -788,7 +788,9 @@ def _process_initializer(self, n_bits: int, values: numpy.ndarray):
             QuantizedArray: a quantized tensor with integer values on n_bits bits
         """
 
-        if isinstance(values, numpy.ndarray) and numpy.issubdtype(values.dtype, numpy.integer):
+        if isinstance(values, numpy.ndarray) and numpy.issubdtype(
+            values.dtype, numpy.integer
+        ):  # pragma:no cover
```
RomanBredehoft, line 793: why is this part not used anymore ?

src/concrete/ml/torch/compile.py
```diff
@@ -81,10 +81,7 @@ def build_quantized_module(
         convert_torch_tensor_or_numpy_array_to_numpy_array(val) for val in to_tuple(torch_inputset)
     )
 
-    # Tracing needs to be done with the batch size of 1 since we compile our models to FHE with
-    # this batch size. The input set contains many examples, to determine a representative
-    # bit-width, but for tracing we only take a single one. We need the ONNX tracing batch size to
-    # match the batch size during FHE inference which can only be 1 for the moment.
+    # No batch dimension (i.e., 0 instead of [0]) because else GEMM onnx pass can't be applied
```
RomanBredehoft, line 84: is the gemm pass issue the only reason we have this ? looking at the above comments I'm not so sure

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: good idea

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: what do you mean ?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -1,16 +1,44 @@
...
+
+
+class FHEMode(enum.Enum):
```
RomanBredehoft, line 23: hmm, I don't think it's a good idea to have another `FHEMode` instance here while we already have a similar (but different) one in  `common/utils.py`. Could you either inherit from that latter or just name it differently ?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -1,16 +1,44 @@
...
+def tuple_to_underscore_str(tup: Tuple) -> str:
+    """Convert a tuple to a string representation.
+
+    Args:
+        tup (Tuple): a tuple to change into string representation
+
+    Returns:
+        str: a string representing the tuple
+    """
+    return repr(tup).replace("(", "p_").replace(")", "_p").replace(", ", "_")
```
RomanBredehoft, line 41: why is this needed ? why replacing parenthesis with p_ or _p ?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -53,35 +81,107 @@ def convert_conv1d_to_linear(layer_or_module):
...
         self.uid = str(uuid.uuid4())
-        self.private_q_module = None
-        self.fhe_local_mode = "disable"
-        self.client: Optional[FHEModelClient] = None
-        self.path_to_keys = None
-        self.path_to_client = None
-
-    def init_fhe_client(self, path_to_client: str, path_to_keys: str):
+        self.private_q_module: Optional[QuantizedModule] = None
+        # TODO: figure out if this is good
```
RomanBredehoft, line 102: remove it or make it a FIXME with an issue link !

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -53,35 +81,107 @@ def convert_conv1d_to_linear(layer_or_module):
...
+        self.path_to_clients = path_to_client
+        if self.path_to_clients is None:
+            self.path_to_clients = Path() / "clients"
+        self.path_to_clients.mkdir(exist_ok=True)
+        self.path_to_keys = path_to_keys
+        if self.path_to_keys is None:
+            self.path_to_keys = Path() / "keys"
+        self.path_to_keys.mkdir(exist_ok=True)
+
+        assert self.module_name is not None
```
RomanBredehoft, line 132: could you maybe document a bit more this part ? I'm not sure what is this `shapes_response` + what this whole part does overall

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -91,30 +191,97 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
         Returns:
             (torch.Tensor): The output tensor.
+
+        Raises:
+            ValueError: if fhe_mode is not supported
         """
-        if self.fhe_local_mode != "disable":
-            # for mypy
-            assert self.private_module is not None
+        # - disable: torch module
```
RomanBredehoft, line 198: maybe something to put in the docstring instead of comments (nicer I think)

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -1,16 +1,44 @@
...
 from ..deployment.fhe_client_server import FHEModelClient, FHEModelDev
-from .compile import compile_torch_model
+from .compile import QuantizedModule, compile_torch_model
+
+
+class FHEMode(enum.Enum):
+    """Simple enum for different modes of execution of HybridModel."""
+
+    DISABLE = "disable"  # Use torch weights
+    REMOTE = "remote"  # Use remote FHE server
```
RomanBredehoft, line 27: also the initial FHEMode has "EXECUTE" instead of "REMOTE" so weird to have different conventions

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -91,30 +191,97 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            self.calibration_data.append(x.detach())
             y = self.private_module(x)
+            assert isinstance(y, torch.Tensor)
         # TODO: https://github.com/zama-ai/concrete-ml-internal/issues/3869
-        # else:
-        #     y = self.remote_call(x)
+        elif self.fhe_local_mode == FHEMode.REMOTE:  # pragma:no cover
+            # Remote call
+            y = self.remote_call(x)
+        else:  # pragma:no cover
```
RomanBredehoft, line 226: Better to add a very small test than add pragma: no cover no ?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -91,30 +191,97 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
 
         Args:
             x (torch.Tensor): The input tensor.
+
+        Returns:
+            torch.Tensor: The result of the FHE computation
         """
         # TODO: https://github.com/zama-ai/concrete-ml-internal/issues/3869
         # implement server call and client initialization
+        base_device = x.device
```
RomanBredehoft, line 242: I think the following needs to be documented a bit, not so sure what's going on in some parts

use_case_examples/hybrid_model/compile.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+VIA_MLIR=0
```
RomanBredehoft, line 2: why using a bash script file + `os.environ.get("INDEX", 2)` in the compile.py than use regular argparse ? 😅 I think we should keep it simple

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,140 @@
...
+    module_names: Union[str, List],
+    expected_accuracy,
+    models_dir: Path,
+):
+    """Run the test for any model with its private module names."""
+
+    # Enable multi params/precision
+    configuration = Configuration(
+        single_precision=False,
+        parameter_selection_strategy=ParameterSelectionStrategy.MULTI,
```
RomanBredehoft, line 29: you can remove the MULTI part as it has been made default in CP now

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+if __name__ == "__main__":
+    configs = [
+        ("transformer.h.0.mlp", 0.934),  # Full MLP
+        (["transformer.h.0.mlp", "transformer.h.1.mlp"], 0.42),  # Two full MLPs
+        ("transformer.h.0.mlp.c_proj", 0.986),  # only projection in MLP
+        ("transformer.h.0.attn.c_proj", 0.986),  # only projection in MLP
+    ]
+    config_index = int(os.environ.get("INDEX", 2))
```
RomanBredehoft, line 79: see my comment in compile.sh

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,140 @@
...
+    config = configs[config_index][0]
+    expected_accuracy = configs[config_index][1]
+
+    # Compilation should be done on CPU
+    device = "cpu"
+    print(f"Using device: {device}")
+
+    # Get GPT2 from Huggingface
```
RomanBredehoft, line 87: Hugging Face

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,140 @@
...
+    # In this case we compile for only one sample
+    # We might want to compile for multiple samples
+    # To do this the easiest solution is to compile on contexts of different sizes.
+    # They should all have the same lengths
+    # We might hack something based on HuggingFace dataset with some truncation
+    # Without truncation or selection it would require some knowledge of the tokenizer
+    max_context_size = 20
+    num_samples = 50
+
+    dataset = load_dataset("wikipedia", "20220301.en")
```
RomanBredehoft, line 112: didn't we said it was better to keep it local and avoid any remove loading like this (to avoid any p_error flaky issues like last time) ?

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
RomanBredehoft: also, they are already defined in the compile file : maybe refactor this into a single config file ?

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
```diff
@@ -0,0 +1,89 @@
...
+    config = configs[config_index][0]
+
+    device = "cpu"
+    if torch.cuda.is_available():
+        device = "cuda"
+    if mps.is_available():
+        device = "mps"
+    print(f"Using device: {device}")
+
+    # Get GPT2 from Huggingface
```
RomanBredehoft, line 28: Hugging Face

use_case_examples/hybrid_model/serve.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+uname_str=$(uname)
+echo "${uname_str}"
+if [[ $uname_str != "Darwin" ]]; then
+  echo "Not Darwin"
+  # tune the cpu-list according to the resources you want to allocate to it
+  PATH_TO_MODELS="compiled_models" PORT=8000 taskset --cpu-list 0-12 python serve_model.py
+  # No-limit
+  # PATH_TO_MODELS="compiled_models" PORT=8000 python serve_model.py
```
RomanBredehoft, line 9: to remove ?

use_case_examples/hybrid_model/serve.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
RomanBredehoft, line 1: same comment as for the compile.sh file, why not sure argparse directly into serve_model.py ?

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,193 @@
...
+from fastapi import FastAPI, Form, HTTPException, UploadFile
+from fastapi.responses import FileResponse, StreamingResponse
+from loguru import logger
+
+# No relative import here because when not used in the package itself
+from concrete.ml.deployment import FHEModelServer
+
+
+def underscore_str_to_tuple(tup):
+    return ast.literal_eval(tup.replace("p_", "(").replace("_p", ")").replace("_", ", "))
```
RomanBredehoft, line 30: the same function has been defined in `torch/hybrid_model.py` right ? maybe make  a `common/utils.py` ?

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,193 @@
...
+    app = FastAPI(debug=False)
+    # Model-name -> Module-Name -> Input-shape
+
+    FILE_FOLDER = Path(__file__).parent
+    KEY_PATH = Path(os.environ.get("KEY_PATH", FILE_FOLDER / Path("user_keys")))
+    KEY_PATH.mkdir(exist_ok=True)
+    MODELS_PATH = Path(os.environ.get("PATH_TO_MODELS", FILE_FOLDER / Path("model")))
+    PORT = os.environ.get("PORT", "5000")
+    MODULES = defaultdict(dict)
+    # Populate modules -> could be done dynamically on each query tbh
```
RomanBredehoft, line 43: what is this comment ?

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,193 @@
...
+                    continue
+                input_shape = str(underscore_str_to_tuple(input_shape_path.name))
+                MODULES[model_name][module_name][input_shape] = {
+                    "path": input_shape_path.resolve(),
+                    "module_name": module_name,
+                    "model_name": model_name,
+                    "shape": input_shape,
+                }
+
+    @lru_cache(maxsize=None)
```
RomanBredehoft, line 64: what is this ?

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Returns:
+            Dict[str, str]
+                - uid: uid a personal uid
+        """
+        check_inputs(model_name, module_name, input_shape)
+        uid = str(uuid.uuid4())
+        key_bytes = await key.read()
+        dump_key(key_bytes, uid)
+        # TODO: we should probably store for which circuit the key was generated for
```
RomanBredehoft, line 148: maybe make it a fixme + issue link ?

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,193 @@
...
+        return {"uid": uid}
+
+    @app.post("/compute")
+    async def compute(
+        model_input: UploadFile,
+        uid: str = Form(),
+        model_name: str = Form(),
+        module_name: str = Form(),
+        input_shape: str = Form(),
+    ):  # noqa: B008
```
RomanBredehoft, line 159: what is this noqa ?

RomanBredehoft (changes requested):
huge work here, thanks a lot ! I made several comments, most of them are related to some cleaning / comments to write / fixme to add

src/concrete/ml/torch/hybrid_model.py
jfrery: The hybrid model uses compile_torch_model only so it won't work for brevitas networks.

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
```diff
@@ -0,0 +1,89 @@
...
+        model,
+        config,
+        server_remote_address="http://0.0.0.0:8000",
+        model_name=f"{model_name}_{config_index}",
+        verbose=False,
+    )
+    path_to_clients = Path(__file__).parent / "clients"
+    hybrid_model.init_client(path_to_clients=path_to_clients)
+    for module in hybrid_model.remote_modules.values():
+        module.fhe_local_mode = FHEMode.REMOTE
```
jfrery, line 50: Would be great to add a comment here. Also I guess we could have a function that does this in hybrid model instead of letting users do this loop

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
jfrery: This is done already no? The following lines use the automodel

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: ah indeed, but maybe there's not much more to do for supporting that feature right ?

src/concrete/ml/torch/compile.py
fd0r: `ceinture et bretelles expression` as we say

src/concrete/ml/torch/hybrid_model.py
jfrery: Yes should be straightforward. We can do this in another PR if it requires another CI run and more tests.

src/concrete/ml/torch/compile.py
jfrery: As long as they don't interfere I am fine.

use_case_examples/llm/QGPT2Evaluate.ipynb
fd0r: I didn't modify the file myself

src/concrete/ml/torch/hybrid_model.py
fd0r: Alright with you if we tackle this in another PR?

src/concrete/ml/onnx/convert.py
```diff
@@ -47,7 +47,58 @@ def get_equivalent_numpy_forward_and_onnx_model(
         opset_version=OPSET_VERSION_FOR_ONNX_EXPORT,
     )
     equivalent_onnx_model = onnx.load_model(str(output_onnx_file_path))
-
+    # List of all currently supported onnx passes
```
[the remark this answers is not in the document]
fd0r: Link should be enough, I'll remove

src/concrete/ml/torch/hybrid_model.py
jfrery: Yes sure. Just make a TODO + issue please.

src/concrete/ml/onnx/convert.py
```diff
@@ -2,23 +2,116 @@
...
+    nodes_list = list(onnx_model.graph.node)
+
+    # Iterate through graph nodes
+    for matmul_node in nodes_list:
+        # Run only if the node is a MatMul node
+        if matmul_node.op_type != "MatMul":
+            continue
+        # Store MatMul node output name
+        matmul_node_output_name = matmul_node.output[0]
+        assert len(matmul_node.output) == 1
```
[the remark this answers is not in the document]
fd0r: Yes that is expected, this is more of a guide for someone reading the code than a real assert

src/concrete/ml/torch/hybrid_model.py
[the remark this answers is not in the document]
fd0r: Indeed, I'm surprised this isn't enforced by our docstring checks tbh

src/concrete/ml/torch/hybrid_model.py
fd0r: I'll change this to the normal defaults (`None`) this was here from a previous PR

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
fd0r: I was thinking about it but creating a file just for that seemed a bit overkill.

Can do it if you think it's necessary.
I also thought about dumping a json in the compile file with the configuration that is then re-used in the inference script.
Would that work for you?

src/concrete/ml/deployment/fhe_client_server.py
fd0r: It could indeed, in another PR is that fine with you.

They are already quite a bit of changes here.

src/concrete/ml/onnx/convert.py
fd0r: indeed nice catch

src/concrete/ml/quantization/post_training.py
fd0r: Not sure, I'll remove and we'll see.

src/concrete/ml/torch/hybrid_model.py
fd0r: Yes ofc

src/concrete/ml/torch/hybrid_model.py
fd0r: Indeed I'll rename

src/concrete/ml/torch/hybrid_model.py
fd0r: Makes it easier to parse.

src/concrete/ml/torch/hybrid_model.py
fd0r: I'll add more comments along the way, thanks for the feedback on this!

src/concrete/ml/torch/hybrid_model.py
fd0r: This is more to have a quick reminder close to where this is used in the code but I can add it to the docstring indeed.

src/concrete/ml/torch/hybrid_model.py
fd0r: This can't happen since we always call using `HybridModel` and not `RemoteModule` directly.

src/concrete/ml/torch/hybrid_model.py
fd0r: I'll add inline comments

use_case_examples/hybrid_model/compile.sh
fd0r: I find env vars easier to use than argparse personnaly.

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: I'd rather have it explicit even if default.

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: This is a use case not a test, so not an issue imho.

use_case_examples/hybrid_model/serve.sh
fd0r: I'd rather keep it here so that people can easily un-comment it if they want

use_case_examples/hybrid_model/serve_model.py
fd0r: No it's the reverse function in `hybrid_model.py` 😉

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,193 @@
...
+
+    FILE_FOLDER = Path(__file__).parent
+    KEY_PATH = Path(os.environ.get("KEY_PATH", FILE_FOLDER / Path("user_keys")))
+    KEY_PATH.mkdir(exist_ok=True)
+    MODELS_PATH = Path(os.environ.get("PATH_TO_MODELS", FILE_FOLDER / Path("model")))
+    PORT = os.environ.get("PORT", "5000")
+    MODULES = defaultdict(dict)
+    # Populate modules -> could be done dynamically on each query tbh
+    for model_path in MODELS_PATH.iterdir():  # Model
+        # TODO: change with a struct/obj
```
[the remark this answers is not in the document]
fd0r: I'll remove this line

use_case_examples/hybrid_model/serve_model.py
fd0r: This is a way to cache the result of a function.
Avoids re-computing the function.

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -53,35 +81,107 @@ def convert_conv1d_to_linear(layer_or_module):
...
+                    "model_name": self.model_name,
+                    "input_shape": shape,
+                },
+            )
+            if client_response.status_code != 200:
+                # Add link to request content
+                raise ValueError(
+                    f"Couldn't get client from server:\n{client_response.content.decode('utf-8')}"
+                )
+            path_to_client = self.path_to_clients / tuple_to_underscore_str(ast.literal_eval(shape))
```
RomanBredehoft, line 157: is this `ast.literal_eval(shape)` necessary ? isn't `shape` already some kind of string, which could be directly be given to `tuple_to_underscore_str` as an input ?

use_case_examples/hybrid_model/serve_model.py
RomanBredehoft: ah ok I see my bad

so if I understand correctly, some files (circuits ?) are saved with names based on their shape, using the first function. Then, we load these files and parse their name in order to load back the shapes using `ast.literal_eval`, right ?

but is this safe enough ? are all shapes unique ? And most of all, wasn't there any better solutions than this ? I'm not sure I followed everything but wasn't it possible to have regular namings and retrieve the shapes differently ?

use_case_examples/hybrid_model/serve_model.py
RomanBredehoft: ok didn't know about it ! is it safe to cache keys like this ? I mean here it's just a use case example but maybe not a good practice to show ?

use_case_examples/hybrid_model/serve_model.py
RomanBredehoft: ah yes I see yes FastAPI is weird on that part, thanks

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
RomanBredehoft: I think having a config file is fine but as you wish, second solutions seems ok

in any case as Andrei said a comment explaining what these configs are would be great as well !

src/concrete/ml/deployment/fhe_client_server.py
RomanBredehoft: yes sure : https://github.com/zama-ai/concrete-ml-internal/issues/3993

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: as long as FHEMode is renamed I think it should be fine as "remote" is indeed very clear in the context of Hybrid LLM

use_case_examples/hybrid_model/compile_hybrid_llm.py
RomanBredehoft: don't really agree as I believe we are going more and more towards not exposing strategies (meaning MULTI should be best and thus default, without new users being exposed to it), but I get your point, so if no one else pushed for this it's fine for me

use_case_examples/hybrid_model/serve.sh
RomanBredehoft: could become an ENV var to enable or something like that then, but sure

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: weird

src/concrete/ml/torch/hybrid_model.py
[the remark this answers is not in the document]
RomanBredehoft: I think init are exempted, and we never put args in __init__ (mostly because we didn't want to add init docstrings for built-in models, which would require copy pasting from sklearn)

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: thanks, looks better now !

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: thanks, looks better !

src/concrete/ml/torch/hybrid_model.py
fd0r: Changed it back to defaults

tests/torch/test_compile_torch.py
fd0r: Changed it back to what it was

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Removed the comment, it was outdated

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Added

use_case_examples/hybrid_model/load_and_analyze_data.py
fd0r: Removed the file

use_case_examples/llm/QGPT2Evaluate.ipynb
fd0r: Can't do much about this one

src/concrete/ml/torch/hybrid_model.py
fd0r: 🤷🏼

src/concrete/ml/onnx/convert.py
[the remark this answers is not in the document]
fd0r: I removed and kept only the link to the repository

src/concrete/ml/onnx/convert.py
```diff
@@ -2,23 +2,116 @@
...
+from onnx import checker, helper
 
 from .onnx_utils import IMPLEMENTED_ONNX_OPS, execute_onnx_with_numpy, get_op_type
 
 OPSET_VERSION_FOR_ONNX_EXPORT = 14
 
 
-def get_equivalent_numpy_forward_and_onnx_model(
+# pylint: disable=too-many-branches
+def fuse_matmul_bias_to_gemm(onnx_model: onnx.ModelProto):
```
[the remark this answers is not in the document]
fd0r: Thanks! 🙏🏼

src/concrete/ml/onnx/convert.py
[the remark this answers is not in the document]
fd0r: Removed them

src/concrete/ml/torch/hybrid_model.py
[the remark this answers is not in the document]
fd0r: I added some documentation

src/concrete/ml/torch/hybrid_model.py
fd0r: Changed defaults back to normal

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
fd0r: I removed the comment

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
fd0r: Compilation dumps a json that can then be used by the client

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Fixed it

use_case_examples/hybrid_model/serve.sh
fd0r: Could be indeed but I'd rather have people modifying this file, we don't all have the same number of CPU cores for example

use_case_examples/hybrid_model/serve_model.py
fd0r: Yes for each module we need to have a variety of circuits compiled for each shape.
I found this was the quickest way to get something working.

Safe against what? Well you can't have multiple folders with the same name so yes no issue I believe.

Didn't think of another one, but I am open to suggestions 🙂 
We could have dumped a file on the same and have some sort of unique id but I found it easier to have this information in the folder's name. We can change this afterwards if we find that this creates too many problems.

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,193 @@
...
+        model_name = model_path.name
+        MODULES[model_name] = defaultdict(dict)
```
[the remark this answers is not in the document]
fd0r: Added some comments

use_case_examples/hybrid_model/serve_model.py
fd0r: What would be unsafe about this?

src/concrete/ml/torch/hybrid_model.py
fd0r: https://docs.python.org/3/library/ast.html#ast.literal_eval

No attack possible other than memory exhaustion so pretty safe in our situation imho

use_case_examples/hybrid_model/compile_hybrid_llm.py
RomanBredehoft: still here not ? 😅

use_case_examples/llm/QGPT2Evaluate.ipynb
RomanBredehoft: probably a version difference or something

use_case_examples/hybrid_model/serve_model.py
RomanBredehoft: well not exactly sure I was just thinking that CP's caching key features is considered an unsafe feature , so there might be some reasons

use_case_examples/hybrid_model/serve_model.py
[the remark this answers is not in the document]
RomanBredehoft: thanks, I was more expecting comments for each main steps (like one-line comments above some code lines) but at least we understand a bit more now

use_case_examples/hybrid_model/serve_model.py
andrei-stoian-zama: these are boostrapping keys right ?

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
fd0r: This is actually redundant with what is done in the forward of the HybridModel.

BUT to have all remote module calling the server when using the Module directly and not the hybrid model itself this is needed.

Would you prefer to have a setter for this attribute for the hybrid model that also sets this attribute in the remote modules?

use_case_examples/hybrid_model/serve_model.py
fd0r: These are eval keys.

Unsafe because they are cached in the file-system where things might change, here we are talking in memory caching.

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Humm indeed I forgot to push this, should be fine now

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: I agree with Roman here. I don't expect our users to understand what are these parameters (it's already hard for us to fully understand what they do). And there are more with the comparison options. I think this should all be handled behind the scene as the version of CML is enough to know what config is being used.

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: thanks!

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,119 @@
...
+        parameter_selection_strategy=ParameterSelectionStrategy.MULTI,
+    )
+
+    # Create a hybrid model
+    hybrid_model = HybridFHEModel(model, module_names)
+    hybrid_model.compile_model(
+        inputs,
+        n_bits=8,
+        # We only do linear layers so no need to round
+        rounding_threshold_bits=None,
```
jfrery, line 38: Same here, I don't think any of our users is going to understand what this comment means. Let's just remove it as well as the rounding_bit_threshold perhaps?

fd0r: Well it's documented no?

I can remove if you prefer yes

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Alright then I'll remove everything

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,110 @@
...
+
+if __name__ == "__main__":
+    configs = [
+        "transformer.h.0.mlp",  # Full MLP
+        ["transformer.h.0.mlp", "transformer.h.1.mlp"],  # Two full MLPs
+        "transformer.h.0.mlp.c_proj",  # only projection in MLP
+        "transformer.h.0.attn.c_proj",  # only projection in MLP
+    ]
+    config_index = int(os.environ.get("INDEX", 2))
+    config = configs[config_index]
```
jfrery, line 49: I find this a bit complicated to get the module names from a env variable in this case. I feel like compile.sh is not really useful and we could just launch compule_hybrid_llm.py directly with a arg.

Also may be worth to rename config at this point as it's actually module_names

jfrery: Anything that makes the user a bit more aware about what he launches.
Something like this?

```
python compile_hybrid_llm.py --private_module first_mlp/two_first_mlp/first_linear
```

fd0r: Moved to argparser

use_case_examples/hybrid_model/compile_hybrid_llm.py
fd0r: Moved to argparser

use_case_examples/hybrid_model/serve_model.py
RomanBredehoft: ok sure then, fine for me !

use_case_examples/hybrid_model/serve_model.py
RomanBredehoft: I see, well yes maybe we'll change that later if we think it's not great , thanks for the explanation !

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: ok great, then fine for me !

use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -0,0 +1,123 @@
...
+
+if __name__ == "__main__":
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument(
+        "--module-names",
+        dest="module_names",
+        default=["transformer.h.0.attn.c_proj"],
+        type=module_names_parser,
+        help="""The module(s) name(s) to compile to FHE.
+Examples for GPT-2 model:
```
RomanBredehoft, line 53: weird indent, but great thanks !

use_case_examples/hybrid_model/serve.sh
RomanBredehoft: @fd0r if you removed compile.sh then maybe you could do the same for serve.py ?

use_case_examples/hybrid_model/serve.sh
RomanBredehoft: got it !

RomanBredehoft (dismissed):
I would try to see if it's possible to use argparse in serve.py but if not, than looks good to me ! Huge work here thanks a lot for that

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: I am not sure. Let's keep it we will see.

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: thanks a lot

jfrery (approved):
Looks good to me. I am testing this with phi1.5 to see how it generalizes.

andrei-stoian-zama (approved):

src/concrete/ml/torch/hybrid_model.py
jfrery: thanks

src/concrete/ml/torch/hybrid_model.py
jfrery: I guess it's because the hybridmodel.py import something from concrete.ml.torch? Not sure there is an easy solution then. We leave it like this?

use_case_examples/hybrid_model/requirements.txt
jfrery, line 1: It misses loguru I think?

use_case_examples/hybrid_model/serve_model.py
```diff
@@ -0,0 +1,209 @@
...
+from fastapi import FastAPI, Form, HTTPException, UploadFile
+from fastapi.responses import FileResponse, StreamingResponse
+from loguru import logger
+
+# No relative import here because when not used in the package itself
+from concrete.ml.deployment import FHEModelServer
+
+
+def underscore_str_to_tuple(tup):
+    return ast.literal_eval(tup.replace("p_", "(").replace("_p", ")").replace("_", ", "))
```
jfrery, line 30: This is the only blocking thing while trying to use the use case on phi1.5. Maybe something to improve here.

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
```diff
@@ -0,0 +1,86 @@
...
+    model_name_no_special_char = model_name.replace("/", "_")
+    tokenizer = AutoTokenizer.from_pretrained(model_name)
+    model = AutoModelForCausalLM.from_pretrained(
+        model_name,
+        device_map=device,
+        trust_remote_code=True,
+    )
+
+    # Modify model to use remote FHE server instead of local weights
+    hybrid_model = HybridFHEModel(
```
jfrery, line 37: Why are we rebuilding the hybrid model from scratch here instead of re-using the one computed in compile_hybrid_llm.py?

use_case_examples/hybrid_model/serve_model.py
fd0r: The shape thing doesn't work properly?

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
fd0r: How would you do that?
I had two solutions in mind:
- Just use `torch.load` but that uses pickle and I'm not a big fan of it
- Use the `.pth` file but that would have mean re-creating the HybridModel anyways.

jfrery: I think you can store the .pth along with the state dict and then when you will load the custom remote layer will already be in there. So you can then simply call model().

fd0r:
Fixed @jfrery 's issue and squashed commits.

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
fd0r: You can't get the remote modules without some pickle

jfrery: Yes you can because you have them imported. When torch.load is called it will try to call the RemoteModule. If it's not imported then it will fail otherwise it will use the one we have defined.

jfrery: We can probably do this in another issue

jfrery (approved):
Looks good thanks

andrei-stoian-zama (approved):

PR #247, opened 2023-09-14
chore: only run tests that changed in regular CI if nothing else did

RomanBredehoft:
Basically : 
- improve regular tests so that if a PR only changes specific test files, we only run these tests (as they are all independent). We still run all tests if pytest utils or tests data has been changed
- fix a problem where regular pytest step would not be triggered if the CI has been manually triggered 
- better comments

closes https://github.com/zama-ai/concrete-ml-internal/issues/3893


RomanBredehoft:
the CI was tested in other PRs and everything went well

.github/workflows/continuous-integration.yaml
```diff
@@ -631,24 +631,35 @@ jobs:
...
+              steps.changed-files-in-pr.outcome == 'success' 
+              && (
+                steps.changed-files-in-pr.outputs.src_any_changed == 'true' 
+                || steps.changed-files-in-pr.outputs.tests_utils_any_changed == 'true'
+                || steps.changed-files-in-pr.outputs.dependencies_any_changed == 'true'
+                || steps.changed-files-in-pr.outputs.conftest_any_changed == 'true'
+                || steps.changed-files-in-pr.outputs.makefile_any_changed == 'true'
+              )
+            )
+            || fromJSON(env.IS_WORKFLOW_DISPATCH)
```
RomanBredehoft, line 662: regular tests should be run if CI is manually triggered

.github/workflows/continuous-integration.yaml
```diff
@@ -688,6 +699,24 @@ jobs:
           path: failed_tests_comment.txt
           recreate: true
 
+      # If regular pytest step has been skipped but some changes has been detected in test files, 
+      # meaning there was no other changed impacting our testing suite, we only need to run these 
+      # modified tests
+      # Note that if pytest utils or test data are changed, the regular pytest step should have been
+      # triggered instead
+      - name: PyTest on modified tests only
```
RomanBredehoft, line 707: main feature

.github/workflows/continuous-integration.yaml
jfrery: What do you mean by manually triggered?

jfrery (commented):
For my understanding, here we only rerun tests that have been modified unless something from src has been modified, right?

.github/workflows/continuous-integration.yaml
RomanBredehoft: through GitHub's [Action interface](https://github.com/zama-ai/concrete-ml/actions)

RomanBredehoft: meaning we manually trigger the workflow (it is not triggered automatically when opening a PR or during a weekly release for example)

RomanBredehoft:
> For my understanding, here we only rerun tests that have been modified unless something from src has been modified, right?

yes exactly, more precisely, we rerun tests that have been modified instead of all tests unless one of the following it true : 
       - the current workflow does no take place in a weekly or release CI
       - if the CI has been triggered manually (through GitHub's Action interface)
       - the source code has been changed
       - any tests utils (pytest, data) has been changed
       - any dependency has been updated 
       - conftest.py has been changed
       - Makefile has been changed

which does sometimes happen when we fix some tests or refactor

fd0r (approved):

PR #258, opened 2023-09-20
chore: be sure to have keyring installed

bcm-at-zama:
we have seen that, sometimes, keyring was not available, for no obvious reason. I am not able to say when keyring is or is not installed, so let's be sure it's here, by running installation when it's needed (will not do anything if it's already installed) 

closes #https://github.com/zama-ai/concrete-ml-internal/issues/3950

### DIFF
--- a/script/make_utils/generate_authenticated_pip_urls.sh
+++ b/script/make_utils/generate_authenticated_pip_urls.sh
@@ -20,6 +20,10 @@ SECRET_EXTRA_INDEX_URL="${SECRET_EXTRA_INDEX_URL//\'/}"
 # echo "pip extra-index-url: ${SECRET_EXTRA_INDEX_URL}"
 
 if [[ "${SECRET_EXTRA_INDEX_URL}" != "" ]]; then
+    # Sometimes, for no obvious reason, keyring is not installed, so let's reinstall it for more
+    # reliance
+    poetry run python -m pip install keyring
+
     CRED_JSON="$(python script/make_utils/pip_auth_util.py \
     --get-credentials-for "${SECRET_EXTRA_INDEX_URL}" \
     --check-netrc-first \


RomanBredehoft (changes requested):
keyring is a dev dependency in pyproject.toml, could that be the issue (and thus we should put it as a real dependency) ?

bcm-at-zama:
> keyring is a dev dependency in pyproject.toml, could that be the issue (and thus we should put it as a real dependency) ?

- I would not be against moving it to real dependency
- but, I mean, to `make docker_start` etc, you're already supposed to be a dev, and still, I saw the error
- and remark that we _already_ have some strange keyring install in the makefile

```
setup_env:
	@# The keyring install is to allow pip to fetch credentials for our internal repo if needed
       (...) 
	poetry run python -m pip install keyring
```

So, what should I do? Not sure your solution would make it better, since I am already a dev with a dev setting, and still, I see (sometimes) the problem.

RomanBredehoft:
I'm not too familiar with this but could it be that we need keyring before installing dependencies or something like this ? 

It just feels weird to blindly add lines like this without knowing the issue !

bcm-at-zama:
> It just feels weird to "blindly" add lines like this without knowing the issue !

I agree but I have not a 100% reproducible setting for the issue. And here, I thought adding this was better than having the error popping from time to time.

bcm-at-zama:
@andrei-stoian-zama / @fd0r , what would you do for this issue / PR?

fd0r:
Well it can't hurt to install it 🤷🏼 

If it's already installed nothing will changed, if it isn't it will fix the issue.

fd0r (approved):

RomanBredehoft (approved):
unblocking this PR as I don't have much more to say

bcm-at-zama:
Let's give the final word to @andrei-stoian-zama

andrei-stoian-zama (approved):
Ok, looks good to me

andrei-stoian-zama:
Does it close https://github.com/zama-ai/concrete-ml-internal/issues/3950 ?

bcm-at-zama:
> Does it close [zama-ai/concrete-ml-internal#3950](https://github.com/zama-ai/concrete-ml-internal/issues/3950) ?

yes. I had written 

closes #https://github.com/zama-ai/concrete-ml-internal/issues/3950

but apparently that's not the right syntax, I have to learn it

PR #259, opened 2023-09-20
fix: fix flaky padding test

fd0r:
The issue was that we assumed that none of the values in the test input were mapped to 0 by the quantization.
This assumption was proved to be false.
A simple way to force that no value is set to 0 by the quantizer, learned over [0,1], is to simply set the test input to 1 everywhere.

### DIFF
--- a/tests/torch/test_compile_torch.py
+++ b/tests/torch/test_compile_torch.py
@@ -1174,9 +1174,6 @@ def test_shape_operations_net(
             assert "lookup_table" not in quantized_module.fhe_circuit.mlir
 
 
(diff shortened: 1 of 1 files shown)


RomanBredehoft (approved):
fine for me if we don't lose any generalization in this test, thanks !

andrei-stoian-zama:
image

the right stuff @RomanBredehoft !!

fd0r:
closes https://github.com/zama-ai/concrete-ml-internal/issues/3989

fd0r:
wow yeah that's awesome, thanks @RomanBredehoft for this feature!

PR #261, opened 2023-09-20
feat: expose statuses to compile torch

fd0r:


andrei-stoian-zama (dismissed):

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -598,9 +599,13 @@ def compile(
                 error to a default value.
             verbose (bool): Indicate if compilation information should be printed
                 during compilation. Default to False.
+            statuses (Optional[Dict[str, str]]): same dict as Concrete Python encryption statuses
```
RomanBredehoft, line 602: I would maybe better describe this, I think users should be able to understand what this is without having the need to look at CP !

src/concrete/ml/torch/compile.py
```diff
@@ -142,6 +143,7 @@ def _compile_torch_or_onnx_model(
         global_p_error (Optional[float]): probability of error of the full circuit. In FHE
             simulation `global_p_error` is set to 0
         verbose (bool): whether to show compilation information
+        statuses (Optional[Dict[str, str]]): same dict as Concrete Python encryption statuses
```
RomanBredehoft, line 146: same

src/concrete/ml/torch/compile.py
```diff
@@ -229,6 +233,7 @@ def compile_torch_model(
...
```
RomanBredehoft, line 236: same

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: like `Encryption statuses for all input values, either "clear" or "encrypted" `

tests/quantization/test_quantized_module.py
```diff
@@ -351,6 +351,110 @@ def test_quantized_module_rounding_fhe(model_class, input_shape, default_configu
...
+    torch_fc_model.eval()
+
+    # Create random input
+    numpy_input = numpy.random.uniform(size=input_shape)
+    torch_input = torch.from_numpy(numpy_input).float()
+    assert isinstance(torch_input, torch.Tensor)
+    assert isinstance(numpy_input, numpy.ndarray)
+
+    # Compile with rounding activated
+    with pytest.raises(ValueError):
```
RomanBredehoft, line 368: I would prefer `with pytest.raises(ValueError, match=...)` here, especially since we have two different types of value errors, it's more robust !

tests/quantization/test_quantized_module.py
```diff
@@ -351,6 +351,110 @@ def test_quantized_module_rounding_fhe(model_class, input_shape, default_configu
...
+            False,
+            default_configuration,
+            n_bits=2,
+            p_error=0.01,
+            rounding_threshold_bits=6,
+            statuses={"_onnx__Gemm_0": "encrypted", "test": "encrypted"},
+        )
+
+    # No encrypted value
+    with pytest.raises(RuntimeError):
```
RomanBredehoft, line 421: this is a CP restriction right ? is there a clear (🙂) message associated to it ? If not, maybe worth to add our own one, else maybe add a `match=` here as well

RomanBredehoft (changes requested):
thanks for this ! I have a few comments

tests/quantization/test_quantized_module.py
fd0r: Indeed, I was a bit lazy there!
Will fix this

tests/quantization/test_quantized_module.py
fd0r: If there is no encrypted value CP can't compile indeed.

Their error message is quite explicit, I'll add a `match=""`

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -577,6 +577,7 @@ def compile(
         p_error: Optional[float] = None,
         global_p_error: Optional[float] = None,
         verbose: bool = False,
+        statuses: Optional[Dict[str, str]] = None,
```
jfrery, line 580: I don't feel like statuses is very talkative. Should be referring to the input I think. `input_statuses` or `input_encryption` where the user passes booleans in a list following the order of the input in the forward call? so that you don't have to specify onnx node names.

jfrery (commented):
This could be needed indeed thanks!

fd0r:
Looks like KNN test is flaky

jfrery (approved):
Looks great thanks!

fd0r:
What I did:
- changed the name of the variable to a more explicit naming.
- updated the doc-strings to something clearer.
- added `match=` to the tests
- changed the type of the variable from a dict to a list to avoid having to handle onnx names

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -598,9 +599,15 @@ def compile(
                 error to a default value.
             verbose (bool): Indicate if compilation information should be printed
                 during compilation. Default to False.
+            inputs_encryption_status (Optional[Sequence[str]]): encryption status ('clear',
```
RomanBredehoft, line 602: I would still mention that everything is encrypted if the parameter is set to None !

src/concrete/ml/torch/compile.py
```diff
@@ -142,6 +143,8 @@ def _compile_torch_or_onnx_model(
         global_p_error (Optional[float]): probability of error of the full circuit. In FHE
             simulation `global_p_error` is set to 0
         verbose (bool): whether to show compilation information
+        inputs_encryption_status (Optional[Sequence[str]]): encryption status ('clear', 'encrypted')
```
RomanBredehoft, line 146: same

tests/quantization/test_quantized_module.py
```diff
@@ -351,6 +355,103 @@ def test_quantized_module_rounding_fhe(model_class, input_shape, default_configu
     # TODO: https://github.com/zama-ai/concrete-ml-internal/issues/3800
 
 
+# TODO: extend this test with multi-input encryption status
```
RomanBredehoft, line 358: worth a fimxe + link to an issue here right ?

RomanBredehoft (changes requested):
I still have a few comments

fd0r:
Adressed the feedback made by @RomanBredehoft

fd0r:
KNN flaky, relaunching

jfrery (approved):

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -598,9 +599,15 @@ def compile(
...
+                'encrypted') for each input.
 
         Returns:
             Circuit: The compiled Circuit.
+
+        Raises:
+            ValueError: if inputs_encryption_status does not match with the
```
RomanBredehoft, line 609: it's a bit more than that right ?

src/concrete/ml/quantization/quantized_module.py
fd0r: It's mentionned in all the user facing APIs, I don't think many users will use this one tbh

src/concrete/ml/quantization/quantized_module.py
fd0r: Well it's if the configuration is wrong, how would you formulate it?

src/concrete/ml/torch/compile.py
fd0r: Same what? There is a comment on the fact that by default everything is encrypted

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: I mean it was mostly for making the docstring consistent with what this parameters does, as we now provide a way to chose what inputs to encrypt, to avoid confusion that we don't chose which ones are encrypted or not based on whatever other rules. But sure if no one else pushes for this then go for it

src/concrete/ml/torch/compile.py
RomanBredehoft: same as in "same comment as above" in order to not avoid making changes in both places if there are some

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: I would usually very quickly describe what are bad configurations

RomanBredehoft (approved):
if no one pushes for these last comments, fine for me, thanks a lot for this feature !

PR #263, opened 2023-09-20
feat: enable import of fitted linear sklearn models

RomanBredehoft:
Several things have been done here : 
- it's now possible to import fitted **linear** sklearn models in Concrete ML through the `.from_sklearn_model` method (name to be confirmed)
- test suite has been improved : 
     - it's easier to filter in/out models for tests (only consider linear models, exclude "DecistionTree" models, only consider "XGBClassifier", ect), instead of having to put "if ... then: return" (which made pytest and debuggers still display irrelevant test configurations)
     - `sklearn_models_and_datasets` has been replaced by `MODELS_AND_DATASETS` (more relevant)
     - `UNIQUE_MODELS_AND_DATASETS` replaced `get_random_extract_of_sklearn_models_and_datasets` (the function's name was misleading, as nothing was random + it's now easier to use)
     -  some tests run less configurations (by using `UNIQUE_MODELS_AND_DATASETS` instead of `MODELS_AND_DATASETS` , which considers a single config for each models, since some tests do not depend on the models' different init params). I might have missed more tests where this could be done as well (closing https://github.com/zama-ai/concrete-ml-internal/issues/3365 then)
     - fixed `_filter_models` which did not filter correctly by names : this lead to fixing the dump_onnx test for DT and XGB models
     - made all "get_slearkn_XXX_models" functions private

I did not expected this refactor to be that big, as I initially only wanted a way to easily get linear models only for my test. It probably would have been better to merge it in a dedicated PR before implementing the import feature. Anyway, I think it's still fine as the refactor mostly affects tests.

Also, we might want to add a small documentation about the new feature, so let me know what you think !

refs https://github.com/zama-ai/concrete-ml-internal/issues/4020
closes https://github.com/zama-ai/concrete-ml-internal/issues/3809
closes https://github.com/zama-ai/concrete-ml-internal/issues/3365

### DIFF
--- a/src/concrete/ml/sklearn/__init__.py
+++ b/src/concrete/ml/sklearn/__init__.py
@@ -1,8 +1,12 @@
 """Import sklearn models."""
-from typing import List
+from typing import Dict, List, Optional, Union
 
 from ..common.debugging.custom_assert import assert_true
-from ..common.utils import is_classifier_or_partial_classifier, is_regressor_or_partial_regressor
+from ..common.utils import (
+    get_model_name,
+    is_classifier_or_partial_classifier,
+    is_regressor_or_partial_regressor,
+)
 from .base import (
     _ALL_SKLEARN_MODELS,
     _LINEAR_MODELS,
@@ -20,11 +24,11 @@ from .tree import DecisionTreeClassifier, DecisionTreeRegressor
 from .xgb import XGBClassifier, XGBRegressor
 
 
-def get_sklearn_models():
-    """Return the list of available models in Concrete ML.
+def get_sklearn_models() -> Dict[str, List]:
+    """Return the list of available scikit-learn models in Concrete ML.
 
     Returns:
-        the lists of models in Concrete ML
+        sklearn_models (Dict[str, List]): The lists of scikit-learn models available in Concrete ML.
     """
 
     # Import anything in sklearn, just to force the import, to populate _ALL_SKLEARN_MODELS list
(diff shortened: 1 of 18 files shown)


src/concrete/ml/sklearn/base.py
```diff
@@ -1482,6 +1490,54 @@ def __init__(self, n_bits: Union[int, Dict[str, int]] = 8):
 
         BaseEstimator.__init__(self)
 
+    @classmethod
+    def from_sklearn_model(
```
RomanBredehoft, line 1494: this is the new feature

tests/sklearn/test_sklearn_models.py
```diff
@@ -1036,7 +1019,47 @@ def check_mono_parameter_warnings(model, x, default_configuration):
         model.compile(x, default_configuration)
 
 
-@pytest.mark.parametrize("model_class, parameters", sklearn_models_and_datasets)
+def check_load_fitted_sklearn_linear_models(model_class, n_bits, x, y):
```
RomanBredehoft, line 1022: the new feature's test

src/concrete/ml/sklearn/__init__.py
```diff
@@ -26,100 +30,173 @@ def get_sklearn_models():
...
+
+    # Return a sorted list in order to avoid issues when used in @pytest.mark.parametrize
+    return sorted(selected_models, key=lambda m: m.__name__)
 
 
 def get_sklearn_linear_models(
-    classifier: bool = True, regressor: bool = True, str_in_class_name: List[str] = None
+    classifier: bool = True,
+    regressor: bool = True,
+    str_in_class_name: Optional[Union[str, List[str]]] = None,
```
RomanBredehoft, line 97: I would have preferred to put `include` and `exclude` (as I've done for the new or private functions) but I did not because that would make this an API breaking change (+ it's not only used in our test suite)

fd0r: Do people really use these methods?

Maybe we should be more clear between what's private and what's not because I honestly don't see any user using these methods tbh

src/concrete/ml/sklearn/base.py
fd0r: That's awesome!

fd0r (approved):
The API design looks good to me.

Let's avoid doing these refactorings in features PRs though (I know I have a tendency to do the same)

A great addition would be a notebook showing this feature in the documentation and comparing the float model with the quantized one, and with one trained via Concrete ML.

RomanBredehoft:
> Let's avoid doing these refactorings in feature PRs though (I know I have a tendency to do the same)

I started doing this because the test was only made for linear models, but we didn't have a way to properly only select linear models for it 😅 

> A great addition would be a notebook showing this feature in the documentation and comparing the float model with the quantized one, and with one trained via Concrete ML.

not sure about the notebook, normally importing a scikit learn model should give the same results as when we init a cml model + fit with the same parameters, so comparing quantized vs float would be doing exactly the same thing as our other notebooks

but yes as I mentioned, if everyone is fine with that I'm planning on adding a small documentation page on that feature right after !

fd0r:
> not sure about the notebook, normally importing a scikit learn model should give the same results as when we init a cml model + fit with the same parameters, so comparing quantized vs float would be doing exactly the same thing as our other notebooks

still a good thing to showcase imho (but in another PR for sure)

andrei-stoian-zama (dismissed):

src/concrete/ml/sklearn/__init__.py
RomanBredehoft: > really? where is it used elsewhere?

in our serialization's decoder, because we need to provide a list of supported/accepted models from CML !

I kinda of agree with you @fd0r and would say it's fine if I make this change, but I'd prefer to be sure if it's fine with you first !

tests/sklearn/test_dump_onnx.py
```diff
@@ -216,16 +217,12 @@ def test_dump(
 ) {
   %/_operators.0/Gemm_output_0 = Gemm[alpha = 1, beta = 0, transB = 1](%_operators.0.weight_1, %input_0)
```
RomanBredehoft, line 218: the following is directly linked to the fact that the `str_in_name` from 
```
    if not is_model_class_in_a_list(
        model_class, get_sklearn_tree_models(str_in_class_name="RandomForest")
    ):
```

was not working, and thus made this test skip all tree-based models (instead of just RF models). Not sure when these ONNX changes appeared, but it could come from a long time ago

tests/sklearn/test_sklearn_models.py
```diff
@@ -1055,7 +1038,49 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+
+# Neural network models are skipped for this test
+# The `fit_benchmark` function of QNNs returns a QAT model and a FP32 model that is similar
+# in structure but trained from scratch. Furthermore, the `n_bits` setting
+# of the QNN instantiation in `instantiate_model_generic` takes `n_bits` as
+# a target accumulator and sets 3-b w&a for these tests. Thus it's
+# impossible to reach R-2 of 0.99 when comparing the two NN models returned by `fit_benchmark`
+@pytest.mark.parametrize(
+    "model_class, parameters",
+    get_sklearn_linear_models_and_datasets()
```
RomanBredehoft, line 1080: could be replaced by `get_sklearn_all_models_and_datasets(exclude="NeuralNet")`, not sure what's best

tests/sklearn/test_sklearn_models.py
```diff
@@ -438,11 +443,6 @@ def check_offset(model_class, n_bits, x, y):
     """Check offset."""
     model = instantiate_model_generic(model_class, n_bits=n_bits)
 
-    # Offsets are not supported by XGBoost
-    if is_model_class_in_a_list(model_class, get_sklearn_tree_models(str_in_class_name="XGB")):
```
RomanBredehoft, line 442: all these removals (2 or 3 more below) are moved to where we initialize the tests (in mark.parametrize)

src/concrete/ml/sklearn/__init__.py
fd0r: Fine with me to break this API without changing a major.
But it would probably be best to use the `_` prefix to signal to the user that the function is private.

They can still use if they want to do but then we properly signal that this API can change.

fd0r (approved):

src/concrete/ml/sklearn/__init__.py
RomanBredehoft: ok got it, I'll make them private then !

RomanBredehoft: done

fd0r (approved):

src/concrete/ml/sklearn/base.py
jfrery: In BaseEstimator so all models have access to this feature?

andrei-stoian-zama (dismissed):

jfrery (commented):
Thanks for this. Looks like the feature related to this PR is pretty straightforward but there is a huge refacto around it which I am not sure how it is related to the feature. I would probably have had 2 PRs here.

src/concrete/ml/sklearn/base.py
RomanBredehoft: tree-based and QNN models both need re-fitting with quantizers, so no it's only a feature for linear models for noow

src/concrete/ml/sklearn/base.py
```diff
@@ -1464,6 +1472,54 @@ def __init__(self, n_bits: Union[int, Dict[str, int]] = 8):
...
+        # classes during initialization (for example: link or power attributes in GLMs)
+        # Without it, these attributes will have default values instead of the ones used by the
+        # scikit-learn models
+        # This should be fixed once Concrete ML models initialize their underlying scikit-learn
+        # models during initialization
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3373
+        model = cls(n_bits=n_bits, **init_params)
+
+        # Update the underlying scikit-learn model with the given fitted one
+        model.sklearn_model = sklearn_model
```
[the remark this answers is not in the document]
RomanBredehoft: yes indeed thanks !

jfrery (dismissed):

jfrery (dismissed):

fd0r (dismissed):

PR #269, opened 2023-09-21
chore: fix dependency vulnerabilities

RomanBredehoft:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4006


pyproject.toml
```diff
@@ -54,15 +54,13 @@ protobuf = "3.20.3"
...
 # Deployment
 boto3 = "^1.23.5"
-fastapi = "^0.93.0"
+fastapi = "^0.103.1"
 uvicorn = "^0.21.0"
-python-multipart = "^0.0.6"
 tqdm = "^4.64.1"
 transformers = "^4.30.2"
 certifi = "2023.07.22"
-cryptography = "41.0.2"
```
RomanBredehoft, line 63: @fd0r do you confirm that this is not useful anymore ?

pyproject.toml
```diff
@@ -100,8 +98,6 @@ LinkChecker = "^10.1.0"
 kaggle = "^1.5.12"
 tf2onnx = "^1.10.1"
 boto3 = "^1.23.5"
-paramiko = "^2.11.0"
```
RomanBredehoft, line 103: same for paramiko ?

bcm-at-zama:
Is there a tool which tells you when a dependency is actually not needed? Envoyé de mon iPhoneLe 21 sept. 2023 à 18:55, Roman ***@***.***> a écrit :﻿
@RomanBredehoft commented on this pull request.



In pyproject.toml:
> @@ -100,8 +98,6 @@ LinkChecker = "^10.1.0"
 kaggle = "^1.5.12"
 tf2onnx = "^1.10.1"
 boto3 = "^1.23.5"
-paramiko = "^2.11.0"

same for paramiko ?

pyproject.toml
fd0r: This was set to avoid a vulnerability, that means that the package that required it updated the version on their side, that good!

pyproject.toml
fd0r: We don't need paramiko since some time now, not sure what it was doing here.

fd0r (approved):

fd0r (dismissed):

fd0r (approved):

RomanBredehoft:
@bcm-at-zama poetry does not seem to provide such a tool but looks like this package does : https://github.com/fpgmaas/deptry !

fd0r (approved):

jfrery (approved):
Thanks

PR #270, opened 2023-09-22
chore: remove mnist use case

jfrery:
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3353


fd0r (approved):
Not sure why but fine with me

README.md
```diff
@@ -126,8 +126,6 @@ Various tutorials are proposed for the [built-in models](docs/built-in-models/ml
 - [Health diagnosis](use_case_examples/disease_prediction/): based on a patient's symptoms, history and other health factors, gives
   a diagnosis using FHE to preserve the privacy of the patient.
 
-- [MNIST](use_case_examples/mnist): a Python script and notebook showing quantization-aware training following FHE constraints. A fully-connected neural network is implemented with [Brevitas](https://github.com/Xilinx/brevitas) and is converted to FHE with Concrete ML.
```
bcm-at-zama, line 129: nothing to replace with? eg, in our presentations, we clearly say MNIST 97%, it would be nice that people can repro easily

jfrery: We have it in our example. I will push a 98% version of mnist that runs in 20 seconds. This is going to be 20 seconds better than the current CNN mnist but also 6% accuracy better.

bcm-at-zama: great, so can we keep a MNIST entry in the README, and point it there?

fd0r (approved):

README.md
jfrery: Those only target the use case folder but yes maybe we can just redirect to the advanced example for mnist. WDYT @andrei-stoian-zama ?

RomanBredehoft (approved):
some remaining problems with docs but otherwise fine for me !

jfrery:
along with the new CNN version for mnist -> https://github.com/zama-ai/concrete-ml/pull/276

fd0r (dismissed):

README.md
jfrery: What should we do about this @andrei-stoian-zama ? Should I add it back in https://github.com/zama-ai/concrete-ml/pull/276? or we remove mnist completely from the doc as well.

RomanBredehoft (approved):

README.md
andrei-stoian-zama: I think we remove this one. For custom networks there is CIFAR

jfrery: Alright let's merge

andrei-stoian-zama (approved):

kcelia (approved):
Thanks

fd0r (approved):

RomanBredehoft (approved):

PR #273, opened 2023-09-22
chore: flag test_client_server as flaky because of KNN

RomanBredehoft:
refs https://github.com/zama-ai/concrete-ml-internal/issues/4009


fd0r (approved):

andrei-stoian-zama (changes requested):
Why not fix the KNN test if it is actually flaky ?

fd0r:
@kcelia is working on it, in the meantime it's better to flag it as flaky

RomanBredehoft:
yes @andrei-stoian-zama the flaky seems to be pretty recurrent, so unless it can be fixed in 1 hour, I suggest we flag it in the mean time (allows the CI to re-run the test once if it stumbles on it) ! @kcelia will remove it with her fix

jfrery (approved):
Agreed but could you add a ref in the issue supposed to fix this so that we unmark it? Maybe a fixme here too

fd0r (approved):

andrei-stoian-zama (approved):

PR #275, opened 2023-09-22
feat: support qat models in hybrid model

fd0r:
closes https://github.com/zama-ai/concrete-ml-internal/issues/3992

### DIFF
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1387,3 +1387,25 @@ class ConcatFancyIndexing(nn.Module):
         x = torch.relu(self.fc2(x))
         x = self.fc4(self.quant_3(x))
         return x
+
+
+class PartialQATModel(torch.nn.Module):
+    """A model with a QAT Module."""
+
+    def __init__(self, input_shape: int, output_shape: int, n_bits: int):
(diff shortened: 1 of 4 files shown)


jfrery (dismissed):
Nice update. Thanks!

andrei-stoian-zama (approved):
good idea! a unified compilation interface for both qat & ptq with automatic detection of qat.

src/concrete/ml/pytest/torch_models.py
```diff
@@ -1387,3 +1387,25 @@ def forward(self, x):
         x = torch.relu(self.fc2(x))
         x = self.fc4(self.quant_3(x))
         return x
+
+
+class PartialQATModel(torch.nn.Module):
```
RomanBredehoft, line 1392: more like `HybridQatMOdel` right ?

fd0r: Well it depends, the idea here is that you can have non-qat module afterwards and only compile the `sub_module` that uses qat and everything will go fine.

RomanBredehoft: ah got it thanks !

RomanBredehoft (approved):
great thanks for this !

PR #287, opened 2023-09-25
chore: fix server to client transfer time + add Q K V projection

jfrery:
time per token down from ~3.5s down to 0.4s
 
closes https://github.com/zama-ai/concrete-ml-internal/issues/4005

### DIFF
--- a/use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
+++ b/use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
@@ -39,7 +39,7 @@ if __name__ == "__main__":
         module_names,
         server_remote_address="http://0.0.0.0:8000",
         model_name=f"{model_name}",
-        verbose=False,
+        verbose=True,
     )
     path_to_clients = Path(__file__).parent / "clients"
     hybrid_model.init_client(path_to_clients=path_to_clients)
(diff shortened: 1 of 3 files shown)


use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
```diff
@@ -39,7 +39,7 @@
         module_names,
         server_remote_address="http://0.0.0.0:8000",
         model_name=f"{model_name}",
-        verbose=False,
+        verbose=True,
```
fd0r, line 42: I wouldn't put it as default as it breaks the streaming decoding.

fd0r (approved):

use_case_examples/hybrid_model/infer_hybrid_llm_generate.py
jfrery: ah I didn't realize that. I will revert.

fd0r: (I mean it works but it's just lost in the flow of stdout)

fd0r (approved):

RomanBredehoft (approved):
thanks !

PR #289, opened 2023-09-25
chore: factor package checking code in fhe_client_server.py

RomanBredehoft:
This factorization makes the client.load method check for the python version as well 

closes https://github.com/zama-ai/concrete-ml-internal/issues/3993

### DIFF
--- a/src/concrete/ml/deployment/fhe_client_server.py
+++ b/src/concrete/ml/deployment/fhe_client_server.py
@@ -25,6 +25,59 @@ except ImportError:  # pragma: no cover
     from importlib_metadata import version
 
 
+def check_concrete_versions(zip_path: Path):
+    """Check that current versions match the ones used in development.
+
+    This function loads the version JSON file found in client.zip or server.zip files and then
+    checks that current package versions (Concrete Python, Concrete ML) as well as the Python
+    current version all match the ones that are currently installed.
+
+    Args:
+        zip_path (Path): The path to the client or server zip file that contains the version.json
+            file to check.
+
+    Raises:
+        ValueError: If at least one version mismatch is found.
+    """
+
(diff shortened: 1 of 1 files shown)


src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -238,34 +266,17 @@ def load(self):  # pylint: disable=no-value-for-parameter
...
-        if errors:  # pragma: no cover
-            raise ValueError(
-                "Version mismatch for packages: \n"
-                + "\n".join(f"{error[0]}: {error[1]} != {error[2]}" for error in errors)
-            )
+        # Load and check versions
+        check_concrete_versions(client_zip_path)
 
         # Make sure the version in serialized_model is the same as CML_VERSION
         assert_true(
```
RomanBredehoft, line 282: is the following code still necessary ? it seems like Concrete ml's version is also tracked by the `serialized_processing` object, but it looks like it's redundant : I imagine this is kept  for backward compatibility ? @fd0r

jfrery (approved):
Looks good to me thanks!

src/concrete/ml/deployment/fhe_client_server.py
fd0r: Hum I believe we should be able to remove this now.
Could be done in the PR if okay with you

fd0r (changes requested):
I agree with you about removing the other check on CML version

src/concrete/ml/deployment/fhe_client_server.py
RomanBredehoft: yes done !

fd0r (approved):

jfrery (approved):
Looks great thanks!

PR #290, opened 2023-09-25
chore: update ptest-xdist package

RomanBredehoft:
This removes the `py` dependency, which has a high risk vulnerability : https://github.com/zama-ai/concrete-ml/security/dependabot/8


fd0r (dismissed):
Great! Thanks for that

pyproject.toml
```diff
@@ -135,6 +135,7 @@ filterwarnings = [
     "ignore:Implicitly cleaning up:ResourceWarning",
     "ignore:non-integer arguments to randrange\\(\\) have been deprecated since Python 3\\.10 and will be removed in a subsequent version:DeprecationWarning",
     "ignore:Deprecated call to `pkg_resources.declare_namespace:DeprecationWarning",
+    "ignore:The --rsyncdir command line argument and rsyncdirs config variable are deprecated.:DeprecationWarning",
```
RomanBredehoft, line 138: There is a non-breaking compatibility problem between pytest-cov and pytest-xdist 3.x : https://github.com/pytest-dev/pytest-cov/issues/557 (closed but not fixed) , so we need to filter out this warning

andrei-stoian-zama (approved):

PR #300, opened 2023-09-27
chore: add phi 1.5 to hybrid use case

jfrery:


use_case_examples/hybrid_model/compile_hybrid_llm.py
```diff
@@ -43,7 +43,13 @@ def module_names_parser(string: str) -> List[str]:
 
 
 if __name__ == "__main__":
-    arg_parser = argparse.ArgumentParser()
+    arg_parser = argparse.ArgumentParser(description="Showcase for the hybrid model converter.")
+    arg_parser.add_argument(
+        "--model-name",
+        default="gpt2",
+        type=str,
+        help="The name of the model to compile. Default is 'gpt2'.",
```
andrei-stoian-zama, line 51: Can you list the options the user has (which LLMs have we tried?) cc @fd0r

use_case_examples/hybrid_model/README.md
```diff
@@ -12,3 +12,13 @@ In this case we apply a fully connected layer of a GPT-2 model in FHE.
 1. Run FHE client using `python infer_hybrid_llm_generate.py`
    - You will first be asked about the number of tokens that you want to generate
    - Then you will be able to enter your prompt
+
+## Adapt to any LLM
+
+This use case can easily be used with a different model than GPT2 with any other model on Hugging Face.
+
+For example, you can use Phi 1.5 from microsoft and protect the weights that are used to compute Query, Keys, Values as follows:
```
andrei-stoian-zama, line 20: Maybe not mention Microsoft but just give a link to Phi 1.5 on HF or whatever side hosts it ?

andrei-stoian-zama (changes requested):

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: We can't really have a list of all possibility here. Any model would work as long as the module to be converted is compatible.

use_case_examples/hybrid_model/README.md
jfrery: I gave a link to microsoft HF phi-1.5 model.

fd0r (approved):
Looks good to me, thanks for this @jfrery

fd0r (dismissed):

fd0r (approved):

use_case_examples/hybrid_model/compile_hybrid_llm.py
andrei-stoian-zama: But it could just  be the models that you tried..

RomanBredehoft (approved):
thanks !

use_case_examples/hybrid_model/compile_hybrid_llm.py
jfrery: We have a specific section in the readme about this with phi1.5 as an example. Adding more models here would also mean having there module names so it would make this docstring quite confusing. I would keep it with gpt2 and have the readme detail that you can use any kind of model.

PR #305, opened 2023-09-28
fix: fix aws region in deployment notebook

fd0r:


jfrery:
Took the commit in #304

PR #306, opened 2023-09-28
chore: improve inference tests (decision_function, predict, predict_proba) + post_processing 

RomanBredehoft:
Initially this PR was for taking care of improving tests related to post_processing and deployment (https://github.com/zama-ai/concrete-ml-internal/issues/3996). However, while working on it, I realized that it implied more general changes in our test suite. basically, here are the problems I found out : 
- `post_processing` should not be tested in deployment tests but rather while testing our API equivalence (ie, by testing that circuit + `post_processing` = `predict` (for regressors and KNNclassifier) or `predict_proba` (other classifiers)
- `predict_proba` and `decision_function` were not properly tested (some issues were already opened on that: https://github.com/zama-ai/concrete-ml-internal/issues/3344, 
- we have some shape incoherence with regressors compared to sklearn : https://github.com/zama-ai/concrete-ml-internal/issues/4029 (maybe related to https://github.com/zama-ai/concrete-ml-internal/issues/3312)
- we had two tests that were essentially doing the same thing (`test_quantization` and `test_correctness_with_sklearn`, both testing correctness with sklearn but in different ways)
- `test_predict_correctness` looked much more complicated than it should have. Consequence : we were testing weird things like `predict(x)` vs `predict(x, fhe="disable")`
- some correctness tests were testing irrelevant things which led to issues (https://github.com/zama-ai/concrete-ml-internal/issues/3661, https://github.com/zama-ai/concrete-ml-internal/issues/2450)

So I took the time to refactor some test to make them more robust, readable and relevant. See in following comments the main changes !

refs https://github.com/zama-ai/concrete-ml-internal/issues/4029
refs https://github.com/zama-ai/concrete-ml-internal/issues/3962
refs https://github.com/zama-ai/concrete-ml-internal/issues/3045
refs https://github.com/zama-ai/concrete-ml-internal/issues/3312
closes https://github.com/zama-ai/concrete-ml-internal/issues/3996
closes https://github.com/zama-ai/concrete-ml-internal/issues/3344
closes https://github.com/zama-ai/concrete-ml-internal/issues/2450
closes https://github.com/zama-ai/concrete-ml-internal/issues/3661


conftest.py
```diff
@@ -288,41 +288,40 @@ def check_circuit_precision():
...
 
-def check_array_equality_impl(actual: Any, expected: Any, verbose: bool = True):
-    """Assert that `actual` is equal to `expected`."""
-
-    assert numpy.array_equal(actual, expected), (
-        ""
-        if not verbose
-        else f"""
+@pytest.fixture
+def check_array_equal():
```
RomanBredehoft, line 292: just a rename to be coherent with `numpy.array_equal` method (we should probably replace all `numpy.array_equal` with `check_array_equal` by the way)

conftest.py
```diff
@@ -288,41 +288,40 @@ def check_circuit_precision():
...
+            """
+        )
 
-    return check_array_equality_impl
+    return check_array_equal_impl
 
 
 @pytest.fixture
-def check_float_arrays_equal():
+def check_float_array_equal():
```
RomanBredehoft, line 318: just a rename to make it coherent with `check_array_equal`

src/concrete/ml/sklearn/base.py
```diff
@@ -2093,11 +2093,11 @@ def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.
 
         topk_labels = []
         for query in X:
```
RomanBredehoft, line 2095: by the way @kcelia I have a question on this : why can't we use `.predict(self, X, fhe)`here ? Because here we basically loop and run `predict` on queries with shape (1, n) but we handle (m,n) shapes ! I tried to change it to make this simpler but found some issues when running with `fhe="disable`, does that mean the `_inference` method was only made for (1, n) samples ?

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -123,3 +124,26 @@ def load_dict(cls, metadata: Dict):
         obj.metric_params = metadata["metric_params"]
         obj.n_jobs = metadata["n_jobs"]
         return obj
+
+    # KNeighborsClassifier does not provide a predict_proba method for now
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3962
+    def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
```
RomanBredehoft, line 130: better to raise a clear error while it's not implemented, and then have a test on it (similar to NN regressor)

tests/deployment/test_client_server.py
```diff
@@ -97,172 +99,192 @@ def test_client_server_sklearn(
...
-    n_bits = 2
 
-    # Get the quantized module from the model
+    # Get the quantized module from the model and compile it
     quantized_numpy_module = compile_torch_model(
         torch_model,
         x_train,
         configuration=default_configuration,
-        n_bits=n_bits,
-        global_p_error=1 / 100_000,
```
RomanBredehoft, line 146: I thought we wanted to limit the use of global_p_error, why do we have this here ?

tests/deployment/test_client_server.py
```diff
@@ -97,172 +99,192 @@ def test_client_server_sklearn(
...
-    # No user is expected to load a FHEModelDev instance from a FHEModelClient's model. This is
-    # only made a testing for making sure the model has the expected attributes
-    client_model = fhemodel_client.model
-    client_model.fhe_circuit = model.fhe_circuit
 
-    # pylint: disable-next=protected-access
-    client_model._is_compiled = True
-
-    network.cleanup()
+def check_client_server_files(model):
```
RomanBredehoft, line 174: improved the client-server test by having a dedicated test to make sure the clinet/server files are properly generated

tests/deployment/test_client_server.py
```diff
@@ -97,172 +99,192 @@ def test_client_server_sklearn(
...
+    """Test the client server interface API generates the expected file.
 
+    This test expects that the given model has been trained and compiled in development.
+    """
     # Create a new network
-    network = OnDiskNetwork()
+    disk_network = OnDiskNetwork()
 
     # And try to save it again
-    fhemodel_dev_ = FHEModelDev(path_dir=network.dev_dir.name, model=client_model)
```
RomanBredehoft, line 208: I basically spitted the initial test in two tests : 
- client/server file generation test
- predict result tests

It was not clear to me why we were loading two `FHEModelDev` objects using `client_model` in the first place. Comments were saying that this was for making sure the client's model had all necessary attributes, but I believe that checking the file generation + computing the predictions is more than enough (especially since it required this test to manually transfer attributes between instances, something that should never happen)

tests/sklearn/test_sklearn_models.py
```diff
@@ -198,33 +186,79 @@ def check_correctness_with_sklearn(
...
+        if is_model_class_in_a_list(model, _get_sklearn_linear_models()):
+
+            # Check outputs from the 'decision_function' method (for linear classifiers)
+            y_scores_sklearn = sklearn_model.decision_function(x)
+            y_scores_fhe = model.decision_function(x, fhe=fhe)
+
+            # Currently, for single target data sets, Concrete models' outputs have shape (n, 1)
+            # while scikit-learn models' outputs have shape (n, )
+            # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4029
+            # assert y_scores_sklearn.shape == y_scores_fhe.shape, (
```
RomanBredehoft, line 220: -> https://github.com/zama-ai/concrete-ml-internal/issues/4029

tests/sklearn/test_sklearn_models.py
```diff
@@ -198,33 +186,79 @@ def check_correctness_with_sklearn(
...
     model_name = get_model_name(model_class)
-    acceptance_r2score = acceptance_r2score_dic.get(model_name, 0.9)
-    threshold_accuracy = threshold_accuracy_dic.get(model_name, 0.9)
+    acceptance_r2score = acceptance_r2scores.get(model_name, 0.9)
+    threshold_accuracy = threshold_accuracies.get(model_name, 0.9)
+
+    # If the model is a classifier
+    # KNeighborsClassifier does not provide a predict_proba method for now
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3962
+    if (
```
RomanBredehoft, line 207: the following tests all inference methods (predict, predict_proba, decision_function) and checks that results are close to sklearn

tests/sklearn/test_sklearn_models.py
```diff
@@ -462,10 +496,18 @@ def check_offset(model_class, n_bits, x, y):
         model.fit(x, y)
 
 
-def check_subfunctions(fitted_model, model_class, x):
-    """Check subfunctions."""
+def check_inference_methods(model, model_class, x, check_float_array_equal):
```
RomanBredehoft, line 499: this test now not only tests that all inference methods (predict, predict_proba, decision_function) work but also makes sure results (including shapes) are coherent between `fhe="disable"` and `fhe="simulate"`, which was not done before

tests/sklearn/test_sklearn_models.py
```diff
@@ -474,75 +516,119 @@ def check_subfunctions(fitted_model, model_class, x):
...
+            assert y_scores_clear.shape == y_scores_simulated.shape, (
+                "Method 'decision_function' from Concrete ML outputs different shapes when executed"
+                "in the clear and with simulation."
+            )
+            check_float_array_equal(y_scores_clear, y_scores_simulated)
 
 
-def check_subfunctions_in_fhe(model, fhe_circuit, x):
-    """Check subfunctions in FHE: calls and correctness."""
+def check_separated_inference(model, fhe_circuit, x, check_float_array_equal):
```
RomanBredehoft, line 567: this test is this Pr's initial goal : it tests that all inference methods (predict, predict_proba, decision_function) match the expected results using the circuit's ouputs + other additional computations (post_processing, argmax)

tests/sklearn/test_sklearn_models.py
```diff
@@ -678,69 +766,19 @@ def check_grid_search(model_class, x, y, scoring):
...
         ]:
-            # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3962
             pytest.skip("Skipping predict_proba for KNN, doesn't work for now")
 
         _ = GridSearchCV(
             model_class(), param_grid, cv=2, scoring=scoring, error_score="raise", n_jobs=1
         ).fit(x, y)
 
 
-def check_sklearn_equivalence(model_class, n_bits, x, y, check_accuracy, check_r2_score):
```
RomanBredehoft, line 693: we now only test sklearn equivalence with `check_correctness_with_sklearn`

tests/sklearn/test_sklearn_models.py
```diff
@@ -1153,35 +1146,36 @@ def test_correctness_with_sklearn(
...
-    n_bits,
     load_data,
     check_r2_score,
     check_accuracy,
     is_weekly_option,
     verbose=True,
 ):
     """Testing hyper parameters."""
+
+    n_bits = N_BITS_THRESHOLD_FOR_SKLEARN_CORRECTNESS_TESTS
```
RomanBredehoft, line 1167: the whole point of this test is to load hyper-parameters and test correctness, so let's use this default parameter here as well

tests/sklearn/test_sklearn_models.py
```diff
@@ -1153,35 +1146,36 @@ def test_correctness_with_sklearn(
     )
 
 
-@pytest.mark.parametrize("model_class, parameters", MODELS_AND_DATASETS)
+# Neural network hyper-parameters are not tested
 @pytest.mark.parametrize(
-    "n_bits",
-    N_BITS_WEEKLY_ONLY_BUILDS + N_BITS_REGULAR_BUILDS,
```
RomanBredehoft, line 1159: not exactly sure why but `get_hyper_param_combinations` does not provide anything for QNNs, meaning they were not tested here at all

tests/sklearn/test_sklearn_models.py
```diff
@@ -1363,25 +1357,26 @@ def test_input_support(
...
-    n_bits,
     load_data,
     is_weekly_option,
+    check_float_array_equal,
+    default_configuration,
     verbose=True,
 ):
-    """Test subfunctions."""
+    """Test inference methods."""
+    n_bits = min(N_BITS_REGULAR_BUILDS)
```
RomanBredehoft, line 1370: this test only checks for result coherence between clear and simulation executions, so I believe we don't really need to use lots of different numbers of bits

tests/sklearn/test_sklearn_models.py
```diff
@@ -1436,98 +1439,96 @@ def test_predict_correctness(
     load_data,
     default_configuration,
     check_is_good_execution_for_cml_vs_circuit,
-    check_circuit_has_no_tlu,
     is_weekly_option,
-    test_subfunctions_in_fhe=True,
     verbose=True,
 ):
-    """Test correct execution, if there is sufficiently n_bits."""
+    """Test prediction correctness between clear quantized and FHE simulation or execution."""
```
RomanBredehoft, line 1445: this test was overly complicated for basically only testing `check_is_good_execution_for_cml_vs_circuit` on a few FHe samples, so I made it simpler + extracted a small part of it with the new `test_linear_models_have_no_tlu` test (see below)

the only thing I removed here is the fact that we were testing `predict(x)` against `predict(x, fhe="disable")`, which I don't think is very relevant

tests/sklearn/test_sklearn_models.py
```diff
@@ -1716,3 +1717,32 @@ def test_load_fitted_sklearn_linear_models(
         print("Run check_load_pre_trained_sklearn_models")
 
     check_load_fitted_sklearn_linear_models(model_class, n_bits, x, y)
+
+
+# Only circuits from linear models do not have any TLUs
+@pytest.mark.parametrize("model_class, parameters", get_sklearn_linear_models_and_datasets())
+def test_linear_models_have_no_tlu(
```
RomanBredehoft, line 1724: this was part of the `test_predict_correctness` test above

src/concrete/ml/sklearn/base.py
kcelia: The sort algorithm is designed to sort an array of 1d.
So, you will have to loop at some point.

src/concrete/ml/sklearn/base.py
```diff
@@ -2093,11 +2093,11 @@ def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.
...
-            topk_labels.append(super().predict(query[None], fhe))
+            topk_labels.append(BaseEstimator.predict(self, query[None], fhe=fhe))
```
kcelia, line 2096: thanks

src/concrete/ml/sklearn/neighbors.py
kcelia: thanks

tests/deployment/test_client_server.py
```diff
@@ -97,172 +99,192 @@ def test_client_server_sklearn(
         warnings.simplefilter("ignore", category=ConvergenceWarning)
         model.fit(x_train, y_train)
 
-    # Compile
-    extra_params = {"global_p_error": 1 / 100_000}
```
kcelia, line 101: why did you remove it ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -20,10 +20,6 @@
   - calls to predict_proba
   - calls to decision_function
 
-Are currently missing
-  - check of predict_proba
-  - check of decision_function
-
 More information in https://github.com/zama-ai/concrete-ml-internal/issues/2682
```
kcelia, line 23: is this comment linked to what you deleted above ?

RomanBredehoft: seems like an old issue indeed, I'll remove it in this PR, thanks !

src/concrete/ml/sklearn/base.py
RomanBredehoft: ok got it, yeah it's too bad that we therefore need to have somewhat 2 unnecessary loops here instead of 1 but oh well, thanks !

src/concrete/ml/sklearn/base.py
```diff
@@ -260,15 +260,6 @@ def fhe_circuit(self) -> Optional[Circuit]:
         assert isinstance(self.fhe_circuit_, Circuit) or self.fhe_circuit_ is None
         return self.fhe_circuit_
 
-    @fhe_circuit.setter
-    def fhe_circuit(self, value: Circuit) -> None:
```
RomanBredehoft, line 264: no longer used in our tests (consequence of deployment test cleaning)

src/concrete/ml/sklearn/base.py
```diff
@@ -883,10 +876,6 @@ def output_quantizers(self, value: List[UniformQuantizer]) -> None:
     def fhe_circuit(self) -> Circuit:
         return self.quantized_module_.fhe_circuit
 
-    @fhe_circuit.setter
```
RomanBredehoft, line 886: same as above

tests/sklearn/test_sklearn_models.py
```diff
@@ -81,20 +75,11 @@
 # sufficiently number of bits for precision
 N_BITS_THRESHOLD_FOR_SKLEARN_CORRECTNESS_TESTS = 26
 
-# We check correctness with check_is_good_execution_for_cml_vs_circuit or predict in
-# fhe="disable" only if n_bits >= N_BITS_THRESHOLD_FOR_PREDICT_CORRECTNESS_TESTS. This is
-# because we need sufficiently number of bits for precision
-N_BITS_THRESHOLD_FOR_PREDICT_CORRECTNESS_TESTS = 6
```
RomanBredehoft, line 87: removes this as I believe it was not very relevant : it was only used for test that were checking CML clear vs CML FHE/simulation, but N_BITS should not impact such behavior. Also, that won't change anything in regular CIs as they use at least 6 bits anyway. On the other hand, it enables KNNClassifier to be tested on things that were skipped previously because of this (since KNNClassifier can only be tested for n_bits <= 5 when compiled) cc @kcelia

tests/sklearn/test_sklearn_models.py
```diff
@@ -1539,7 +1551,12 @@ def test_fitted_compiled_error_raises(
     verbose=True,
 ):
     """Test Fit and Compile error raises."""
-    n_bits = min(N_BITS_REGULAR_BUILDS)
+    if get_model_name(model_class) == "KNeighborsClassifier":
```
RomanBredehoft, line 1554: I added a few more of these as I believe they could be come future flaky tests (since `min(N_BITS_REGULAR_BUILDS)` = 6 but we have issues with KNN > 5 bits) cc @kcelia 

also, I made this like this as I'm not sure if the issue is temporary, but if it's not then mabe we just want to replace `min(N_BITS_REGULAR_BUILDS)` by `min(N_BITS_REGULAR_BUILDS + N_BITS_WEEKLY_BUILDS)`

tests/deployment/test_client_server.py
jfrery: This is legacy code when we had global_p_error = 0.01 by default in CML. Great that you remove it. Not sure about the n_bits though.

RomanBredehoft: I see thanks ! regarding the n_bits, I just removed the variable and directly put `n_bits=2` in the `compile` method as it wasn't used anywhere else

andrei-stoian-zama (approved):

PR #307, opened 2023-09-28
chore: add FL to showcase & encrypt/decrypt example

andrei-stoian-zama:
Closes https://github.com/zama-ai/concrete-ml-internal/issues/4026
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3974


kcelia (approved):

docs/README.md
```diff
@@ -54,9 +54,10 @@ Executing these steps separately is equivalent to calling `predict_proba` on the
 <!--pytest-codeblocks:cont-->
 
 ```python
+# Predict probability for a single example
 y_proba_fhe = model.predict_proba(X_test[[0]], fhe="execute")
 
-# Quantize an input (float)
+# Quantize a original float input
```
kcelia, line 60: an*

bcm-at-zama: this appears to be your speciality, @kcelia

RomanBredehoft (approved):
thanks !

README.md
```diff
@@ -107,6 +107,34 @@ print(f"Similarity: {int((y_pred_fhe == y_pred_clear).mean()*100)}%")
...
+
+# Decrypt the result (integer)
+q_y = model.fhe_circuit.decrypt(q_y_enc)
+
+# De-quantize and post-process the result
+y0 = model.post_processing(model.dequantize_output(q_y))
+
+print("Probability with `predict_proba`: ", y_proba_fhe)
+print("Probability with encrypt/run/decrypt calls: ", y0)
+```
```
bcm-at-zama, line 136: thx; I would add an assert to be sure it works well

andrei-stoian-zama: I wanted to but here we return floats so it's a bit of a pain to assert on allclose.

andrei-stoian-zama: we have pytests for that

bcm-at-zama: pytest will break: you'll fix pytest
but no one will think about fixing the doc
that's why having minimal checks in the doc is better, I think

PR #313, opened 2023-09-29
chore: make refresh notebook still commit successful notebooks on fail

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/3764

### DIFF
--- a/script/make_utils/jupyter.sh
+++ b/script/make_utils/jupyter.sh
@@ -15,7 +15,7 @@ WHAT_TO_DO="open"
 
 # Create a list of notebooks with long execution times in order not to consider them when refreshing
 # all notebooks at the same time.
-LONG_EXECUTION_TIMES_NOTEBOOKS=()
+LONG_EXECUTION_TIMES_NOTEBOOKS=("ClassifierComparison.ipynb")
 
 while [ -n "$1" ]
 do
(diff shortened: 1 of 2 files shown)


jfrery:
I am refreshing the notebooks here with the new workflow and will cherry-pick the commit to another PR.

Also need to revert the on: push.

script/make_utils/jupyter.sh
```diff
@@ -15,7 +15,7 @@ WHAT_TO_DO="open"
 
 # Create a list of notebooks with long execution times in order not to consider them when refreshing
 # all notebooks at the same time.
-LONG_EXECUTION_TIMES_NOTEBOOKS=()
+LONG_EXECUTION_TIMES_NOTEBOOKS=("ClassifierComparison.ipynb")
```
RomanBredehoft, line 18: do we need to do something here ? should we add a fixme then ?

jfrery: Let me remove this now that knn is gone from the comparison notebook.

RomanBredehoft (changes requested):
I'm not sure these changes makes successful notebooks be committed even if another fails ? or at least I can't see what changes would make this possible here !

jfrery:
> I'm not sure these changes makes successful notebooks be committed even if another fails ? or at least I can't see what changes would make this possible here !

Yes they have been committed in #304 while Deployment notebook failed.

script/make_utils/jupyter.sh
```diff
@@ -75,7 +80,11 @@ then
         echo "Refreshing ${NOTEBOOK}"
 
         START=$(date +%s)
-        jupyter nbconvert --to notebook --inplace --execute "${NOTEBOOK}"
+        if jupyter nbconvert --to notebook --inplace --execute "${NOTEBOOK}"; then
+            echo "${NOTEBOOK}" >> "${SUCCESSFUL_NOTEBOOKS}"
+        else
+            echo "${NOTEBOOK}" >> "${FAILED_NOTEBOOKS}"
+        fi
         END=$(date +%s)
```
jfrery, line 88: This is what makes the refresh notebook not crash on failure @RomanBredehoft .

RomanBredehoft (approved):
I find it weird that this makes it possible but I trust you ! thanks a lot for this !!

PR #331, opened 2023-10-03
chore: fix concurrency group in release action

RomanBredehoft:
So basically, I recently changed concurrency groups in workflows to make sure they have unique references using `group: "${{ github.ref }}-${{ github.event_name }}-${{ github.workflow }}"` . Turns out, it's not well supported when using reusable workflow.  

Basically, `github.workflow` are made the same between the called and caller workflows (source: https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow). This made the test CI workflow cancel the release workflow since they both share the same group ref but the CI one cancels all previous runs in progress when not triggered by a push on main (https://github.com/zama-ai/concrete-ml/blob/1516ee25024f618b01ee047efac332544cb25358/.github/workflows/continuous-integration.yaml#L75C3-L75C3)

I therefore made the release workflow group different (by just calling it "release") and I believe this should solve the problem. Also, I made sure that running a new release process actually cancels all other ones in progress (this makes sure we only have one release process at a time)

Another solution could have been to make the CI not cancel any workflows when triggered by a `workflow_dispatch` with `manual_call` and `release` as inputs. But I believe this is a bit more verbose and less comprehensible

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,9 +18,16 @@ on:
       - src/concrete/ml/version.py
       - 'docs/developer-guide/api/**'
 
+# We need to make sure that the concurrency group is not identical to the continuous-integration.yaml
+# one, else the release workflow will be canceled when calling that latter for tests. More 
+# specifically, we should avoid using the usual "${{ github.ref }}-${{ github.event_name }}-${{ github.workflow }}"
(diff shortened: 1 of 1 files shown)


fd0r (approved):
Interesting bug, good find!

PR #333, opened 2023-10-03
chore: knn tuto

kcelia:
In this PR, I added a new tutorial for KNN.

As you know, this model is really restrictive, as we are not able to exceed 9bits in the circuit, without having to wait > 1h for geyken + facing unstable execution with the compiler.

To apply the rounded features, we need to :
- provide the  "rounding_threshold_bits (int)" parameter 
- once rounding_threshold_bits parameter is provided, it will allow the instantiation of an auto-rounder object that will be used after the distance computation in '_inference_'
- force the compiler to use auto_adjust_rounders to True.

I want your advice to know where rounding_threshold_bits should be set for the user:
- init: 
        - naturel place for arguments
        - easy to add
        - but knn  won't follow other CML patterns
- in the compiler:
        - which seems more appropriate. 
        - The way I see it, the auto-rounder object  will be instantiated in this function. But it won't work if we use 'predict=disable', unless if I put a condition to skip the auto-rounder. But the big problem is that we won't have same results with and without auto-rounder.
- fit, the paramater

(for Q4, we will try to add this feature it to our built-in models, so it's like an advanced conversion)

CC @zama-ai/machine-learning 

------------------

We agreed to hide the parameter `rounding_threshold_bits` to the user and add a default value in base.py.
I suggest the value 5, you can see the plots in this branch 'testing/impact_of_rouning_on_knn'  in use_case_example folder.

Update:
~<!> The tests when we do not compile before the prediction, typically for fhe=disable are crushing, because we use the autorounder and we havent compiled the model !~

New update:
I used the adjust method instead of the compilation to use the autorounder feature.
The concrete python autorounder must become a serialized object !


docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,296 @@
...
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# The novel aspect introduced by Concret-ML models is the hyperparameter `n_bits`, which represents\n",
+    "# the precision for quantizing input data and intermediate values of the model.\n",
+    "# This quantization step is essential after the training phase, since FHE exclusively operates\n",
+    "# over integers.\n",
+    "concrete_knn = ConcreteKNeighborsClassifier(n_bits=3, n_neighbors=3)\n",
```
jfrery, line 84:
```
and intermediate values of the model
```
I don't think this is true? The accumulator depends on the n_bits but the precision of intermediate values will differ based on the operations.

I think we can have a shorter here that just says input is quantized to n_bits.

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,296 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Generating a key for an 9-bit circuit\n"
```
jfrery, line 132: Should we use round PBS? I think it's a great feature here as it's pretty hard to control the accumulator.

jfrery: Let's just make sure there is no crash and keep accumulator at 6-7 bits.

kcelia: I'll add back the round PBS here.

src/concrete/ml/sklearn/base.py
```diff
@@ -2080,6 +2079,9 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
         Returns:
             numpy.ndarray: The majority vote.
         """
+
+        self.topk = y_preds.squeeze()
```
RomanBredehoft, line 2083: this is weird no ? I don't see the KNN having this attribute anywhere else than here !

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+   "metadata": {},
+   "source": [
+    "# K-Nearest Neighbors\n",
+    "\n",
+    "\n",
+    "K-Nearest Neighbors (KNN) is a non-parametric supervised learning algorithm, suitable for both classification and regression tasks.\n",
+    "\n",
+    "In classification, KNN aims to identify the nearest points by measuring their similarity, often through distance metrics. The new labels are then assigned through majority voting, considering the most frequent labels among the neighboring points.\n",
+    "\n",
+    "In Fully Homomorphic Encryption (FHE), classification with KNN poses significant computational challenges due to the distance calculations and the sorting algorithms, which is currently a non-stable algorithm.\n",
```
RomanBredehoft, line 15: can you clarify what we mean by `which is currently a non-stable algorithm` ?

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "# Model instantiation"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# The novel aspect introduced by Concret-ML models is the hyperparameter `n_bits`, which represents\n",
```
RomanBredehoft, line 79: Concret-ML -> Concrete ML (without the -)

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "# the precision for quantizing input data\n",
+    "# This quantization step is essential after the training phase, since FHE exclusively operates\n",
+    "# over integers\n",
+    "\n",
+    "n_neighbors = 3\n",
+    "\n",
+    "concrete_knn = ConcreteKNeighborsClassifier(n_bits=3, n_neighbors=n_neighbors)\n",
+    "\n",
+    "# Fit both the Concrete ML and its equivalent float estimators on clear data\n",
```
RomanBredehoft, line 88: estimator (without the s)

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Saved in 47\n",
```
RomanBredehoft, line 108: what is this "Saved in 47" ?

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Generating a key for an 8-bit circuit\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(f\"Generating a key for an {circuit.graph.maximum_integer_bit_width()}-bit circuit\")"
```
RomanBredehoft, line 139: "-bit" -> "-bits"

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+     "output_type": "stream",
+     "text": [
+      "Key generation time: 829.81 seconds\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Note that this step may be time-consuming for circuits exceeding 8-bits\n",
+    "time_begin = time.time()\n",
+    "circuit.client.keygen(force=False)\n",
```
RomanBredehoft, line 165: not sure we want to expose the `force=False` feature imo

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Key generation time: 829.81 seconds\n"
```
RomanBredehoft, line 158: 829 seconds ? wow

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "# Note that this step may be time-consuming for circuits exceeding 8-bits\n",
```
RomanBredehoft, line 163: even with 8 bits we got 829s so maybe say more something like "This step takes several minutes" if this timing is expected

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "print(f\"Key generation time: {time.time() - time_begin:.2f} seconds\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Inference with Concrete ML:\n",
+    "\n",
+    "a. __clear__: inference on unencrypted quantized data, without any FHE execution \n",
```
RomanBredehoft, line 175: "unencrypted" feels like it has already been encrypted at one point, maybe "non-encrypted" ?

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "\n",
+    "b. __Execution in FHE__: inference on encrypted data, using actual FHE execution\n",
+    "\n",
+    "c. __Simulation__: inference on unencrypted quantized data, without secure FHE execution, while simulating the p_error failure probability. For further information, please consult: [TODO]()"
```
RomanBredehoft, line 179: same for "unencrypted", but simulation is not only p_error simulation, I would literally say that it simulates all FHE operations with similar crypo-parameters or something like that

also, what is this TODO ?

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Inference with Concrete ML:\n",
+    "\n",
+    "a. __clear__: inference on unencrypted quantized data, without any FHE execution \n",
+    "\n",
+    "b. __Execution in FHE__: inference on encrypted data, using actual FHE execution\n",
```
RomanBredehoft, line 177: I would talk about simulation first, and then FHE exec, as we usually do, making the first two inferences the ones in clear !

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+      "Time inference: 40.05 seconds per sample\n"
+     ]
+    }
+   ],
+   "source": [
+    "# a- Clear inference\n",
+    "pred_cml_clear = concrete_knn.predict(X_test, fhe=\"disable\")\n",
+    "score_cml_clear = (pred_cml_clear == y_test).mean()\n",
+    "\n",
+    "# b- FHE inference\n",
```
RomanBredehoft, line 211: same as above, I would put clear -> simulation -> FHE exec

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "time_begin = time.time()\n",
+    "pred_cml_fhe = concrete_knn.predict(X_test[0, None], fhe=\"execute\")\n",
+    "print(f\"Time inference: {time.time() - time_begin:.2f} seconds per sample\")\n",
```
RomanBredehoft, line 214: "time inference" -> "FHE inference execution time"

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    }
+   ],
+   "source": [
+    "# a- Clear inference\n",
+    "pred_cml_clear = concrete_knn.predict(X_test, fhe=\"disable\")\n",
+    "score_cml_clear = (pred_cml_clear == y_test).mean()\n",
+    "\n",
+    "# b- FHE inference\n",
+    "time_begin = time.time()\n",
+    "pred_cml_fhe = concrete_knn.predict(X_test[0, None], fhe=\"execute\")\n",
```
RomanBredehoft, line 213: how many samples is that ? maybe say why we use less samples here. Also, if that's a single sample, is it relevant to compute the score ? I feel like we should onl talk about the clear and simulated score, and then show the FHE exec time spearately, and only say it matches simulated accuracy

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# scikit-learn inference\n",
+    "predict_sklearn = sklearn_model.predict(X_test)\n",
+    "score_sklearn = (predict_sklearn == y_test).mean()"
```
RomanBredehoft, line 190: also maybe it's better to use sklearn's accuracy score for computing all scores, looks cleaner imo

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# The novel aspect introduced by Concret-ML models is the hyperparameter `n_bits`, which represents\n",
+    "# the precision for quantizing input data\n",
```
RomanBredehoft, line 80: I would put these comments in the markdown cell instead of comments like this, simpler to read I believe !

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Saved in 47\n",
+      "Compilation time: 2.56 seconds\n"
+     ]
+    }
+   ],
+   "source": [
+    "# The compilation step aims to:\n",
```
RomanBredehoft, line 114: same, why not put these comments in the markdown cell ?

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+   ],
+   "source": [
+    "print(f\"Generating a key for an {circuit.graph.maximum_integer_bit_width()}-bit circuit\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Key generation"
```
RomanBredehoft, line 146: maybe add a small intro sentence here

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "time_begin = time.time()\n",
+    "circuit.client.keygen(force=False)\n",
+    "print(f\"Key generation time: {time.time() - time_begin:.2f} seconds\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Inference with Concrete ML:\n",
```
RomanBredehoft, line 173: same, maybe add a small intro sentence here

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+     },
+     "execution_count": 29,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "df = pd.DataFrame(\n",
+    "    {\n",
+    "        \"Distance\": distance[:, 0],\n",
```
RomanBredehoft, line 384: does it also makes sense to have the CML distance ?

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "df.style.apply(highlight_diff, axis=1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Conclusion\n",
+    "\n",
+    "Due to its notable time and memory complexity, the K-nearest neighbors classifier is currently not well-suited for real-world applications, especially those involving large data-sets.\n",
```
RomanBredehoft, line 409: not sure we want to put it like this actually

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "\n",
+    "Performance-wise, the K-nearest neighbors classifier matches pretty good its scikit-learn equivalent. The loss in accuracy is caused by the quantization and to the fact that the sorting algorithm is non-stable."
```
RomanBredehoft, line 411: "Performance-wise, the K-nearest neighbors classifier matches pretty good its scikit-learn equivalent."
-> 
"The K-nearest neighbors classifier matches its scikit-learn equivalent's performance."

"The loss in accuracy is caused by the quantization and to the fact that the sorting algorithm is non-stable." -> 
"The small loss in accuracy comes from quantization artifacts as well as the fact that the sorting algorithm is non-stable." (still, I would explain what you mean by non-stable here)

also, maybe insist on the fact that FHE is exact, as we usually do

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Future work\n",
+    "\n",
+    "In future releases of **Concrete ML** we will improve the K-nearest neighbors classifier to consider more features and larger data-sets."
```
RomanBredehoft, line 420: not sure we want to say that in advanced examples (I mean, we'll always improve our models !)

RomanBredehoft (changes requested):
I have several comments !

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: It was for me (saving the mlir, in case it crashed). 
Forgot to delete it.

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: removed

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: For 8bits it's like 829 s ~ 10 mn.
For more 8bits it's more like  > 50 mn.

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: No.

In the _inference method, we return neither  the distance nor the argsort. But the topk labels.
Otherwise, we will end up with information leakage issues

kcelia: In the _inference method, we return the top-k labels, not the distance or argsort (to prevent information leakage issues).

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: otherwise i would recalculate them.

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: answered above

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: Changed unencrypted by non-encrypted.

TODO: a remainder to add a link.

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: good idea

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: Done

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "print(f\"Concrete ML (clear) score: {score_cml_clear:.2%}\")\n",
+    "print(f\"Concrete ML FHE (simulation) score: {score_cml_simulate:.2%}\")\n",
+    "print(f\"Concrete ML FHE score: {score_cml_fhe:.2%}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Concrete KNN vs. scikit-learn KNN"
```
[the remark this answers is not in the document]
kcelia: Done

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: There's a whole text after this title, what would you like to add?

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: We talked about it the weekly, it's really long

src/concrete/ml/sklearn/base.py
kcelia: I saved it to get the topk-labels returned by _inference and compare them with those of scikit learn.
The reason I squeeze it is that _inference returns a tuple, so it will return a tensor of (X_test, 1, n_neighorbs).
I squeezed it to get a tensor of (X_test, n_neighorbs).

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: We now need to discuss the design part, in which part rounding_threshold_bits would be most appropriate, either in :
- the init
- the compile 
- the fit

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: with rounding_threshold_bits features, it's faster. So, I ll perform the prediction for all the testing set.
I added a paragraph below to say that we get the same accuracy for simulate and real fhe execution

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: I agree, but unlike the other models, KNN is the most restrictive model that we have in CML.
It is rapidly complex, we cannot compile for n_bits >= 4 bits, the keygen is huge, we cannot be applied to databases with more than a dozen examples...

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: added a small sentence.

src/concrete/ml/sklearn/base.py
RomanBredehoft: what I meant is that attributes should not be exposed like that without being defined in the init ! But more than that, why is this needed ? is topk something exposed by the scikit-learn model ?  if so, only after running the predict method ? If not, then I would not add this attribute in CML at all

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,515 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "FHE inference execution time: 548.33 seconds per sample\n"
```
RomanBredehoft, line 451: it's not per sample here isn't it ?

docs/advanced_examples/KNearestNeighbors.ipynb
RomanBredehoft: then maybe you could add a small sentence when you talk about that in the intro, but it feels a bit weird to have it ad the end like this imo !

RomanBredehoft (changes requested):
still have some comments, but thanks for the changes you've already made !

src/concrete/ml/sklearn/base.py
kcelia: Yes, it's exposed by scikit-learn, through:  `distance, topk_args = sklearn_model.kneighbors(X_test)` after performing the prediction.
The reason that i added it, is to compare it with the one of SK as discussed in weekly.
I'll add it to the init.

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: It's already mentioned in the intro: 
```

In Fully Homomorphic Encryption (FHE), classification with KNN poses significant computational challenges due to the distance calculations and the sorting algorithms, which is currently a non-stable algorithm (i.e., does not consider the order of the elements).

It is therefore recommended to use it on small datasets (up to dozens of examples) with strong quantization (n_bits <= 4).

This tutorial focuses exclusively on the KNN classifier, comparing the performance of Concrete KNN with its scikit-learn counterpart on a small data-set.

```

I'll delete it, to avoid confusion.

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,515 @@
...
+   "cell_type": "code",
+   "execution_count": 22,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "sckit-learn score: 100.00%\n",
+      "Concrete ML (clear) score: 100.00%\n",
```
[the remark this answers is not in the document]
kcelia: Yes, perfect match for this seed, for ~ 13 examples.

src/concrete/ml/sklearn/base.py
RomanBredehoft: I'm not sure I get it , can we do `sklearn_model.topk` ? and does it work before calling `predict` ? if you are saying that it's available by calling the `kneighbors` method then it would be prefereable to have such a method. And if it requires to store predictions /topk in the class somewhere (which I still find intriguing) then it's probably done through a private attribute (which is better at least) ! 

Also, be careful of double fitting, serialization and deployment tests, I think it should be fine as it's made in post-processing but still !

docs/advanced_examples/KNearestNeighbors.ipynb
RomanBredehoft: the following is only bulet points, I was just saying that adding a small sentence before like "there are several ways to run the inference in CML" would be nicer

docs/advanced_examples/KNearestNeighbors.ipynb
RomanBredehoft: Then I would formulate a bit differently, 13min is already very time consuming ! Maybe "Circuits exceeding 8 bits can make the key generation time get up to close to an hour"

docs/advanced_examples/KNearestNeighbors.ipynb
RomanBredehoft: I'm not sure how the cell order is related to the markdown table. Also, top-k labels are done in the clear, so even if there is some kind of re-computation, we are talking of less than a second right ?

docs/advanced_examples/KNearestNeighbors.ipynb
RomanBredehoft: same

docs/advanced_examples/KNearestNeighbors.ipynb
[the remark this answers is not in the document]
RomanBredehoft: ah yes ok there not a lot of examples I get it

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -27,6 +28,7 @@ def __init__(
         self,
         n_bits=2,
         n_neighbors=3,
+        rounding_threshold_bits=6,
```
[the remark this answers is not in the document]
RomanBredehoft: @kcelia could you explain a bit more here ? I really don't think this should be done like this and I feel it needs further discussion about it

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: ah ok !

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: In this notebook, we want to compare the topk label of cml and that of sckit learn.

For SK, the topk label is available in a function called kneigbours, which returns the distance and indices of the k nearest neighbors.

for cml, this function doesn't exist ( which makes me think I should add a warning for this function).
for cml, only topk labels are retrieved to avoid information leaks.

as i was doing the FHE inference on 1 example, self.topk_label had 1 example. 
that's why i changed the order, to get the simulated pridictions on the whole test set.

Nervermind, I've restored the original order,, it takes ~ 4 mn, and I'm computing the predictions for the whole test set, now

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: fixed

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: I rephrased it, btw

src/concrete/ml/sklearn/neighbors.py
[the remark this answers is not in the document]
kcelia: I'll discuss it with Andrei, and update this PR.

fd0r (commented):
Do we have the pretty 2d comparison plot somewhere else?

docs/advanced_examples/KNearestNeighbors.ipynb
RomanBredehoft: > ( which makes me think I should add a warning for this function).

indeed you could add the method and raise a warning in `knn.py` , keep in mind that you'll need to update a test or add a new one to catch the error

> for cml, only topk labels are retrieved to avoid information leaks.

instead of a having a `_topk_labels` parameter (which you access here and is not great), would it be possible to have a similar/different `neighbors` method for CML models that only returns these labels ?

kcelia:
> We don't have the 2d comparison plot in the comparison notebook, shouldn't we include it in this notebook?

@fd0r, wichi second plot ?

src/concrete/ml/sklearn/neighbors.py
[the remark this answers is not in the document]
kcelia: We agreed to find the best default value for rounding_threshold_bits (to avoid sharing it in the init for the user).

src/concrete/ml/sklearn/base.py
```diff
@@ -1766,6 +1768,11 @@ def __init__(self, n_bits: int = 3):
         self._y: numpy.ndarray
         # _q_fit_X_quantizer: The quantizer to use for quantizing the model's training set
         self._q_fit_X_quantizer: Optional[UniformQuantizer] = None
+        # Topk labels
+        self._topk_labels: numpy.ndarray
+        # Number of bits to keep
+        self.rounding_threshold_bits = rounding_threshold_bits
+        self.rounder = cnp.AutoRounder(target_msbs=rounding_threshold_bits)
```
RomanBredehoft, line 1775: I think rounder should be private !

src/concrete/ml/sklearn/base.py
```diff
@@ -1766,6 +1768,11 @@ def __init__(self, n_bits: int = 3):
         self._y: numpy.ndarray
         # _q_fit_X_quantizer: The quantizer to use for quantizing the model's training set
         self._q_fit_X_quantizer: Optional[UniformQuantizer] = None
+        # Topk labels
+        self._topk_labels: numpy.ndarray
```
RomanBredehoft, line 1772: I you keep the following attributes , don't forget to : 
- add typings
- add a small comment of the form `# variable: description`

src/concrete/ml/sklearn/base.py
```diff
@@ -1955,7 +1962,7 @@ def gather1d(x, indices):
                 arr = []
                 for i in indices:
                     arr.append(x[i])
-                enc_arr = fhe_array(arr)
+                enc_arr = cnp.array(arr)
```
RomanBredehoft, line 1965: don't do it here I think but we'll remove this `cnp` mention later : https://github.com/zama-ai/concrete-ml-internal/issues/4074

src/concrete/ml/sklearn/base.py
```diff
@@ -1976,20 +1983,21 @@ def scatter1d(x, v, indices):
...
             n, k = x.size, self.n_neighbors
+            # Determine the number of stages for a sequence of length n
             ln2n = int(numpy.ceil(numpy.log2(n)))
 
-            # Number of stages
+            # Stage loop
             for t in range(ln2n - 1, -1, -1):
                 p = 2**t
                 r = 0
-                # d: Length of the bitonic sequence
```
RomanBredehoft, line 1988: could you add a comment for what t, p, r (and then q, bq) represent ?

src/concrete/ml/sklearn/base.py
```diff
@@ -2012,49 +2020,78 @@ def scatter1d(x, v, indices):
...
         # computations.
         # Being a monotonic function, it does not affect the logic of the calculation, notably for
         # the argsort.
 
         topk_labels = topk_sorting(distance_matrix.flatten(), self._y)
-
         return numpy.expand_dims(topk_labels, axis=0)
 
     def compile(self, *args, **kwargs) -> Circuit:
+        def force_auto_adjust_rounder_in_configuration(configuration):
```
RomanBredehoft, line 2075: could be made a utils function I think, also don't forget to put docstrings !

src/concrete/ml/sklearn/neighbors.py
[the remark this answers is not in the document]
RomanBredehoft: ok great to know !

fd0r:
> > We don't have the 2d comparison plot in the comparison notebook, shouldn't we include it in this notebook?
> 
> @fd0r, which second plot ?

Sorry not second, 2 dimension

src/concrete/ml/sklearn/base.py
kcelia: ok

src/concrete/ml/sklearn/base.py
kcelia: I added it to common/utils

src/concrete/ml/sklearn/base.py
RomanBredehoft: what I mean is that `cnp` is a reference to concrete-numpy which we don't use anymore

src/concrete/ml/sklearn/base.py
kcelia: I'll close this comment, as we agreed to set a new method called get_topk_labels.

src/concrete/ml/sklearn/base.py
kcelia: this variable has been removed

src/concrete/ml/sklearn/base.py
kcelia: I understood. 
I mean, I prefer to keep this naming, since we are planning to remove them in https://github.com/zama-ai/concrete-ml-internal/issues/4074.

Tracking cnp.X is easier than tracking fhe_array, fhe_zeros + cnp.X.  Don't you think?

RomanBredehoft: yes I agree that's what I said "don't do it here I think" 😅  I meant "we should update this cnp naming but later, it was just a comment saying that I've just realized this in your PR, which made me created the above PR !

src/concrete/ml/common/utils.py
```diff
@@ -582,3 +583,19 @@ def all_values_are_of_dtype(*values: Any, dtypes: Union[str, List[str]]) -> bool
         supported_dtypes[dtype] = supported_dtype
 
     return all(_is_of_dtype(value, supported_dtypes) for value in values)
+
+
+def force_auto_adjust_rounder_in_configuration(configuration, **kwargs):
+    """Force auto_adujet_rounder to True in configuration to use rounding features.
```
RomanBredehoft, line 589: `auto_adujet_rounder` -> auto_adjust_rounder

src/concrete/ml/common/utils.py
```diff
@@ -582,3 +583,19 @@ def all_values_are_of_dtype(*values: Any, dtypes: Union[str, List[str]]) -> bool
         supported_dtypes[dtype] = supported_dtype
 
     return all(_is_of_dtype(value, supported_dtypes) for value in values)
+
+
+def force_auto_adjust_rounder_in_configuration(configuration, **kwargs):
```
RomanBredehoft, line 588: also maybe type the configuration parameter + add the type in the docstring (surprise mape pcc didn't raise any issue here)

src/concrete/ml/sklearn/base.py
```diff
@@ -65,6 +64,8 @@
 from hummingbird.ml import convert as hb_convert  # noqa: E402
 from hummingbird.ml.operator_converters import constants  # noqa: E402
 
+# Default number of bits to keep
+DEFAULT_ROUNDING_THRESHOLD_BITS = 6
```
RomanBredehoft, line 68: do we only use this here ? either way I would still put it in common/utils.py I think ! Also, how did we choose this value ? maybe comment a bit on why we chose 6 !

src/concrete/ml/sklearn/base.py
```diff
@@ -2069,22 +2105,53 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
...
 
         topk_labels = []
         for query in X:
             topk_labels.append(BaseEstimator.predict(self, query[None], fhe=fhe))
 
-        y_preds = self.post_processing(numpy.array(topk_labels))
+        return numpy.array(topk_labels).squeeze()
+
+    def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
+        """Predict values for X, in FHE or in the clear.
```
RomanBredehoft, line 2135: you don't need to put docstrings for `predict` (since the class inherits from BaseEstimator !

src/concrete/ml/sklearn/base.py
```diff
@@ -2069,22 +2105,53 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
...
-            y_preds_processed.append(vote)
+        return numpy.array([self.majority_vote(y.flatten()) for y in y_preds])
 
-        return numpy.array(y_preds_processed)
+    def get_topk_labels(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
+        """Return the K-nearest labels of a point.
 
-    def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
+        Args:
+            X (Data): The input values to predict, as a Numpy array, Torch tensor, Pandas DataFrame
```
RomanBredehoft, line 2114: "The input values to predict" -> "The input values to consider" or something like that I think

src/concrete/ml/sklearn/base.py
```diff
@@ -2069,22 +2105,53 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
...
+                or List.
+            fhe (Union[FheMode, str]): The mode to use for prediction.
+                Can be FheMode.DISABLE for Concrete ML Python inference,
+                FheMode.SIMULATE for FHE simulation and FheMode.EXECUTE for actual FHE execution.
+                Can also be the string representation of any of these values.
+                Default to FheMode.DISABLE.
+
+        Returns:
+            numpy.ndarray: The topk labels for each point.
```
RomanBredehoft, line 2123: "topk" -> "K-Nearest" as you put in the first sentence

src/concrete/ml/sklearn/base.py
```diff
@@ -2069,22 +2105,53 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
...
+                Can also be the string representation of any of these values.
+                Default to FheMode.DISABLE.
+
+        Returns:
+            np.ndarray: The predicted values for X.
+        """
+
+        X = check_array_and_assert(X)
+
+        topk_labels = self.get_topk_labels(X, fhe)
```
RomanBredehoft, line 2152: great !

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -147,3 +147,22 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
...
+            X (Data): The input values to predict, as a Numpy array, Torch tensor, Pandas DataFrame
+                or List.
+
+        Raises:
+            NotImplementedError: The method is not implemented for now.
+        """
+
+        raise NotImplementedError(
+            "The `kneighbors` method is not implemented for KNeighborsClassifier. Please call "
+            "`get_topk_labels` to retieve the topk labels for each point, or `predict` method to "
```
RomanBredehoft, line 166: same, maybe "K-Nearest" instead of "topk"

src/concrete/ml/sklearn/base.py
```diff
@@ -2069,22 +2105,53 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
...
         """
-        y_preds_processed = []
-        for y in y_preds:
-            vote = self.majority_vote(y.flatten())
-            y_preds_processed.append(vote)
+        return numpy.array([self.majority_vote(y.flatten()) for y in y_preds])
 
-        return numpy.array(y_preds_processed)
+    def get_topk_labels(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
+        """Return the K-nearest labels of a point.
```
RomanBredehoft, line 2111: of a point or of each point ?

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -147,3 +147,22 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
             "The `predict_proba` method is not implemented for KNeighborsClassifier. Please "
             "call `predict` instead."
         )
+
+    # KNeighborsClassifier does not provide a kneighbors method
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4080
+    def kneighbors(self, X: Data) -> numpy.ndarray:
```
RomanBredehoft, line 153: yes nice

src/concrete/ml/sklearn/base.py
```diff
@@ -2012,49 +2018,79 @@ def scatter1d(x, v, indices):
...
+        Returns:
+            Circuit: The compiled Circuit.
+        """
+        # If a configuration instance is given as a positional parameter, set auto_adjust_rounder
+        if len(args) >= 2:
+            configuration = force_auto_adjust_rounder_in_configuration(args[1])
+            args_list = list(args)
+            args_list[1] = configuration
+            args = tuple(args_list)
+        # Else, retrieve the configuration in kwargs if it exists, or create a new one, and set the
```
RomanBredehoft, line 2088: newline just above the comment please !

RomanBredehoft (changes requested):
a few remaining comments, mostly on the form !

docs/advanced_examples/KNearestNeighbors.ipynb
RomanBredehoft: (because I still think that saying "could become time-consuming" while we already see 13 minutes is quite confusing)

src/concrete/ml/sklearn/base.py
kcelia: of each point

src/concrete/ml/sklearn/base.py
kcelia: it was a copy past, do you want me to change the original one ?

@RomanBredehoft

docs/advanced_examples/KNearestNeighbors.ipynb
kcelia: the optimizer has chosen a circuit of 7 bits, that's why it's  13 min.
If it's 8 or 9. It would be like 40 mn

src/concrete/ml/sklearn/base.py
kcelia: I added a branch with some experiments (testing/impact_of_rouning_on_knn) in the PR's description.

src/concrete/ml/sklearn/base.py
RomanBredehoft: yes but because the original one was for "predict", here it's not really the predict method so it feels less relevant to keep it that way

RomanBredehoft: not very important anyway

src/concrete/ml/sklearn/base.py
RomanBredehoft: yes thanks for that, I'm just saying that if we add something like this in the code, it should probably be better to put in utils  instead + add some comments around it to explain why it was chosen like this !

src/concrete/ml/common/utils.py
RomanBredehoft: (I think you forgot this change !)

src/concrete/ml/sklearn/base.py
```diff
@@ -2069,22 +2106,39 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
         Returns:
             numpy.ndarray: The majority vote.
         """
-        y_preds_processed = []
-        for y in y_preds:
-            vote = self.majority_vote(y.flatten())
-            y_preds_processed.append(vote)
+        return numpy.array([self.majority_vote(y.flatten()) for y in y_preds])
```
RomanBredehoft, line 2109: not sure why you have made this change, it's a bit harder to follow with one-liner (especially without any comments) !

src/concrete/ml/sklearn/base.py
```diff
@@ -1766,6 +1767,8 @@ def __init__(self, n_bits: int = 3):
         self._y: numpy.ndarray
         # _q_fit_X_quantizer: The quantizer to use for quantizing the model's training set
         self._q_fit_X_quantizer: Optional[UniformQuantizer] = None
+        # Use the rounding feature
+        self._rounder = cnp.AutoRounder(target_msbs=DEFAULT_ROUNDING_THRESHOLD_BITS)
```
RomanBredehoft, line 1771: question (and I see you more or less started the discussion in the PR's description) : should users still be able to modify the number of bits to use ? If not, fine with me then. But if so, maybe we could add a method like `update_auto_rounder` or something. 

Anyway, if we start to integrate the rounding feature in built-in models, we should indeed decide what we want to do, maybe you can start a discussion on slack about it and we'll talk at the next weekly !

kcelia: No, the user won't be able to update this value.
The idea in testing/impact_of_rouning_on_knn was to find an interesting default rounding value.
We will definitely talk about it in the weekly.

RomanBredehoft: got it thanks !

src/concrete/ml/sklearn/base.py
kcelia: The comment is in the docsting, for me it was obvious...
shall I get it back ?

src/concrete/ml/common/serialization/decoder.py
```diff
@@ -193,6 +193,7 @@ def object_hook(d: Any) -> Any:
                     QuantizationOptions,
                     UniformQuantizationParameters,
                     MinMaxQuantizationStats,
+                    AutoRounder,
```
kcelia, line 196: Add autoRounder to the serializable classes

src/concrete/ml/sklearn/base.py
```diff
@@ -2012,49 +2025,68 @@ def scatter1d(x, v, indices):
...
         # The square root in the Euclidean distance calculation is not applied to speed up FHE
         # computations.
         # Being a monotonic function, it does not affect the logic of the calculation, notably for
         # the argsort.
 
         topk_labels = topk_sorting(distance_matrix.flatten(), self._y)
-
         return numpy.expand_dims(topk_labels, axis=0)
 
     def compile(self, *args, **kwargs) -> Circuit:
```
RomanBredehoft, line 2079: you can remove this method now that it directly inherits from BaseEstimator without doing anything else !

src/concrete/ml/sklearn/base.py
```diff
@@ -1859,6 +1860,14 @@ def fit(self, X: Data, y: Target, **fit_parameters):
 
         self._is_fitted = True
 
+        # To enable the AutoRounder feature, it has to be adjusted using an inputset to determine
+        # how many of the least significant bits to remove.
+        # This can be done manually using fhe.AutoRounder.adjust(function, inputset), or
+        # by setting auto_adjust_rounders configuration to True during compilation.Rounding
+        # We chose to adjust it manually to get the same result between clear and FHE inference
+        inputset = numpy.array(list(numpy.expand_dims(q_input, 0) for q_input in self._q_fit_X))
```
RomanBredehoft, line 1868: may be better to use the `_get_inputset_generator` utils from quantized modules (I think it does exactly that but probably check before)

src/concrete/ml/sklearn/neighbors.py
```diff
@@ -110,7 +110,7 @@ def load_dict(cls, metadata: Dict):
         obj._q_fit_X_quantizer = metadata["_q_fit_X_quantizer"]
         obj._q_fit_X = metadata["_q_fit_X"]
         obj._y = metadata["_y"]
-
+        obj._rounder = metadata["_rounder"]
```
kcelia, line 113: This is generating flaky tests: 
```
            elif not lsbs_to_remove.is_adjusted:
                message = (
                    "AutoRounders cannot be used before adjustment, "
                    "please call AutoRounder.adjust with the function that will be compiled "
                    "and provide the exact inputset that will be used for compilation"
                )
>               raise RuntimeError(message)
E               RuntimeError: AutoRounders cannot be used before adjustment, please call AutoRounder.adjust with the function that will be compiled and provide the exact inputset that will be used for compilation
```
AutoRounders must become a serialized  class from concrete side.

andrei-stoian-zama (changes requested):
Looks pretty good! But I'd still like to see some graphs for rounding bits vs accuracy first. 

1. "b. Simulation: inference on non-encrypted quantized data, while simulating all FHE operations, failure probabilities and crypto-parameters. This mode of inference is recommended in the deployment phase. For further information, please consult [this link](https://docs.zama.ai/concrete-ml/advanced-topics/compilation#fhe-simulation)"

I think you mean "development" not "deployment"

2. you mention "accuracy loss" in the last line in the notebook but there is actually a 5% gain. You could say "accuracy difference"

src/concrete/ml/sklearn/neighbors.py
RomanBredehoft: have you taken a look at the class  ? could it be possible to just manually serialize all of its attributes, like we've done for some objects ?

also, we can't just re-create a new autorounder object when loading the model, am I wrong ? since they auto-rounder object is modified during fitting (right ?), re-creating the object would just be useless .. ? This is just to make sure we absolutely need to serialize it !

src/concrete/ml/sklearn/base.py
```diff
@@ -565,7 +564,7 @@ def compile(
         assert isinstance(self.fhe_circuit, Circuit)
 
         # CRT simulation is not supported yet
-        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3841
+        # TODO: https://github.com/zama-ai/concrete-ml-internal/issues/3841
```
RomanBredehoft, line 567: we prefer to keep everything as a FIXME ! Most of our scripts only look at fixme's and not todo's

kcelia: I didn't change it, I thought it appeared during the rebase.
I'm going to change it.

RomanBredehoft: yes I recently changed todos to fixme, you might have just accepted the wrong change when rebasing indeed !

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,439 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Key generation time: 627.87 seconds\n"
+     ]
+    }
+   ],
+   "source": [
+    "# For circuits exceeding 8-bits, the key generation process might become time-consuming\n",
```
RomanBredehoft, line 165: I still believe this sentence is confusing, when I see `Key generation time: 627.87 seconds` I'm already telling myself it's time-consuming. I would at least add "much more" or change it to something like "might take up to an hour", as you told me last time !

docs/advanced_examples/KNearestNeighbors.ipynb
```diff
@@ -0,0 +1,439 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Sckit-learn score: 62.50%\n",
```
RomanBredehoft, line 235: maybe "accuracy" instead of "score" is more precise here !

RomanBredehoft (approved):
looks good to me, thanks !

andrei-stoian-zama (approved):
Good job on this !

PR #336, opened 2023-10-04
chore: integrate Concrete Python 2.5.0rc1

RomanBredehoft:
This enables WoPBS simulation, which mean we can start using and testing the new simulation for our tests

Currently trying out the new simulation

PR #339, opened 2023-10-05
chore: integrate the new CP and the new VL

bcm-at-zama:
- integrate the new CP
- check it works fine with the new VL

refs #https://github.com/zama-ai/concrete-ml-internal/issues/3841

### DIFF
--- a/src/concrete/ml/common/utils.py
+++ b/src/concrete/ml/common/utils.py
@@ -37,7 +37,7 @@ MAX_BITWIDTH_BACKWARD_COMPATIBLE = 8
 # Use new VL with .simulate() by default once CP's multi-parameter/precision bug is fixed
 # TODO: https://github.com/zama-ai/concrete-ml-internal/issues/3856
 # Indicate if the old simulation method should be used when simulating FHE executions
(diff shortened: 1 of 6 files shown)


src/concrete/ml/common/utils.py
```diff
@@ -37,7 +37,7 @@
 # Use new VL with .simulate() by default once CP's multi-parameter/precision bug is fixed
 # TODO: https://github.com/zama-ai/concrete-ml-internal/issues/3856
 # Indicate if the old simulation method should be used when simulating FHE executions
-USE_OLD_VL = True
+USE_OLD_VL = False
```
bcm-at-zama, line 40: this is a test, we'll see if we change the default (ie, does it break some tests or does it change build execution times)

bcm-at-zama:
so looks like we have issues in pytest. Not obvious if they are due to CP or if some tests in CML need to be updated, due to the new VL: will make issues

bcm-at-zama:
Related to https://github.com/zama-ai/concrete-ml-internal/issues/4048

bcm-at-zama:
sorry dup of #336

PR #348, opened 2023-10-09
chore: update packages used in tests with pypi

RomanBredehoft:
We currently have some coverage issues with pypi tests in the weekly CI (https://github.com/zama-ai/concrete-ml/actions/runs/6450198846/job/17509408289). Not sure why it happens but let's start with updating the packages used 

Besides, we should probably take another look at https://github.com/zama-ai/concrete-ml-internal/issues/2005 and more specifically https://github.com/zama-ai/concrete-ml-internal/issues/2685 in order to avoid such manual updates


fd0r (approved):
Fine for now but would be great to have optional requirements so that we could do `pip concrete-ml[dev]`

RomanBredehoft:
> Fine for now but would be great to have optional requirements so that we could do pip concrete-ml[dev]

I agree, hence the links I referenced above !

PR #352, opened 2023-10-10
chore: fix drift seen in xgboost regressor

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4076
closes https://github.com/zama-ai/concrete-ml-internal/issues/2604
closes https://github.com/zama-ai/concrete-ml-internal/issues/4079

PR #355, opened 2023-10-11
chore: fix fhe=disable inference

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4076
closes https://github.com/zama-ai/concrete-ml-internal/issues/2604

### DIFF
--- a/src/concrete/ml/sklearn/base.py
+++ b/src/concrete/ml/sklearn/base.py
@@ -1382,7 +1382,17 @@ class BaseTreeEstimatorMixin(BaseEstimator, sklearn.base.BaseEstimator, ABC):
     def _inference(self, q_X: numpy.ndarray) -> numpy.ndarray:
         assert self._tree_inference is not None, self._is_not_fitted_error_message()
 
-        return self._tree_inference(q_X)[0]
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4077
+        # Incorrect batch inference for XGBoost. We predict one example at a time.
+        y_preds = []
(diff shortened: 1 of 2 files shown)


src/concrete/ml/sklearn/base.py
```diff
@@ -1382,7 +1382,17 @@ def compile(self, *args, **kwargs) -> Circuit:
     def _inference(self, q_X: numpy.ndarray) -> numpy.ndarray:
         assert self._tree_inference is not None, self._is_not_fitted_error_message()
 
-        return self._tree_inference(q_X)[0]
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4077
```
kcelia, line 1385: I need some clarifications please:
You said: Ensembles of trees are cut right before the reduce sum node. But here we can see a + 0.5 right after which hasn't been added to the post_processing.
But you haven't changed the post_processing, why ?

kcelia: you said: a batch of example in the disable still gives wrong predictions vs. one example at a time.

Here you are looping through each element at a time, isn't enough to fix the issue ?

jfrery: it's the last fix for 4096 so it's closing it but the add node thing is tackled in #352 indeed.

kcelia: Thanks !

jfrery:
Ok so it seems the batch problem I observed came from torch 2.0. Closing this PR.

PR #361, opened 2023-10-13
chore: revert cleaning of FAILED_TESTS_ARE_FLAKY in CI

RomanBredehoft:
without this (which is reverting a small commit from last week), the CI remains green even if some tests or coverage fails (as we can see in https://github.com/zama-ai/concrete-ml/actions/runs/6496948965/job/17645139299?pr=333) 🙂  So better keep it that way


.github/workflows/continuous-integration.yaml
```diff
@@ -728,8 +731,6 @@ jobs:
             fi
           fi
 
-          echo "FAILED_TESTS_ARE_FLAKY=${FAILED_TESTS_ARE_FLAKY}" >> "$GITHUB_ENV"
```
RomanBredehoft, line 731: basically the problem is that this step needs `shell: bash +e {0}`, which made this echo cancel the above `exit 1` command in case of test or coverage failure !

jfrery (approved):

PR #363, opened 2023-10-13
chore: fix triggering of release CI after prepapre_release

RomanBredehoft:
Small bug that made the CI release fail when triggered automatically after merging the "prepare release" PR.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -76,11 +76,17 @@ concurrency:
 
 env:
   ACTION_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
-  IS_PR: ${{ github.event_name == 'pull_request' }}
+  # The CI can be triggered by the release workflow which itself can be triggered by the merge of a 
+  # pull-request (following the 'prepare_release' workflow). Since GitHub weirdly propagates the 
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -76,11 +76,17 @@ concurrency:
...
+  # differentiate the release CI from regular CIs by using 'inputs.event_name', which should be set
+  # to "release" by the release workflow
+  IS_PR: ${{ github.event_name == 'pull_request' && inputs.event_name != 'release' }}
   IS_WEEKLY: ${{ github.event_name == 'schedule' || ((github.event_name == 'workflow_dispatch') && (inputs.event_name == 'weekly')) }}
   # The 'IS_RELEASE' variable indicates that the workflow has been triggered by the releasing 
-  # process itself, before publishing it
-  IS_RELEASE: ${{ github.event_name == 'workflow_dispatch' && !(inputs.manual_call) && inputs.event_name == 'release' }}
+  # process itself, before publishing it. It should only happen when the release workflow triggers
+  # the CI, in which 'inputs.event_name' is set to "release"
+  IS_RELEASE: ${{ inputs.event_name == 'release' }}
```
RomanBredehoft, line 89: this condition should be enough (and simpler to understand) as we only set `release` as an event_name input in the release workflow

PR #367, opened 2023-10-16
chore: integrate Concrete Python 2.5.0rc1

RomanBredehoft:
Integrate Concrete Python 2.5.0rc1 : the new simulation still does not properly work and is not enabled by default yet 

refs https://github.com/zama-ai/concrete-ml-internal/issues/4091


fd0r (dismissed):

fd0r (approved):

andrei-stoian-zama (changes requested):
As discussed, let's keep the old VL for now

fd0r (approved):

PR #371, opened 2023-10-17
chore: update privacy tree paper experiment

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4089


RomanBredehoft (approved):
thanks !

RomanBredehoft (changes requested):
actually, would it be possible to better format the printed tables ? I guess there are meant to be put in a latex format, but it's not very readable here

jfrery:
> actually, would it be possible to better format the printed tables ? I guess there are meant to be put in a latex format, but it's not very readable here

The main point is to copy paste directly to latex. But maybe we can instead print the dataframe yes

fd0r (dismissed):

RomanBredehoft:
> The main point is to copy paste directly to latex. But maybe we can instead print the dataframe yes

I don't really agree, imo the main point of putting this notebook in our directory is to give our users the possibility to be able to see where these timings and plots come from. So it should be kept easily readable like the previous one ! We can keep the latex format  in some internal repo if we want

docs/advanced_examples/ExperimentPrivacyTreePaper.ipynb
```diff
@@ -105,89 +206,114 @@
...
-      "Average precision (fp32): 0.4878108468555631\n",
-      "Average precision: 1.0\n",
-      "Average precision (fp32): 1.0\n",
-      "Average precision: 1.0\n",
-      "Average precision (fp32): 1.0\n",
-      "Average precision: 1.0\n",
-      "Average precision (fp32): 1.0\n"
+      "DecisionTreeClassifier on spambase (#features: 57) -> Acc: 0.9103, Acc (fp32): 0.9035, FHE inference time: 1.59s\n",
+      "XGBClassifier on spambase (#features: 57) -> Acc: 0.9313, Acc (fp32): 0.9362, FHE inference time: 2.01s\n",
+      "RandomForestClassifier on spambase (#features: 57) -> Acc: 0.9089, Acc (fp32): 0.9184, FHE inference time: 6.88s\n",
```
kcelia, line 211: I agree, this print is not visually appealing.

I can suggest to reformat the print, like:


```
        shift = 52 - len(f"{cls.__name__} on {key_dataset}")

        print(
            f"{cls.__name__} on {key_dataset} {'->':>{shift}} Acc: {np.mean(scores['accuracy']):.4f}, "
            f"Acc (fp32): {np.mean(scores_fp32['accuracy']):.4f}, "
            f"FHE inference time: {scores['inference_time']:.2f}s"
        )

```
image

andrei-stoian-zama:
I think it's not that important that the formatting is nice, it's important that we show our results are reproducible. So @jfrery merge this in the form that you want.

kcelia (approved):

RomanBredehoft (approved):

PR #372, opened 2023-10-17
chore: fix shape

kcelia:
The problem occurred after I merged the changes of this PR (https://github.com/zama-ai/concrete-ml/pull/333) yesterday.
Where I added the `get_topk_labels` method to the `predict` of KNN class.

- The issue comes from the squeeze operation:
If the input shape is (1, 1, k), squeezed turns it to (k,). Which I have corrected it.

- Why the test didn't catch this problem?
`numpy.isclose`, compares elements in arrays but doesn't consider their shape.
Exemple:
```
a = numpy.array([0])
b = numpy.array([0, 0, 0])
numpy.isclose(a, b, rtol=0, atol=0.001)
True
```
In the flaky: https://github.com/zama-ai/concrete-ml-internal/issues/4092:

```
a = numpy.array([0])
b = numpy.array([0, 0, 1])
numpy.isclose(a, b, rtol=0, atol=0.001)
False.
```

In this PR, I corrected the shape mismatch.

I changed `_isclose` by `numpy.testing.assert_allclose`, which takes into account the shape, in our tests.

closes https://github.com/zama-ai/concrete-ml-internal/issues/2806
closes https://github.com/zama-ai/concrete-ml-internal/issues/4092

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -319,10 +319,16 @@ def check_array_equal():
 def check_float_array_equal():
     """Fixture to check if two float arrays are equal with epsilon precision tolerance."""
 
-    def check_float_array_equal_impl(a, b, rtol=0, atol=0.001):
-        assert array_allclose_and_same_shape(
-            a, b, rtol, atol
-        ), f"Not equal to tolerance rtol={rtol}, atol={atol}\na: {a}\nb: {b}"
+    def check_float_array_equal_impl(
+        a, b, rtol=0, atol=0.001, error_information: Optional[str] = ""
+    ):
+
+        error_message = (
+            f"Not equal to tolerance rtol={rtol}, atol={atol}\na: {a}\nb: {b}\n"
+            f"{error_information}"
+        )
+
+        assert array_allclose_and_same_shape(a, b, rtol, atol), error_message
 
     return check_float_array_equal_impl
 
--- a/conftest.py
+++ b/conftest.py
@@ -16,6 +16,7 @@ from sklearn.datasets import make_classification, make_regression
 from concrete.ml.common.utils import (
     SUPPORTED_FLOAT_TYPES,
     all_values_are_floats,
+    array_allclose_and_same_shape,
     is_brevitas_model,
     is_classifier_or_partial_classifier,
     is_model_class_in_a_list,
@@ -318,8 +319,10 @@ def check_array_equal():
 def check_float_array_equal():
     """Fixture to check if two float arrays are equal with epsilon precision tolerance."""
 
-    def check_float_array_equal_impl(a, b):
-        assert numpy.all(numpy.isclose(a, b, rtol=0, atol=0.001))
+    def check_float_array_equal_impl(a, b, rtol=0, atol=0.001):
+        assert array_allclose_and_same_shape(
+            a, b, rtol, atol
+        ), f"Not equal to tolerance rtol={rtol}, atol={atol}\na: {a}\nb: {b}"
 
     return check_float_array_equal_impl
 
@@ -338,8 +341,7 @@ def check_r2_score():
         r2_num = numpy.sum(deltas_actual**2)
 
         # If the values are really close, we consider the test passes
-        is_close = numpy.allclose(expected, actual, atol=1e-4, rtol=0)
-        if is_close:
+        if array_allclose_and_same_shape(expected, actual, atol=1e-4, rtol=0):
             return
 
         # If the variance of the target values is very low, fix the max allowed for residuals
@@ -509,10 +511,7 @@ def check_is_good_execution_for_cml_vs_circuit():
                         "a QuantizedModule object."
                     )
 
-            # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/2806
-            # fp64 comparisons do not pass the numpy.array_equal while the quantized
-            # int64 values do.
-            if numpy.isclose(results_cnp_circuit, results_model).all():
+            if array_allclose_and_same_shape(results_cnp_circuit, results_model):
                 return
 
         raise RuntimeError(
--- a/src/concrete/ml/common/utils.py
+++ b/src/concrete/ml/common/utils.py
@@ -583,3 +583,27 @@ def all_values_are_of_dtype(*values: Any, dtypes: Union[str, List[str]]) -> bool
         supported_dtypes[dtype] = supported_dtype
 
     return all(_is_of_dtype(value, supported_dtypes) for value in values)
+
+
+def array_allclose_and_same_shape(
+    a, b, rtol: float = 1e-05, atol: float = 1e-08, equal_nan: bool = False
+) -> bool:
+    """Check if two numpy arrays are equal within a tolerances and have the same shape.
+
+    Args:
+        a (numpy.ndarray): The first input array
+        b (numpy.ndarray): The second input array
+        rtol (float): The relative tolerance parameter
+        atol (float): The absolute tolerance parameter
+        equal_nan (bool): Whether to compare NaN’s as equal. If True, NaN’s in a will be considered
+            equal to NaN’s in b in the output array
+
+    Returns:
+        bool: True if the arrays have the same shape and all elements are equal within the specified
+            tolerances, False otherwise.
+    """
+
+    a = check_array_and_assert(a, ensure_2d=False)
+    b = check_array_and_assert(b, ensure_2d=False)
+
+    return a.shape == b.shape and numpy.allclose(a, b, rtol, atol, equal_nan)
--- a/src/concrete/ml/common/utils.py
+++ b/src/concrete/ml/common/utils.py
@@ -603,7 +603,7 @@ def array_allclose_and_same_shape(
             tolerances, False otherwise.
     """
 
-    a = check_array_and_assert(a, ensure_2d=False)
-    b = check_array_and_assert(b, ensure_2d=False)
+    assert isinstance(a, numpy.ndarray)
+    assert isinstance(b, numpy.ndarray)
 
     return a.shape == b.shape and numpy.allclose(a, b, rtol, atol, equal_nan)
--- a/src/concrete/ml/quantization/quantized_module_passes.py
+++ b/src/concrete/ml/quantization/quantized_module_passes.py
@@ -242,7 +242,7 @@ class PowerOfTwoScalingRoundPBSAdapter:
             log2_value = int(numpy.rint(numpy.log2(value)))
             # Check that the integer power of two is close to the original value
             # with a small percentage tolerance
-            if numpy.isclose(numpy.power(2.0, log2_value), value, rtol=0.01):
+            if numpy.allclose(numpy.power(2.0, log2_value), value, rtol=0.01):
                 return log2_value, True
             return 0, False
 
--- a/src/concrete/ml/quantization/quantizers.py
+++ b/src/concrete/ml/quantization/quantizers.py
@@ -747,7 +747,7 @@ class UniformQuantizer(UniformQuantizationParameters, QuantizationOptions, MinMa
         assert self.scale is not None
 
         if QUANT_ROUND_LIKE_ROUND_PBS:
-            qvalues = numpy.floor(values / self.scale + self.zero_point + 0.5)
+            qvalues = numpy.floor(values / self.scale + self.zero_point + 0.5)  # pragma: no cover
         else:
             qvalues = numpy.rint(values / self.scale + self.zero_point)
 
--- a/src/concrete/ml/quantization/quantizers.py
+++ b/src/concrete/ml/quantization/quantizers.py
@@ -7,9 +7,9 @@ from typing import Any, Dict, Optional, TextIO, Union, get_type_hints
 
 import numpy
 
-from ..common import utils
 from ..common.debugging import assert_true
 from ..common.serialization.dumpers import dump, dumps
+from ..common.utils import QUANT_ROUND_LIKE_ROUND_PBS, array_allclose_and_same_shape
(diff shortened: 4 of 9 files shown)


fd0r (dismissed):

conftest.py
```diff
@@ -338,9 +338,11 @@ def check_r2_score_impl(expected, actual, acceptance_score=0.99):
         r2_num = numpy.sum(deltas_actual**2)
 
         # If the values are really close, we consider the test passes
-        is_close = numpy.allclose(expected, actual, atol=1e-4, rtol=0)
-        if is_close:
+        try:
```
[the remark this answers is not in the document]
jfrery: You can create a function called arrays_are_close probably and call it here.

RomanBredehoft (changes requested):
I think you should replace check_float_array_equal's impl by numpy.testing.assert_allclose and replace all of our asserts on isclose or allclose by this fixture !

conftest.py
```diff
@@ -16,6 +16,8 @@
 from concrete.ml.common.utils import (
     SUPPORTED_FLOAT_TYPES,
     all_values_are_floats,
+    check_allclose_and_shape,
```
RomanBredehoft, line 19: "check" semantic is more for fixture or stuff that raises errors, I would rename these two ones by something like `array_isclose_and_same_shape`  (same for allclose)

src/concrete/ml/common/utils.py
```diff
@@ -34,10 +34,9 @@
 
 MAX_BITWIDTH_BACKWARD_COMPATIBLE = 8
 
-# Indicate if the old virtual library method should be used instead of the compiler simulation
-# when simulating FHE executions
-# Set 'USE_OLD_VL' to False by default once the new simulation is fixed
-# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4091
+# Use new VL with .simulate() by default once CP's multi-parameter/precision bug is fixed
```
RomanBredehoft, line 37: is this change expected ? I would  revert it as the correct link to put here is https://github.com/zama-ai/concrete-ml-internal/issues/4091 and not https://github.com/zama-ai/concrete-ml-internal/issues/3856 !

src/concrete/ml/common/utils.py
```diff
@@ -583,3 +582,33 @@ def all_values_are_of_dtype(*values: Any, dtypes: Union[str, List[str]]) -> bool
         supported_dtypes[dtype] = supported_dtype
 
     return all(_is_of_dtype(value, supported_dtypes) for value in values)
+
+
+def check_isclose_and_shape(a, b, **kwargs):
```
RomanBredehoft, line 587: better to put the few parameters handled by isclose instead of kwargs imo, it's less error-prone !

src/concrete/ml/common/utils.py
```diff
@@ -583,3 +582,33 @@ def all_values_are_of_dtype(*values: Any, dtypes: Union[str, List[str]]) -> bool
...
+        **kwargs: Additional keyword arguments to be passed to `numpy.isclose`.
+
+    Returns:
+        bool: True if the arrays have the same shape and all elements are equal within the specified
+            tolerances, False otherwise.
+    """
+    return a.shape == b.shape and numpy.isclose(a, b, **kwargs).all()
+
+
+def check_allclose_and_shape(a, b, **kwargs):
```
RomanBredehoft, line 602: same, avoid kwargs imo !

src/concrete/ml/common/utils.py
RomanBredehoft, line 587: actually now that I've checked more carefully, I don't think this is needed : we only call `numpy.isclose` without a .all on a if statement a single time, and I think it's an error ! Basically, `isclose` returns an array, not a bool, so we should never have something like `if array:` !

src/concrete/ml/common/utils.py
```diff
@@ -583,3 +582,33 @@ def all_values_are_of_dtype(*values: Any, dtypes: Union[str, List[str]]) -> bool
...
+    Args:
+        a (numpy.ndarray): The first input array
+        b (numpy.ndarray): The second input array
+        **kwargs: Additional keyword arguments to be passed to `numpy.isclose`.
+
+    Returns:
+        bool: True if the arrays have the same shape and all elements are equal within the specified
+            tolerances, False otherwise.
+    """
+    return a.shape == b.shape and numpy.allclose(a, b, **kwargs)
```
RomanBredehoft, line 614: maybe check that a and b are numpy arrays as well first, else we might have an undesired error where accessing shapes !

src/concrete/ml/quantization/quantized_module_passes.py
```diff
@@ -242,7 +243,7 @@ def integer_log2(value: float) -> Tuple[int, bool]:
             log2_value = int(numpy.rint(numpy.log2(value)))
             # Check that the integer power of two is close to the original value
             # with a small percentage tolerance
-            if numpy.isclose(numpy.power(2.0, log2_value), value, rtol=0.01):
+            if check_isclose_and_shape(numpy.power(2.0, log2_value), value, rtol=0.01):
```
RomanBredehoft, line 246: as said above, I think that this is a mistake originaly, we should use your new _allclose_ util instead here !

src/concrete/ml/quantization/quantizers.py
```diff
@@ -369,7 +369,7 @@ def check_is_uniform_quantized(self, options: QuantizationOptions) -> bool:
 
         re_quant_scales = numpy.rint(unique_scales / min_scale) * min_scale
 
-        return numpy.allclose(unique_scales, re_quant_scales, atol=0.02)
+        return utils.check_allclose_and_shape(unique_scales, re_quant_scales, atol=0.02)
```
RomanBredehoft, line 372: I think you can directly change to `from utils import ...` and call the function directly, it's more readable !

src/concrete/ml/sklearn/base.py
```diff
@@ -2086,9 +2086,9 @@ def get_topk_labels(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -
 
         topk_labels = []
         for query in X:
-            topk_labels.append(BaseEstimator.predict(self, query[None], fhe=fhe))
+            topk_labels.append(BaseEstimator.predict(self, numpy.expand_dims(query, 0), fhe=fhe)[0])
```
RomanBredehoft, line 2089: you can maybe separate this into two lines as we usually do, it would make this more readable !

tests/onnx/test_onnx_ops_impl.py
```diff
@@ -52,9 +53,9 @@ def test_numpy_gemm(alpha, beta, trans_a, size_a, trans_b, size_b):
 
     # Can be a bit different if we have floats
     if isinstance(alpha, float) or isinstance(beta, float):
-        assert numpy.isclose(
+        assert check_isclose_and_shape(
```
RomanBredehoft, line 56: use the new `check_float_array_equal` fixture instead !

conftest.py
```diff
@@ -319,7 +321,7 @@ def check_float_array_equal():
     """Fixture to check if two float arrays are equal with epsilon precision tolerance."""
 
     def check_float_array_equal_impl(a, b):
-        assert numpy.all(numpy.isclose(a, b, rtol=0, atol=0.001))
+        assert check_isclose_and_shape(a, b, rtol=0, atol=0.001)
```
RomanBredehoft, line 324: would be great to add an error message here like we do for others (like check_array_equal) ! ex: `f"expected {expected}, got {got}, abs diff is {numpy.abs(got - expected)}"`

tests/quantization/test_quantizers.py
```diff
@@ -42,7 +43,7 @@ def test_quant_dequant_update(values, n_bits, is_signed, is_symmetric, check_arr
 
     # Check that all values are close
     tolerance = quant_array.quantizer.scale / 2
-    assert numpy.isclose(dequant_values, values, atol=tolerance).all()
+    assert check_isclose_and_shape(dequant_values, values, atol=tolerance)
```
RomanBredehoft, line 46: use the new check_float_array_equal fixture instead !

tests/sklearn/test_sklearn_models.py
```diff
@@ -1131,7 +1132,7 @@ def check_load_fitted_sklearn_linear_models(model_class, n_bits, x, y):
     y_pred_simulate = concrete_model.predict(x, fhe="simulate")
     y_pred_simulate_loaded = loaded_concrete_model.predict(x, fhe="simulate")
 
-    assert numpy.isclose(y_pred_simulate, y_pred_simulate_loaded).all(), (
+    assert check_isclose_and_shape(y_pred_simulate, y_pred_simulate_loaded), (
```
RomanBredehoft, line 1135: use the new check_float_array_equal fixture instead !

tests/torch/test_compile_torch.py
```diff
@@ -1166,7 +1170,7 @@ def test_shape_operations_net(
         # In PTQ the results do not match because of a-priori set quantization options
         # Currently no solution for concat/reshape/transpose correctness in PTQ is proposed.
         if is_qat:
-            assert numpy.allclose(torch_output, predictions, atol=0.05, rtol=0)
+            assert check_allclose_and_shape(torch_output, predictions, atol=0.05, rtol=0)
```
RomanBredehoft, line 1173: use the new check_float_array_equal fixture instead !

src/concrete/ml/common/utils.py
kcelia: Really don't know what happened here

src/concrete/ml/quantization/quantized_module_passes.py
kcelia: Good catch

src/concrete/ml/common/utils.py
kcelia: agreed to keep is_allclose

kcelia:
This change impacted other change ! 
Turn it to draft

src/concrete/ml/quantization/quantized_module_passes.py
```diff
@@ -242,7 +242,7 @@ def integer_log2(value: float) -> Tuple[int, bool]:
             log2_value = int(numpy.rint(numpy.log2(value)))
             # Check that the integer power of two is close to the original value
             # with a small percentage tolerance
-            if numpy.isclose(numpy.power(2.0, log2_value), value, rtol=0.01):
+            if numpy.allclose(numpy.power(2.0, log2_value), value, rtol=0.01):
```
kcelia, line 245: np.isclose returns an array. So, this  test is wrong

tests/onnx/test_onnx_ops_impl.py
```diff
@@ -52,9 +52,10 @@ def test_numpy_gemm(alpha, beta, trans_a, size_a, trans_b, size_b):
 
     # Can be a bit different if we have floats
     if isinstance(alpha, float) or isinstance(beta, float):
-        assert numpy.isclose(
+
+        assert numpy.allclose(
```
RomanBredehoft, line 56: you can use your new `check_float_array_equal` here as well !

tests/quantization/test_quantizers.py
```diff
@@ -42,7 +44,7 @@ def test_quant_dequant_update(values, n_bits, is_signed, is_symmetric, check_arr
 
     # Check that all values are close
     tolerance = quant_array.quantizer.scale / 2
-    assert numpy.isclose(dequant_values, values, atol=tolerance).all()
+    assert check_float_array_equal(dequant_values, values, atol=tolerance)
```
RomanBredehoft, line 47: `check_float_array_equal` should not be asserted, it already calls an assert !

tests/sklearn/test_sklearn_models.py
```diff
@@ -1131,10 +1130,13 @@ def check_load_fitted_sklearn_linear_models(model_class, n_bits, x, y):
     y_pred_simulate = concrete_model.predict(x, fhe="simulate")
     y_pred_simulate_loaded = loaded_concrete_model.predict(x, fhe="simulate")
 
-    assert numpy.isclose(y_pred_simulate, y_pred_simulate_loaded).all(), (
-        "Simulated predictions from the initial model do not match the ones made from the "
-        "loaded one."
-    )
+    try:
+        check_float_array_equal(y_pred_simulate, y_pred_simulate_loaded)
```
RomanBredehoft, line 1134: if you really want to give a proper message, it's better to add an `error_message` parameter in `check_float_array_equal` and pass it through that + append it to the initial error message instead of try except like this

tests/torch/test_compile_torch.py
```diff
@@ -1166,7 +1171,7 @@ def test_shape_operations_net(
         # In PTQ the results do not match because of a-priori set quantization options
         # Currently no solution for concat/reshape/transpose correctness in PTQ is proposed.
         if is_qat:
-            assert numpy.allclose(torch_output, predictions, atol=0.05, rtol=0)
+            assert check_float_array_equal(torch_output, predictions, atol=0.05, rtol=0)
```
RomanBredehoft, line 1174: same, remove the assert !

RomanBredehoft (changes requested):
a few changes are needed !

tests/onnx/test_onnx_ops_impl.py
kcelia: got and expected are not arrays, that's why we can't use check_float_array_equal

RomanBredehoft (dismissed):
thanks for this !

RomanBredehoft (approved):
thanks !

PR #382, opened 2023-10-30
chore: only apply ceil/floor op in tree_to_numpy function when necessary

RomanBredehoft:
Three minors things : 
- fix `tree_to_numpy` when we apply ceil/floor op : only apply it for the bia_1 tensor (where it's supposed to be applied)
- remove initializers associated with nodes cut from the graph
- fix comments in `tree_to_numpy.py` : sklearn uses `LessOrEqual` instead of `Greater` 

refs https://github.com/zama-ai/concrete-ml-internal/issues/196
closes https://github.com/zama-ai/concrete-ml-internal/issues/4115
closes https://github.com/zama-ai/concrete-ml-internal/issues/4116
closes https://github.com/zama-ai/concrete-ml-internal/issues/4110

cc @jfrery

### DIFF
--- a/src/concrete/ml/onnx/onnx_model_manipulations.py
+++ b/src/concrete/ml/onnx/onnx_model_manipulations.py
@@ -163,34 +163,84 @@ def remove_node_types(onnx_model: onnx.ModelProto, op_types_to_remove: List[str]
 def clean_graph_at_node_op_type(
     onnx_model: onnx.ModelProto, node_op_type: str, fail_if_not_found: bool = True
 ):
-    """Clean the graph of the onnx model by removing nodes at the given node type.
+    """Remove the first node matching node_op_type and its following nodes from the ONNX graph.
 
-    Note: the specified node_type is also removed.
+    Args:
+        onnx_model (onnx.ModelProto): The onnx model.
+        node_op_type (str): The node's op_type whose following nodes as well as itself will be
+            removed.
+        fail_if_not_found (bool): If true, raise an error if no node matched the given op_type.
+
+    Raises:
+        ValueError: If no node matched the given op_type and fail_if_not_found is set to True
+    """
+    # Get the list of node that matches the given operator type
+    nodes_of_type = [node for node in onnx_model.graph.node if node.op_type == node_op_type]
+
+    if nodes_of_type:
+
+        # Get the (first) input's same of the first node that matches the given operator type
+        input_name = nodes_of_type[0].input[0]
+
+        # Find the node that has the input_name as output name (i.e., the node tha precedes the
+        # selected node) and retrieve its name
+        node_to_cut = next(node for node in onnx_model.graph.node if node.output[0] == input_name)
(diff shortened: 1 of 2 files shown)


src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -279,12 +278,12 @@ def tree_values_preprocessing(
             # Get the preprocessed tree predictions to replace the current (non-quantized)
             # values in the onnx_model.
             init_tensor = q_y.qvalues
-        else:
```
RomanBredehoft, line 282: this was wrong, ceil/floor should be applied only on bias_1 should (but it did not create any conflicts as other weights or bias are already quantized)

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -279,12 +278,12 @@ def tree_values_preprocessing(
...
+        elif "bias_1" in initializer.name:
             if framework == "xgboost":
-                # xgboost uses "<" operator thus we must round up.
+                # xgboost uses "<" (Less) operator thus we must round up.
                 init_tensor = numpy.ceil(init_tensor)
             elif framework == "sklearn":
-                # sklearn trees use ">" operator thus we must round down.
+                # sklearn trees use "<=" (LessOrEqual) operator thus we must round down.
```
RomanBredehoft, line 286: do you confirm @jfrery ?

src/concrete/ml/onnx/onnx_model_manipulations.py
```diff
@@ -258,12 +273,19 @@ def clean_graph_after_node_op_type(
...
             output_to_follow = node.output[0]
 
-    # Once the graph has been covered and a operator was found, remove its its following nodes
+    # If the graph has been covered and an operator to remove has been found, remove it and its
+    # following nodes as well as the initializers they used as inputs
     if op_reached:
         for node in nodes_to_remove:
             onnx_model.graph.node.remove(node)
+
+        for initializer in initializers_to_remove:
```
RomanBredehoft, line 282: we now remove initializers associated with deleted nodes as well

jfrery (approved):
Thanks for the refacto! All good for me if tests are green.

bcm-at-zama:
if there was an issue with operators < or >, could we have a test which sees this issue?

RomanBredehoft:
> if there was an issue with operators < or >, could we have a test which sees this issue?

it was not an issue, just a wrong comment about it

bcm-at-zama:
> > if there was an issue with operators < or >, could we have a test which sees this issue?
> 
> it was not an issue, the comment was just wrong !

so you mean, the answer to https://github.com/zama-ai/concrete-ml-internal/issues/4110 is: it was already all good?

RomanBredehoft:
yes,we use them properly, the comment was just misleading (as explained in the issue by @jfrery )

PR #383, opened 2023-10-31
chore: correct accuracy display in XGBclassifier notebook

RomanBredehoft:
These prints are not MCC scores anymore


RomanBredehoft:
also, is is expected that this model gives around 50% accuracy score ? cc @jfrery

fd0r (approved):

jfrery (approved):

PR #393, opened 2023-11-14
feat: encrypted gemm support - 3d inputs - better rounding control

jfrery:
- handling gemm over 2 encrypted inputs (we now have our own matmul trick)
- support multiple rounding operation in one quantized op
- handle 3 dimensional input in gemm (required for training)
- quantized gemm test
- test compile and fhe simulation run based on a torch model

### DIFF
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1412,7 +1412,7 @@ class PartialQATModel(torch.nn.Module):
 
 
 class EncryptedMatrixMultiplicationModel(nn.Module):
-    """PyTorch module for performing matrix multiplication between to encrypted values."""
+    """PyTorch module for performing matrix multiplication between two encrypted values."""
 
     # pylint: disable-next=unused-argument
     def __init__(self, input_output, activation_function):
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1449,13 +1449,13 @@ class ManualLogisticRegressionTraining(torch.nn.Module):
         """Forward function for matrix multiplication.
 
         Args:
-            X (torch.Tensor)        : The training data tensor.
+            x (torch.Tensor)        : The training data tensor.
             y (torch.Tensor)        : The target tensor.
             weights (torch.Tensor)  : The weights to be learned.
             bias (torch.Tensor)     : The bias to be learned.
 
         Returns:
-            torch.Tensor: The trained weights.
+            torch.Tensor: The updated weights after performing a training step.
         """
         z = torch.bmm(x, weights) + bias
         output = torch.sigmoid(z)
@@ -1469,7 +1469,16 @@ class ManualLogisticRegressionTraining(torch.nn.Module):
 
     @staticmethod
     def predict(x, weights, bias):
-        """Predicts based on weights and bias as inputs."""
+        """Predicts based on weights and bias as inputs.
+
+        Args:
+            x (torch.Tensor)       : Input data tensor.
+            weights (torch.Tensor) : Weights tensor.
+            bias (torch.Tensor)    : Bias tensor.
+
+        Returns:
+            torch.Tensor: The predicted outputs for the given inputs.
+        """
         with torch.no_grad():
             batch_size = x.shape[0]
 
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1433,5 +1433,6 @@ class EncryptedMatrixMultiplicationModel(nn.Module):
         output = torch.matmul(input1, input1.transpose(2, 1))
         output = self.act(output)
 
-        output = output.squeeze(-1)  # Squeezes out one of the last two singleton dimensions
+        # Squeeze out one of the last two singleton dimensions
+        output = output.squeeze(-1)
         return output
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1449,10 +1449,10 @@ class ManualLogisticRegressionTraining(torch.nn.Module):
         """Forward function for matrix multiplication.
 
         Args:
-            x (torch.Tensor)        : The training data tensor.
-            y (torch.Tensor)        : The target tensor.
-            weights (torch.Tensor)  : The weights to be learned.
-            bias (torch.Tensor)     : The bias to be learned.
+            x (torch.Tensor): The training data tensor.
+            y (torch.Tensor): The target tensor.
+            weights (torch.Tensor): The weights to be learned.
+            bias (torch.Tensor): The bias to be learned.
 
         Returns:
             torch.Tensor: The updated weights after performing a training step.
@@ -1472,9 +1472,9 @@ class ManualLogisticRegressionTraining(torch.nn.Module):
         """Predicts based on weights and bias as inputs.
 
         Args:
-            x (torch.Tensor)       : Input data tensor.
-            weights (torch.Tensor) : Weights tensor.
-            bias (torch.Tensor)    : Bias tensor.
+            x (torch.Tensor): Input data tensor.
+            weights (torch.Tensor): Weights tensor.
+            bias (torch.Tensor): Bias tensor.
 
         Returns:
             torch.Tensor: The predicted outputs for the given inputs.
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1436,3 +1436,46 @@ class EncryptedMatrixMultiplicationModel(nn.Module):
         # Squeeze out one of the last two singleton dimensions
         output = output.squeeze(-1)
         return output
+
+
+class ManualLogisticRegressionTraining(torch.nn.Module):
+    """PyTorch module for performing SGD training."""
+
+    def __init__(self, learning_rate=0.1):
+        super().__init__()
+        self.learning_rate = learning_rate
+
+    def forward(self, x, y, weights, bias):
+        """Forward function for matrix multiplication.
+
+        Args:
+            X (torch.Tensor)        : The training data tensor.
+            y (torch.Tensor)        : The target tensor.
+            weights (torch.Tensor)  : The weights to be learned.
+            bias (torch.Tensor)     : The bias to be learned.
+
+        Returns:
+            torch.Tensor: The trained weights.
+        """
+        z = torch.bmm(x, weights) + bias
+        output = torch.sigmoid(z)
+
+        difference_z = output - y
+        dweights = torch.bmm(x.transpose(1, 2), difference_z) / x.size(1)
+
+        weights = weights - self.learning_rate * dweights
+
+        return weights
+
+    @staticmethod
+    def predict(x, weights, bias):
+        """Predicts based on weights and bias as inputs."""
+        with torch.no_grad():
+            batch_size = x.shape[0]
+
+            # Expand weights and bias dimensions to match the batch size of X
+            weights_expanded = weights.expand(batch_size, -1, -1)
+            bias_expanded = bias.expand(batch_size, -1, -1)
+
+            outputs = torch.sigmoid(torch.bmm(x, weights_expanded) + bias_expanded)
+        return outputs.squeeze()
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1409,3 +1409,29 @@ class PartialQATModel(torch.nn.Module):
         """
 
         return self.sub_module(x)
+
+
+class EncryptedMatrixMultiplicationModel(nn.Module):
+    """PyTorch module for performing matrix multiplication between to encrypted values."""
+
+    # pylint: disable-next=unused-argument
+    def __init__(self, input_output, activation_function):
+        super().__init__()
+        self.act = activation_function()
+
+    def forward(self, input1):
+        """Forward function for matrix multiplication.
+
+        Args:
+            input1 (torch.Tensor): The first input tensor.
+
+        Returns:
+            torch.Tensor: The result of the matrix multiplication.
+        """
+        input1 = input1.unsqueeze(1)
+
+        output = torch.matmul(input1, input1.transpose(2, 1))
+        output = self.act(output)
+
+        output = output.squeeze(-1)  # Squeezes out one of the last two singleton dimensions
+        return output
--- a/src/concrete/ml/quantization/base_quantized_op.py
+++ b/src/concrete/ml/quantization/base_quantized_op.py
@@ -936,7 +936,7 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
         self,
         x: Union[numpy.ndarray, fhe.tracing.Tracer],
         calibrate_rounding: bool,
-        rounding_operation_id: Optional[str] = None,
+        rounding_operation_id: Optional[str] = "single_rounding_op",
     ) -> numpy.ndarray:
         """Round the input array to the specified number of bits.
 
@@ -946,15 +946,16 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
                 (compute the lsbs_to_remove).
             rounding_operation_id (Optional[str]): The identifier for a specific rounding
                 operation in a quantized operation. Used to create and access the
-                lsbs_to_remove value in the dictionary. Defaults to None
+                lsbs_to_remove value in the dictionary. Defaults to "single_rounding_op"
+                if not provided.
 
         Returns:
             numpy.ndarray: The rounded array.
         """
 
-        # Initialize lsbs_to_remove if not already set
-        if not hasattr(self, "lsbs_to_remove") or self.lsbs_to_remove is None:
-            self.lsbs_to_remove = {} if rounding_operation_id else 0
+        # Ensure lsbs_to_remove is initialized as a dictionary
+        if not hasattr(self, "lsbs_to_remove") or not isinstance(self.lsbs_to_remove, dict):
+            self.lsbs_to_remove = {}
 
         if self.rounding_threshold_bits is not None and calibrate_rounding:
             # Compute lsbs_to_remove only when calibration is True
@@ -966,34 +967,18 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
                 "Can't compute lsbs_to_remove at compilation time.",
             )
 
-            # Key-based logic (dictionary)
-            if rounding_operation_id:
-                # mypy
-                assert isinstance(self.lsbs_to_remove, dict)
-                self.lsbs_to_remove[rounding_operation_id] = max(
-                    self.lsbs_to_remove.get(rounding_operation_id, 0), computed_lsbs_to_remove
-                )
-
-            # No key (integer logic)
-            else:
-                # mypy
-                assert isinstance(self.lsbs_to_remove, int)
-                self.lsbs_to_remove = max(self.lsbs_to_remove, computed_lsbs_to_remove)
+            # Update the lsbs_to_remove value in the dictionary
+            self.lsbs_to_remove[rounding_operation_id] = max(
+                self.lsbs_to_remove.get(rounding_operation_id, 0), computed_lsbs_to_remove
+            )
 
         # Rounding logic
-        if isinstance(self.lsbs_to_remove, dict):
-            lsbs_value = (
-                self.lsbs_to_remove.get(rounding_operation_id, 0) if rounding_operation_id else 0
-            )
-        else:
-            lsbs_value = self.lsbs_to_remove
+        lsbs_value = self.lsbs_to_remove.get(rounding_operation_id, 0)
 
         # mypy
         assert isinstance(lsbs_value, int)
 
         if lsbs_value > 0:
             x = fhe.round_bit_pattern(x, lsbs_to_remove=lsbs_value)
-        else:
-            self.lsbs_to_remove = None
 
         return x
--- a/src/concrete/ml/quantization/base_quantized_op.py
+++ b/src/concrete/ml/quantization/base_quantized_op.py
@@ -936,16 +936,17 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
         self,
         x: Union[numpy.ndarray, fhe.tracing.Tracer],
         calibrate_rounding: bool,
-        key: Optional[str] = None,
+        rounding_operation_id: Optional[str] = None,
     ) -> numpy.ndarray:
         """Round the input array to the specified number of bits.
 
         Args:
             x (Union[numpy.ndarray, fhe.tracing.Tracer]): The input array to be rounded.
             calibrate_rounding (bool): Whether to calibrate the rounding
-                (compute the lsbs_to_remove)
-            key (Optional[str]): The key to access the appropriate
-                lsbs_to_remove value in the dictionary.
+                (compute the lsbs_to_remove).
+            rounding_operation_id (Optional[str]): The identifier for a specific rounding
+                operation in a quantized operation. Used to create and access the
+                lsbs_to_remove value in the dictionary. Defaults to None
 
         Returns:
             numpy.ndarray: The rounded array.
@@ -953,12 +954,12 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
 
         # Initialize lsbs_to_remove if not already set
         if not hasattr(self, "lsbs_to_remove") or self.lsbs_to_remove is None:
-            self.lsbs_to_remove = {} if key else 0
+            self.lsbs_to_remove = {} if rounding_operation_id else 0
 
         if self.rounding_threshold_bits is not None and calibrate_rounding:
-            # Compute new_value only when calibration is True
+            # Compute lsbs_to_remove only when calibration is True
             current_n_bits_accumulator = compute_bits_precision(x)
-            new_value = current_n_bits_accumulator - self.rounding_threshold_bits
+            computed_lsbs_to_remove = current_n_bits_accumulator - self.rounding_threshold_bits
 
             assert_true(
                 not isinstance(x, fhe.tracing.Tracer),
@@ -966,25 +967,29 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
             )
 
             # Key-based logic (dictionary)
-            if key:
+            if rounding_operation_id:
                 # mypy
                 assert isinstance(self.lsbs_to_remove, dict)
-                self.lsbs_to_remove[key] = max(self.lsbs_to_remove.get(key, 0), new_value)
+                self.lsbs_to_remove[rounding_operation_id] = max(
+                    self.lsbs_to_remove.get(rounding_operation_id, 0), computed_lsbs_to_remove
+                )
 
             # No key (integer logic)
             else:
                 # mypy
                 assert isinstance(self.lsbs_to_remove, int)
-                self.lsbs_to_remove = max(self.lsbs_to_remove, new_value)
+                self.lsbs_to_remove = max(self.lsbs_to_remove, computed_lsbs_to_remove)
 
         # Rounding logic
         if isinstance(self.lsbs_to_remove, dict):
-            lsbs_value = self.lsbs_to_remove.get(key, 0) if key else 0
+            lsbs_value = (
+                self.lsbs_to_remove.get(rounding_operation_id, 0) if rounding_operation_id else 0
+            )
         else:
             lsbs_value = self.lsbs_to_remove
 
-        # Ensure lsbs_value is an integer
-        lsbs_value = int(lsbs_value)
+        # mypy
+        assert isinstance(lsbs_value, int)
 
         if lsbs_value > 0:
             x = fhe.round_bit_pattern(x, lsbs_to_remove=lsbs_value)
--- a/src/concrete/ml/quantization/base_quantized_op.py
+++ b/src/concrete/ml/quantization/base_quantized_op.py
@@ -865,7 +865,7 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
     Mixing operators cannot be fused to TLUs.
     """
 
-    lsbs_to_remove: Optional[int] = None
+    lsbs_to_remove: Optional[Union[int, dict]] = None
     rounding_threshold_bits: Optional[int] = None
 
     def __init__(self, *args, rounding_threshold_bits: Optional[int] = None, **kwargs) -> None:
@@ -933,7 +933,10 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
         )
 
     def cnp_round(
-        self, x: Union[numpy.ndarray, fhe.tracing.Tracer], calibrate_rounding: bool
+        self,
+        x: Union[numpy.ndarray, fhe.tracing.Tracer],
+        calibrate_rounding: bool,
+        key: Optional[str] = None,
     ) -> numpy.ndarray:
         """Round the input array to the specified number of bits.
 
@@ -941,31 +944,51 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
             x (Union[numpy.ndarray, fhe.tracing.Tracer]): The input array to be rounded.
             calibrate_rounding (bool): Whether to calibrate the rounding
                 (compute the lsbs_to_remove)
+            key (Optional[str]): The key to access the appropriate
+                lsbs_to_remove value in the dictionary.
 
         Returns:
             numpy.ndarray: The rounded array.
         """
 
-        # Rounding is applied only if specified by user
-        if self.rounding_threshold_bits is not None:
-            if calibrate_rounding:
-                assert_true(
-                    not isinstance(x, fhe.tracing.Tracer),
-                    "Can't compute lsbs_to_remove at compilation time.",
-                )
-                assert_true(
-                    self.lsbs_to_remove is None,
-                    "Rounding has already been calibrated.",
-                )
+        # Initialize lsbs_to_remove if not already set
+        if not hasattr(self, "lsbs_to_remove") or self.lsbs_to_remove is None:
+            self.lsbs_to_remove = {} if key else 0
 
-                current_n_bits_accumulator = compute_bits_precision(x)
-                self.lsbs_to_remove = current_n_bits_accumulator - self.rounding_threshold_bits
+        if self.rounding_threshold_bits is not None and calibrate_rounding:
+            # Compute new_value only when calibration is True
+            current_n_bits_accumulator = compute_bits_precision(x)
+            new_value = current_n_bits_accumulator - self.rounding_threshold_bits
 
-            # mypy
-            assert self.lsbs_to_remove is not None
+            assert_true(
+                not isinstance(x, fhe.tracing.Tracer),
+                "Can't compute lsbs_to_remove at compilation time.",
+            )
 
-            # Apply rounding if needed
-            if self.lsbs_to_remove > 0:
-                x = fhe.round_bit_pattern(x, lsbs_to_remove=self.lsbs_to_remove)
+            # Key-based logic (dictionary)
+            if key:
+                # mypy
+                assert isinstance(self.lsbs_to_remove, dict)
+                self.lsbs_to_remove[key] = max(self.lsbs_to_remove.get(key, 0), new_value)
+
+            # No key (integer logic)
+            else:
+                # mypy
+                assert isinstance(self.lsbs_to_remove, int)
+                self.lsbs_to_remove = max(self.lsbs_to_remove, new_value)
+
+        # Rounding logic
+        if isinstance(self.lsbs_to_remove, dict):
+            lsbs_value = self.lsbs_to_remove.get(key, 0) if key else 0
+        else:
+            lsbs_value = self.lsbs_to_remove
+
+        # Ensure lsbs_value is an integer
+        lsbs_value = int(lsbs_value)
+
+        if lsbs_value > 0:
+            x = fhe.round_bit_pattern(x, lsbs_to_remove=lsbs_value)
+        else:
+            self.lsbs_to_remove = None
 
         return x
--- a/src/concrete/ml/quantization/quantized_ops.py
+++ b/src/concrete/ml/quantization/quantized_ops.py
@@ -10,7 +10,7 @@ from typing import Any, Dict, Optional, Sequence, Set, Union
 import numpy
 from concrete.fhe import conv as cnp_conv
 from concrete.fhe import maxpool as cnp_maxpool
-from concrete.fhe import tag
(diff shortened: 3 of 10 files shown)


src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -174,19 +168,23 @@ def q_impl(
...
         )
 
-        q_input: QuantizedArray = prepared_inputs[0]
-        q_weights: QuantizedArray = prepared_inputs[1]
-        q_bias: Optional[QuantizedArray] = (
-            None if len(prepared_inputs) == 2 or beta == 0 else prepared_inputs[2]
-        )
+        q_input = prepared_inputs[0]
+        assert isinstance(q_input, QuantizedArray)
+        q_weights = prepared_inputs[1]
```
andrei-stoian-zama, line 184: the name "weights" works for training too but if we were to use this in a LLM we might want to change "weights" to "input2" no?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -219,18 +222,90 @@ def q_impl(
 
         p = weights_q_values.shape[0]
 
+        def enc_mul(x, y):
+            """Encrypted multiplication trick.
```
andrei-stoian-zama, line 226: it's not a trick when it's in production :) I don't see anything tricky about it anyway, let's use a more formal name

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -219,18 +222,90 @@ def q_impl(
...
+
+            Args:
+                x (numpy.ndarray): first input numpy array
+                y (numpy.ndarray): second input numpy array
+            """
+            with tag("trick"):
+                add = x + y
```
andrei-stoian-zama, line 233: this won't broadcast when shapes match no? it feels like running on a (N,N) x (N,N) matrix will not work

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -219,18 +222,90 @@ def q_impl(
...
+
+                # Initialize the output matrix
+                c = zeros(shape=(batch, m, p))
+
+                # Compute the matrix multiplication using element wise multiplications
+                for current_batch in range(batch):
+                    a_reshaped = (
+                        a[current_batch].reshape((m, n, 1)) if a.ndim == 3 else a.reshape((m, n, 1))
+                    )
+                    b_reshaped = b.reshape((1, n, p))
```
andrei-stoian-zama, line 275: or is this the trick to force broadcasting ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -219,18 +222,90 @@ def q_impl(
...
+
+            # Remove any unnecessary singleton dimensions from the output
+            c = numpy.squeeze(c, axis=0) if batch == 1 else c
+            return c
+
+        @univariate
+        def copy_function(x):
+            return x
+
+        # This is done to block the precision raising at compilation time
```
andrei-stoian-zama, line 287: nice trick !

andrei-stoian-zama (commented):
well done! pretty hardcore stuff! I'll redo the review when you take the PR out of draft

src/concrete/ml/quantization/quantized_ops.py
jfrery: Very good point. It's not weight anymore we should change this.

src/concrete/ml/quantization/quantized_ops.py
jfrery: I agree.

jfrery: Changed it.

src/concrete/ml/quantization/quantized_ops.py
jfrery: called it input2 instead

src/concrete/ml/quantization/quantized_ops.py
jfrery: I will add (N, N) tests. But it should work yes.

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -865,7 +865,7 @@ class QuantizedMixingOp(QuantizedOp, is_utility=True):
     Mixing operators cannot be fused to TLUs.
     """
 
-    lsbs_to_remove: Optional[int] = None
+    lsbs_to_remove: Optional[Union[int, dict]] = None
```
andrei-stoian-zama, line 868: I think it would have been better to simply duplicate the AutoRounder class. Then Gemm can create 3 auto rounders for the different "keys". 

It's a bit of a hassle though, let me know what you think

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,62 @@ def make_output_quant_parameters(
...
+                self.lsbs_to_remove = max(self.lsbs_to_remove, new_value)
+
+        # Rounding logic
+        if isinstance(self.lsbs_to_remove, dict):
+            lsbs_value = self.lsbs_to_remove.get(key, 0) if key else 0
+        else:
+            lsbs_value = self.lsbs_to_remove
+
+        # Ensure lsbs_value is an integer
+        lsbs_value = int(lsbs_value)
```
andrei-stoian-zama, line 987: shouldn't this be an assert isinstance ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -174,38 +168,49 @@ def q_impl(
...
         prepared_inputs = self._prepare_inputs_with_constants(
             *q_inputs, calibrate=False, quantize_actual_values=True
         )
 
-        q_input: QuantizedArray = prepared_inputs[0]
-        q_weights: QuantizedArray = prepared_inputs[1]
-        q_bias: Optional[QuantizedArray] = (
-            None if len(prepared_inputs) == 2 or beta == 0 else prepared_inputs[2]
-        )
+        q_input = prepared_inputs[0]
```
andrei-stoian-zama, line 182: maybe `q_input1` ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +222,113 @@ def q_impl(
         # Here we follow Eq.7 in https://arxiv.org/abs/1712.05877 to split the core computation
         # from the zero points and scales.
 
-        p = weights_q_values.shape[0]
+        p = input2_q_values.shape[-2]
+
+        # FIXME: https://github.com/zama-ai/concrete-internal/issues/512
```
andrei-stoian-zama, line 227: can you plase use the regular format:

```
# comment explaining the problem
# FIXME: issue id
```

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +222,113 @@ def q_impl(
...
+                add_pow = (add.astype(numpy.float64)) ** 2
+                sub_pow = (sub.astype(numpy.float64)) ** 2
+                add_pow_divide = (add_pow / 4.0).astype(numpy.int64)
+                sub_pow_divide = (sub_pow / 4.0).astype(numpy.int64)
+
+            return add_pow_divide - sub_pow_divide
+
+        # FIXME: https://github.com/zama-ai/concrete-internal/issues/512
+        def matmul(a, b):
+            with tag("Encrypted MatMul"):
```
andrei-stoian-zama, line 257: tags with space.. ugh. please please use lowercase and `_`

andrei-stoian-zama, line 257: maybe explain what's going on in this function in a docstring

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +222,113 @@ def q_impl(
...
         # The following check avoids the computation of the sum of the inputs, which may have
         # large bit-width, in the case where it would be multiplied by zero
-        if q_weights.quantizer.zero_point != 0:
+        if q_input2.quantizer.zero_point != 0:
             # Sum operation in full integers resulting in large integers (INTEGERS)
             with tag(self.op_instance_name + ".matmul_inputsum"):
-                sum_input = -q_weights.quantizer.zero_point * numpy.sum(
-                    input_q_values, axis=1, keepdims=True
+                sum_input = -q_input2.quantizer.zero_point * numpy.sum(
+                    input_q_values_sum, axis=-1, keepdims=True
```
andrei-stoian-zama, line 331: `input_q_values_sum` -> `input_q_values_copy` ? it would be clearer

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -243,22 +341,28 @@ def q_impl(
             # pylint: disable-next=unsubscriptable-object
             self.debug_value_tracker[self.op_instance_name]["output"] = numpy_q_out  # type: ignore
 
-        # sum_weights is a constant
-        sum_weights = q_input.quantizer.zero_point * numpy.sum(
-            weights_q_values, axis=0, keepdims=True
+        input2_q_values_sum = (
```
andrei-stoian-zama, line 344: same as above

src/concrete/ml/quantization/base_quantized_op.py
jfrery: We said we would not use the auto rounder as it's not possible to use rounding during calibration.

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Yes changed it. This is for mypy.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Done

src/concrete/ml/quantization/quantized_ops.py
jfrery: Added the issue.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Done

jfrery: Agreed

src/concrete/ml/quantization/quantized_ops.py
jfrery: Alright

src/concrete/ml/quantization/quantized_ops.py
jfrery: Yes

src/concrete/ml/pytest/torch_models.py
```diff
@@ -1409,3 +1409,29 @@ def forward(self, x):
...
+
+        Returns:
+            torch.Tensor: The result of the matrix multiplication.
+        """
+        input1 = input1.unsqueeze(1)
+
+        output = torch.matmul(input1, input1.transpose(2, 1))
+        output = self.act(output)
+
+        output = output.squeeze(-1)  # Squeezes out one of the last two singleton dimensions
```
RomanBredehoft, line 1436: small detail but I think we prefer to put comments above the line instead of having them inline like this

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,62 @@ def make_output_quant_parameters(
         )
 
     def cnp_round(
-        self, x: Union[numpy.ndarray, fhe.tracing.Tracer], calibrate_rounding: bool
+        self,
+        x: Union[numpy.ndarray, fhe.tracing.Tracer],
+        calibrate_rounding: bool,
+        key: Optional[str] = None,
```
RomanBredehoft, line 939: I don't think key is a good naming here, what is it representing ? even reading the docstring does not make it learer, I would prefer a more explicit name if possible !

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,62 @@ def make_output_quant_parameters(
...
-                    not isinstance(x, fhe.tracing.Tracer),
-                    "Can't compute lsbs_to_remove at compilation time.",
-                )
-                assert_true(
-                    self.lsbs_to_remove is None,
-                    "Rounding has already been calibrated.",
-                )
+        # Initialize lsbs_to_remove if not already set
+        if not hasattr(self, "lsbs_to_remove") or self.lsbs_to_remove is None:
+            self.lsbs_to_remove = {} if key else 0
```
RomanBredehoft, line 956: not sure I like the fact that we constantly handle either an int or a dict honestly, couldn't we just make it a dict in any case ( with a default "key" when considering the "single int" case) ?

src/concrete/ml/pytest/torch_models.py
fd0r: Not sure if we have a rule on this 🤷🏼

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,62 @@ def make_output_quant_parameters(
...
 
-                current_n_bits_accumulator = compute_bits_precision(x)
-                self.lsbs_to_remove = current_n_bits_accumulator - self.rounding_threshold_bits
+        if self.rounding_threshold_bits is not None and calibrate_rounding:
+            # Compute new_value only when calibration is True
```
RomanBredehoft, line 959: what does `new_value` represent here ? maybe make either the name or the comment more explicit here as well

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,62 @@ def make_output_quant_parameters(
...
 
-            # mypy
-            assert self.lsbs_to_remove is not None
+            assert_true(
+                not isinstance(x, fhe.tracing.Tracer),
+                "Can't compute lsbs_to_remove at compilation time.",
+            )
+
+            # Key-based logic (dictionary)
+            if key:
```
RomanBredehoft, line 969: if we keep this dict / int duality, then better change this to `if key is not None` !

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: what does mypy say ? I'm surprised you need to make sure it's an int specifically here (instead of making sure it's not None)

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,62 @@ def make_output_quant_parameters(
...
+
+        # Ensure lsbs_value is an integer
+        lsbs_value = int(lsbs_value)
+
+        if lsbs_value > 0:
+            x = fhe.round_bit_pattern(x, lsbs_to_remove=lsbs_value)
+        else:
+            self.lsbs_to_remove = None
 
         return x
```
RomanBredehoft, line 994: if `lsbs_value <= 0`, then x is not defined right ? I'm surprised mypy does not complain here, but if that's correct I would change the above code by something like : 
```
        if lsbs_value > 0:
            x = fhe.round_bit_pattern(x, lsbs_to_remove=lsbs_value)
            return x
        
        self.lsbs_to_remove = None
```

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -134,9 +134,9 @@ def __init__(
         self,
         n_bits_output: int,
         op_instance_name: str,
-        int_input_names: Set[str] = None,
+        int_input_names: Optional[Set[str]] = None,
```
RomanBredehoft, line 137: yes great thanks

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -174,38 +168,49 @@ def q_impl(
         alpha = self.attrs.get("alpha", 1)
         beta = self.attrs.get("beta", 1)
 
+        # If self.constant_inputs is empty this is an encrypted gemm
+        is_encrypted_gemm = isinstance(self.constant_inputs, dict) and not self.constant_inputs
```
RomanBredehoft, line 172: there's is no better way to know this ? this does not seem very robust honestly

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -174,38 +168,49 @@ def q_impl(
...
+
         # If alpha != 1 or beta not in [0, 1], this function must be modified
         assert_true(alpha == 1)
-        assert_true(beta in [0, 1])
+        assert_true(beta in {0, 1})
```
RomanBredehoft, line 176: why this change ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -174,38 +168,49 @@ def q_impl(
...
-        q_input: QuantizedArray = prepared_inputs[0]
-        q_weights: QuantizedArray = prepared_inputs[1]
-        q_bias: Optional[QuantizedArray] = (
-            None if len(prepared_inputs) == 2 or beta == 0 else prepared_inputs[2]
-        )
+        q_input = prepared_inputs[0]
+        assert isinstance(q_input, QuantizedArray)
+        q_input2 = prepared_inputs[1]
+        assert isinstance(q_input2, QuantizedArray)
+        q_bias = None if len(prepared_inputs) == 2 or beta == 0 else prepared_inputs[2]
```
RomanBredehoft, line 186: maybe this is worth a comment here, more particularly : why is `or beta == 0 ` needed here ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -174,38 +168,49 @@ def q_impl(
...
+        assert isinstance(q_bias, (type(None), QuantizedArray))
```
RomanBredehoft, line 187: `type(None)` ? that really works ? I'm learning things everyday

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: I agree with Andrei, please add a comment to explain why we have a fixme here and what we should change at one point

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: also would love several small comments throughout the function to explain several major steps , it's currently not easy to follow everything !

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +222,113 @@ def q_impl(
...
+                    enc_mul_result = enc_mul(a_reshaped, b_reshaped)
+                    c[i] = numpy.sum(enc_mul_result, axis=1)
+
+                # If both inputs were 2D, remove the first dimension
+                if not a_3d and not b_3d:
+                    c = numpy.squeeze(c, axis=0)
+
+                return c
+
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4127
```
RomanBredehoft, line 298: same please

src/concrete/ml/pytest/torch_models.py
jfrery: Alright I will put it above the line. Should be better to match our 100 chars limit.

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +222,113 @@ def q_impl(
...
+        @univariate
+        def copy_function(x):
+            return x
+
+        # This is done to block the precision raising at compilation time
+        # Here we add a PBS to protect inputs precision. We don't want to do it
```
RomanBredehoft, line 304: maybe explain a bit more how it protects the input precision ? I'm not sure I get it

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft, line 331: is this `axis=-1` change expected ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -281,13 +385,26 @@ def q_impl(
...
 
-        with tag(self.op_instance_name + ".matmul_rounding"):
-            # Apply Concrete rounding (if relevant)
-            numpy_q_out = self.cnp_round(numpy_q_out, calibrate_rounding)
-
-        # Quantization scales and zero points (FLOATS involved)
-        # This is going to be compiled with a PBS (along with the following activation function)
-        numpy_q_out = numpy_q_out.astype(numpy.float64) + final_term - sum_weights
+        # If weights are not encrypted then we can round as the next
+        # line is going ot be done in a PBS
```
RomanBredehoft, line 389: small typo : `ot` -> `to`

src/concrete/ml/quantization/base_quantized_op.py
jfrery: A single quantized op can now have multiple rounding operations. The key refers to a specific rounding operation in a quantized op. I can call it `rounding_key` or `rounding_operation_id` maybe. Not sure it helps much more. I can edit the docstring to make it clearer maybe.

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -281,13 +385,26 @@ def q_impl(
...
+            with tag(self.op_instance_name + ".matmul_rounding"):
+                # Apply Concrete rounding (if relevant)
+                numpy_q_out = self.cnp_round(numpy_q_out, calibrate_rounding)
+
+                # Force a PBS with astype float64
+                numpy_q_out = numpy_q_out.astype(numpy.float64)
+
+        # Quantization scales and zero points
+        # This is done in a PBS if is_encrypted_gemm == False
+        # Otherwise it is done in clear and be compiled with a PBS
```
RomanBredehoft, line 400: >        # Otherwise it is done in clear and be compiled with a PBS
        # (along with the following activation function)

done in the clear and "be compiled with a PBS" ? I think there's something wrong in this comment, or I don't get it

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Yes I can probably have a default name that is used when key is not passed.

RomanBredehoft (changes requested):
Thank a lot for this, I have several comments (mostly namings + adding some comments ) !

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Maybe `computed_lsbs_to_remove`?

src/concrete/ml/quantization/base_quantized_op.py
jfrery: It's the same isn't it?

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Not sure what it brings but yes I can make that change if you want.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Not currently. We would need a bigger PR that propagate the encryption status. I agree, this might be a problem if one of the input is in clear. cc @fd0r

src/concrete/ml/quantization/quantized_ops.py
jfrery: Performance wise, checking if a value is in a dict is much better than checking if it's in a list. Here it's only 2 values so I guess it's doesn't change much but still good practice to have a dict when doing such comparison I think. That's a change made by @fd0r btw

src/concrete/ml/quantization/base_quantized_op.py
jfrery: It's the same for mypy. Making sure it's a int or making sure it's not a None. I can change to check it's not a None if you want.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Yes I am going to push sorry. I answer based on my local changed. I then call for a re-review.

src/concrete/ml/quantization/quantized_ops.py
jfrery: This was there before. Not sure what this means without looking deeper.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Looks like it. That's a @fd0r trick.

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +222,113 @@ def q_impl(
...
+                    sub = self.cnp_round(sub, calibrate_rounding, key="sub")
+
+                add_pow = (add.astype(numpy.float64)) ** 2
+                sub_pow = (sub.astype(numpy.float64)) ** 2
+                add_pow_divide = (add_pow / 4.0).astype(numpy.int64)
+                sub_pow_divide = (sub_pow / 4.0).astype(numpy.int64)
+
+            return add_pow_divide - sub_pow_divide
+
+        # FIXME: https://github.com/zama-ai/concrete-internal/issues/512
```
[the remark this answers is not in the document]
jfrery: Done

src/concrete/ml/quantization/quantized_ops.py
jfrery: Done

src/concrete/ml/quantization/quantized_ops.py
jfrery: Done

src/concrete/ml/quantization/quantized_ops.py
jfrery: Yes. The sum should happen on the last axis. In 2d it was 1 but that doesn't generalize to 3d. -1 does.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Thanks corrected

src/concrete/ml/quantization/quantized_ops.py
jfrery: Indeed very unclear. I meant, "Otherwise it is done in FHE". The end must be something I forgot to remove. Thanks!

src/concrete/ml/quantization/quantized_ops.py
jfrery: Let me add an issue here.

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Actually no it makes the qnn tests break...

tests/torch/test_training.py
jfrery, line 1: New training test

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: I would indeed prefer to check that it's not a None + add a small "#For mypy" comment for example yes !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: ok got it thanks for the issue !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: you're right indeed, but yes for 2 values it's not important

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: maybe @andrei-stoian-zama knows ?

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: still don't see a comment here @jfrery 😅

src/concrete/ml/quantization/base_quantized_op.py
jfrery: done

src/concrete/ml/quantization/quantized_ops.py
jfrery: Done

src/concrete/ml/quantization/quantized_ops.py
jfrery: Should be good now

src/concrete/ml/quantization/quantized_ops.py
fd0r: Yes this is needed because python changed its behavior on `NoneType` between versions that we support.
See: https://stackoverflow.com/questions/21095654/what-is-a-nonetype-object

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Changed the name and updated doc string

src/concrete/ml/quantization/quantized_ops.py
jfrery: Added an explanation

src/concrete/ml/quantization/quantized_ops.py
fd0r: Y = alpha * A' * B' + beta * C

Thus why we need beta == 0

src/concrete/ml/pytest/torch_models.py
```diff
@@ -1409,3 +1409,82 @@ def forward(self, x):
         """
 
         return self.sub_module(x)
+
+
+class EncryptedMatrixMultiplicationModel(nn.Module):
+    """PyTorch module for performing matrix multiplication between to encrypted values."""
```
kcelia, line 1415: two ?

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,67 @@ def make_output_quant_parameters(
...
-            # mypy
-            assert self.lsbs_to_remove is not None
+        # mypy
+        assert isinstance(lsbs_value, int)
 
-            # Apply rounding if needed
-            if self.lsbs_to_remove > 0:
-                x = fhe.round_bit_pattern(x, lsbs_to_remove=self.lsbs_to_remove)
+        if lsbs_value > 0:
+            x = fhe.round_bit_pattern(x, lsbs_to_remove=lsbs_value)
```
kcelia, line 995: @jfrery, since you are manually computing the nbr of lsb to remove,  I'm wondering if you encounter an overflow.

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +225,163 @@ def q_impl(
...
+                sub = x - y
+
+                # Apply Concrete rounding to the addition and substraction
+                with tag(self.op_instance_name + ".pbs_matmul_rounding_add"):
+                    add = self.cnp_round(add, calibrate_rounding, rounding_operation_id="add")
+                with tag(self.op_instance_name + ".pbs_matmul_rounding_sub"):
+                    sub = self.cnp_round(sub, calibrate_rounding, rounding_operation_id="sub")
+
+                # Square the rounded sums and differences, and divide by 4
+                add_pow = (add.astype(numpy.float64)) ** 2
```
kcelia, line 265: Could you explain why you have to cast it to float first ?
For me it's already an integer.

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -217,20 +225,163 @@ def q_impl(
...
+                sub_pow = (sub.astype(numpy.float64)) ** 2
+                add_pow_divide = (add_pow / 4.0).astype(numpy.int64)
```
kcelia, line 267: would be better to do: (add_pow - sub_pow) / 4, to have only one div ?

src/concrete/ml/pytest/torch_models.py
```diff
@@ -1409,3 +1409,82 @@ def forward(self, x):
...
+
+        output = torch.matmul(input1, input1.transpose(2, 1))
+        output = self.act(output)
+
+        # Squeeze out one of the last two singleton dimensions
+        output = output.squeeze(-1)
+        return output
+
+
+class ManualLogisticRegressionTraining(torch.nn.Module):
```
RomanBredehoft, line 1441: is it expected to add training in this PR? it was only for enc GEMM , I don't think it's a good idea to make this PR even more complex

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: ah my bad I didn't get that `x` was one of the method's parameter, no issues then (I thought it was undefined and thus would crash)

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: ah great, worth to add it n a comment here then

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: got it, but just to make sure, it's strictly identical to having `assert q_bias is None or isinstace(q_bias, QuantizedArray)` right ?

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: thanks !

tests/torch/test_training.py
```diff
@@ -0,0 +1,127 @@
+"""Tests for FHE training."""
```
RomanBredehoft, line 1: again here, are we sure we want to add training feature/tests in this PR ? I think I would have preferred it to be separated

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: many thanks !

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: what's the issue ?

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: yes great thanks !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: many thanks for that yes !!

RomanBredehoft (changes requested):
many thanks for the changes ! I still have a few minor questions + do we really want to add an FHE training example/test in this PR ? I feel like it would better if separated from it

src/concrete/ml/pytest/torch_models.py
jfrery: This test will be important in the following PRs. Better have it now.

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: yes but that could be easily changed no ? like testing we have the specific key instead of an int

not saying it's the best solution though, but I feel like handling the same attribute with two quite different types brings some avoidable complexity tothe base code

src/concrete/ml/quantization/quantized_ops.py
jfrery: Casting to floats ensure this step and the following ones are done within a PBS until a cast to integer. This is to make sure operations are all fused within a PBS.

src/concrete/ml/quantization/quantized_ops.py
jfrery: We can't do that. Here we cast to floating points so both values are inputs to a PBS. A PBS can't have 2 encrypted inputs. So here we are basically building two different PBS.

tests/torch/test_training.py
jfrery: We need a test that makes sure we keep that test working all along training in FHE integration.

src/concrete/ml/pytest/torch_models.py
RomanBredehoft: ok sure if no one fights for it then good for me

src/concrete/ml/quantization/quantized_ops.py
fd0r: yes but a bit more compact imo

src/concrete/ml/quantization/base_quantized_op.py
andrei-stoian-zama: I'm actually saying you could have implemented a new AutoRounder class in Concrete ML that implements this functionality during calibration. Then you would have had 3 instances of it in the Gemm class

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -933,39 +933,67 @@ def make_output_quant_parameters(
         )
 
     def cnp_round(
-        self, x: Union[numpy.ndarray, fhe.tracing.Tracer], calibrate_rounding: bool
+        self,
+        x: Union[numpy.ndarray, fhe.tracing.Tracer],
+        calibrate_rounding: bool,
+        rounding_operation_id: Optional[str] = None,
```
andrei-stoian-zama, line 939: and here you would take a CMLAutoRounder

andrei-stoian-zama (dismissed):
Great work! I think there is a small refactoring opportunity by creating our own autorounder class which could then be used for other uses such as trees.

src/concrete/ml/pytest/torch_models.py
jfrery: Fixed

src/concrete/ml/quantization/base_quantized_op.py
jfrery: The overflow thing is computed afterward as far as I know. Here I do not handle it.

src/concrete/ml/quantization/base_quantized_op.py
jfrery: I am not sure what this would bring? The current function is fairly simple and the only thing we need to compute and store is lsbs_to_remove. There are no specific parameters to this. Anyway I don't think this PR is the place to introduce a new auto rounder class. Let's make an issue to discuss if we want to move toward this.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Added a comment

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Alright I made it a dict only.

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: great thanks  !

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: thanks a lot ! I think you still need to update the typing in the class' definition :`    lsbs_to_remove: Optional[Union[int, dict]] = None` (should not be a union anymore)

but apart from that awesome

jfrery: Nice catch. I will update before the merge.

andrei-stoian-zama (approved):

kcelia (approved):

PR #394, opened 2023-11-17
docs: update Apple Silicon install information

aquint-zama:


docs/getting-started/pip_installing.md
```diff
@@ -38,33 +38,3 @@ pip install concrete-ml
 ```
 
 This will automatically install all dependencies, notably Concrete.
-
-## Using Docker
```
bcm-at-zama, line 42: why do you remove docker?
and don't you want to link to https://community.zama.ai/t/troubleshooting-concrete-installation-on-apple-silicon/577?

fd0r: Feels weird to have that in pip installing but also would be best to keep this in the documentation somewhere.

bcm-at-zama: for the name of the file: indeed, it's not perfect, it's more for historical reasons
but if we rename or create a new file, we're going to break links, which is not great

aquint-zama: I removed it as it is not linked anywhere else in the doc and no longer needed. I'm not saying it's not interesting but it was more a workaround for Apple silicon than a real need, for me it would be easier to completely remove it (also less case to handle from a support point of view) but your call?

andrei-stoian-zama: I think we should keep it. Many people use docker as a way to avoid changing their local environment. Sometimes a virtualenv is not enough

andrei-stoian-zama (changes requested):

docs/getting-started/pip_installing.md
aquint-zama: is it tested?

aquint-zama: kept the Docker part

fd0r (dismissed):

bcm-at-zama:
@aquint-zama @andrei-stoian-zama : shouldn't we add a link to https://community.zama.ai/t/troubleshooting-concrete-installation-on-apple-silicon/577?

aquint-zama:
> @aquint-zama @andrei-stoian-zama : shouldn't we add a link to https://community.zama.ai/t/troubleshooting-concrete-installation-on-apple-silicon/577?

We could add it!

fd0r (approved):

RomanBredehoft (approved):
thanks !

bcm-at-zama (commented):
thx @aquint-zama

andrei-stoian-zama (approved):

aquint-zama:
I let you guys merge, I can't

PR #396, opened 2023-11-24
feat: add multi-output support

fd0r:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4139


Makefile
```diff
@@ -32,30 +32,30 @@ setup_env:
 	@# The keyring install is to allow pip to fetch credentials for our internal repo if needed
 	PIP_INDEX_URL=https://pypi.org/simple \
 	PIP_EXTRA_INDEX_URL=https://pypi.org/simple \
-	poetry run python --version
-	poetry run python -m pip install keyring
-	poetry run python -m pip install -U pip wheel
+	python -m poetry run python --version
```
jfrery, line 35: hmm isnt the poetry run a safe execution to make aure we execute in the right project's environment?

fd0r: idk it broke on my systen but I'll revert this in the end yes

benchmarks/deep_learning.py
```diff
@@ -705,7 +705,7 @@ def evaluate_pre_trained_cnn_model(dataset: str, cnn_class: type, config: dict,
         metric_label_prefix="Torch fp32",
     )
 
-    circuit_bitwidth = fhe_module.fhe_circuit.graph.maximum_integer_bit_width()
+    circuit_bitwidth = fhe_module.fhe_circuit.graph.maximum_integer_bit_width(is_encrypted_filter=True)
```
jfrery, line 708: What is going on here? The default has changed?

fd0r: Yes the behavior changed, if we want to keep the behavior that we expect this is what we need to do.

benchmarks/deep_learning.py
```diff
@@ -705,7 +705,9 @@ def evaluate_pre_trained_cnn_model(dataset: str, cnn_class: type, config: dict,
         metric_label_prefix="Torch fp32",
     )
 
-    circuit_bitwidth = fhe_module.fhe_circuit.graph.maximum_integer_bit_width()
+    circuit_bitwidth = fhe_module.fhe_circuit.graph.maximum_integer_bit_width(
+        is_encrypted_filter=True
+    )
```
jfrery, line 710: So we have to put this is_encrypted_filter = True, everywhere now? Shall we build our own API to simplify this?

e.g. `model.max_bitwidth` ?

Makefile
```diff
@@ -19,7 +19,7 @@ OPEN_PR="true"
 # Force the installation of a Concrete Python version, which is very useful with nightly versions
 # /!\ WARNING /!\: This version should NEVER be a wildcard as it might create some
 # issues when trying to run it in the future.
-CONCRETE_PYTHON_VERSION="concrete-python==2.5.0rc1"
+CONCRETE_PYTHON_VERSION="concrete-python==2023.12.5"
```
jfrery, line 22: Maybe it's worth to directly switch to 2023.12.6 which contains the truncate?

src/concrete/ml/onnx/onnx_impl_utils.py
```diff
@@ -14,7 +15,7 @@ def numpy_onnx_pad(
     pads: Tuple[int, ...],
     pad_value: Union[float, int, numpy.ndarray] = 0,
     int_only: bool = False,
-) -> numpy.ndarray:
+) -> Union[numpy.ndarray, Tracer]:
```
jfrery, line 18: Why do we need this now?

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1265,24 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+def test_multi_output(model_object, default_configuration):
+    """Test fancy indexing torch."""
+    model = model_object()
+    x = numpy.random.randint(0, 2, size=(100, 3, 10)).astype(numpy.float64)
+    y = numpy.random.randint(0, 2, size=(100, 3, 10)).astype(numpy.float64)
+    quantized_module = compile_torch_model(
+        model, (x, y), n_bits=4, configuration=default_configuration
+    )
+    quantized_module.forward(x[[0]], y[[0]])
+    quantized_module.forward(x[[0]], y[[0]], fhe="simulate")
```
jfrery, line 1283: Could we check that we have indeed two output quantizers and that they are applied correctly?

Makefile
```diff
@@ -47,7 +47,7 @@ setup_env:
 	fi
 
 	echo "Installing $(CONCRETE_PYTHON_VERSION)" && \
-	poetry run python -m pip install -U --pre "$(CONCRETE_PYTHON_VERSION)" --no-deps
+	poetry run python -m pip install -U --pre "$(CONCRETE_PYTHON_VERSION)"
```
bcm-at-zama, line 50: there was not a reason for us to have `--no-deps`?

jfrery (commented):
Well done for integrating this! Nice feature. I am not super happy about all the `is_encrypted_filter=True` however. Feels a bit boring from a user POV.

Can we do something about it? Also we should definitely document why we need to do this now. I still don't get it. What's the difference between True and False here?

docs/advanced_examples/DecisionTreeClassifier.ipynb
```diff
@@ -235,7 +235,8 @@
     }
    ],
    "source": [
-    "print(f\"Generating a key for an {circuit.graph.maximum_integer_bit_width()}-bit circuit\")"
+    "max_bit_width = circuit.graph.maximum_integer_bit_width(is_encrypted_filter=True)\n",
```
bcm-at-zama, line 238: what is `is_encrypted_filter=True` and shouldn't it be the default in CP? Looks like you've changed it in lot of places (or everywhere?)

benchmarks/deep_learning.py
fd0r: I guess having our own API makes sense in this case yes.

Makefile
fd0r: It wasn't available at the moment of this PR, it can be done but this takes some time between updating licences, squashing and the CI

tests/torch/test_compile_torch.py
fd0r: I can check that we have two quantizers but not sure how you want to assert that they are applied correctly

docs/advanced_examples/DecisionTreeClassifier.ipynb
fd0r: Everywhere in our code base.

I don't know but this is definitely a change in behavior from Concrete.

bcm-at-zama: maybe debate with @umut-sahin , because if you have to change, certainly the other users have to, and it's boring for them

umut-sahin: The default behavior didn't change, you can safely remove it as long as you don't use `show_assigned_bit_widths=True`, and this setting is a new feature so it's not breaking anything.

Makefile
RomanBredehoft: weird but ok

benchmarks/deep_learning.py
RomanBredehoft: yes if we have no choice I'm definitively in favor of having our own API for this ! It's been quite some time I'm already exhausted of having to remember where to find this info

I see two possible solutions : 
- have a utility function like `get_max_bit_width(fhe_circuit)`. But that means we need to import it each time
- have `model.max_bit_width` like @jfrery suggests. This should probably be a property parameter in the BaseEstimator, before we need to check that the model has been properly compiled. But that means we're adding a new attribute, that we'll need to maintain in the future (I can already see this max bit_width becoming more and more obsolete, so not sure how this solution is relevant in the long term)

not sure which one I prefer for now honestly !

Makefile
jfrery: Got it let's do it in the truncate PR then thanks!

src/concrete/ml/onnx/onnx_impl_utils.py
jfrery: Got it. Makes sense to have them here.

benchmarks/deep_learning.py
fd0r: Alright so the conclusion is just to go back to the old behavior for now.

Makefile
fd0r: reverted.

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -293,7 +286,14 @@ def forward(
         *x: numpy.ndarray,
         fhe: Union[FheMode, str] = FheMode.DISABLE,
         debug: bool = False,
-    ) -> Union[numpy.ndarray, Tuple[numpy.ndarray, Optional[Dict[Any, Any]]]]:
+    ) -> Union[
+        numpy.ndarray,
+        Tuple[numpy.ndarray, ...],
+        Tuple[
```
RomanBredehoft, line 292: this one starts to get crazy

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -337,26 +337,31 @@ def forward(
...
             ] = {}
             for (_, layer) in self.quant_layers_dict.values():
                 layer.debug_value_tracker = debug_value_tracker
-            result = self.quantized_forward(*q_x, fhe="disable")
+            q_y_pred = self.quantized_forward(*q_x, fhe="disable")
             for (_, layer) in self.quant_layers_dict.values():
                 layer.debug_value_tracker = None
-            return result, debug_value_tracker
+            # De-quantize the output predicted values
+            y_pred = self.dequantize_output(
```
RomanBredehoft, line 349: are you sure we want to de-quant here ? I believe the debug option was made to get the quantized values no the floating points ? cc @andrei-stoian-zama

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -337,26 +337,31 @@ def forward(
...
+                *([q_y_pred] if isinstance(q_y_pred, numpy.ndarray) else q_y_pred)
+            )
+            return y_pred, debug_value_tracker
 
         q_y_pred = self.quantized_forward(*q_x, fhe=fhe)
 
         # De-quantize the output predicted values
-        y_pred = self.dequantize_output(q_y_pred)
-
+        y_pred = self.dequantize_output(
```
RomanBredehoft, line 357: this seems quite impractical and I feel we could have a nicer way of doing it ! But at the same time, I'm not sure what could be the best, the others possibilities would be : 
- make `quantized_forward` return a tuple, even with a single value, and make `dequantize_output` only handle tuple of arrrays as input
- keep `quantized_forward` like you did, but then make `dequantize_output` handle numpy array or tuple of numpy array as input, and handle that and use the `to_tuple` utility function that we have

I think I prefer the last option as it's already something we do in the `forward` for example (since `quantize_inputs` have this similar behavior of removing the tuple type if there's a single value

worse case you might want to use `to_tuple` here instead of this custom if/else statement !

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -388,19 +394,22 @@ def quantized_forward(
         )
 
         if fhe == "disable":
-            return self._clear_forward(*q_x)
-
-        simulate = fhe == "simulate"
-        return self._fhe_forward(*q_x, simulate=simulate)
+            result = self._clear_forward(*q_x)
```
RomanBredehoft, line 397: not sure I understand this change ?

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -434,7 +443,7 @@ def _clear_forward(self, *q_x: numpy.ndarray) -> numpy.ndarray:
                 # The error message contains the ONNX tensor name that
                 # triggered this error
                 for input_idx in error_tracker:
-                    bad_qat_ops.append((input_names[input_idx], layer.__class__.op_type()))
+                    bad_qat_ops.append((input_names[input_idx], str(layer.__class__.op_type())))
```
RomanBredehoft, line 446: mypy I guess ?

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -445,14 +454,19 @@ def _clear_forward(self, *q_x: numpy.ndarray) -> numpy.ndarray:
...
 
-        assert_true(len(output_quantized_arrays) == 1)
-
         # The output of a graph must be a QuantizedArray
-        assert isinstance(output_quantized_arrays[0], QuantizedArray)
+        assert all(isinstance(elt, QuantizedArray) for elt in output_quantized_arrays)
 
-        return output_quantized_arrays[0].qvalues
+        results = tuple(
+            elt.qvalues for elt in output_quantized_arrays if isinstance(elt, QuantizedArray)
```
RomanBredehoft, line 461: is the `if isinstance(elt, QuantizedArray)` part really necessary ? I imagine this may be mypy but you have an assert for it right above

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -501,10 +514,23 @@ def _fhe_forward(self, *q_x: numpy.ndarray, simulate: bool = True) -> numpy.ndar
             # Execute the forward pass in FHE or with simulation
             q_result = predict_method(*q_input)
 
-            results_cnp_circuit_list.append(q_result)
-
-        results_cnp_circuit = numpy.concatenate(results_cnp_circuit_list, axis=0)
-
+            if isinstance(q_result, tuple):
```
RomanBredehoft, line 517: similar to above, I think it would be simpler if you use the `to_tuple` function here , which would make the code look like : 
```
q_result = to_tuple(q_result)

results_cnp_circuit_list = [[] for _ in q_result]

for elt_index, elt in enumerate(q_result):
    results_cnp_circuit_list[elt_index].append(elt)
```

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -532,22 +558,30 @@ def quantize_input(self, *x: numpy.ndarray) -> Union[numpy.ndarray, Tuple[numpy.
...
 
         Returns:
             numpy.ndarray: De-quantized output values of the last layer.
         """
-        y_preds = tuple(
-            output_quantizer.dequant(q_y_preds) for output_quantizer in self.output_quantizers
+        # Make sure that we have as many predictions as quantizers
+        assert len(q_y_preds) == len(self.output_quantizers), (
+            f"{len(q_y_preds)=} != {len(self.output_quantizers)=} "
+            "but the number of outputs to de-quantize should match the number of quantizers."
```
RomanBredehoft, line 575: small detail but I would maybe specificy ` the number of output quantizers` to be more precise

src/concrete/ml/sklearn/base.py
```diff
@@ -1171,7 +1172,9 @@ def compile(
     def _inference(self, q_X: numpy.ndarray) -> numpy.ndarray:
         self.check_model_is_fitted()
 
-        return self.quantized_module_.quantized_forward(q_X)
+        result = self.quantized_module_.quantized_forward(*[q_X])
```
RomanBredehoft, line 1175: I don't think this is necessary right ? `quantized_forward` should handle single-input fine !

tests/torch/test_compile_torch.py
RomanBredehoft: yes I agree ! We should at least check that we actually have two outputs from this forward !

RomanBredehoft (changes requested):
Thanks for this feature, looking forward to test it ! Besides this `is_encrypted_filter` thing, I have several comments on how to improve some parts !

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: I'm not the one that uses this feature the most honestly so not sure !

src/concrete/ml/quantization/quantized_module.py
fd0r: yes

src/concrete/ml/quantization/quantized_module.py
fd0r: yes type hint again

fd0r:
Thanks a lot for the feedback @RomanBredehoft I'll address your comments.

src/concrete/ml/quantization/quantized_module.py
fd0r: Maybe @andrei-stoian-zama can tell us

src/concrete/ml/quantization/quantized_module.py
fd0r: Ho yeah it was just for debugging purposes, I'll revert that, thanks for noticing

src/concrete/ml/quantization/quantized_module.py
fd0r: yeah makes sense, much better indeed!

src/concrete/ml/sklearn/base.py
fd0r: indeed!

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1259,54 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
 
 
+@pytest.mark.parametrize(
+    "model_object",
+    [
+        pytest.param(MultiOutputModel),
+    ],
+)
+def test_multi_output(model_object, default_configuration):
+    """Test fancy indexing torch."""
```
RomanBredehoft, line 1269: I don't think this docstring is right 😅

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1259,54 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+    model = model_object()
+    x = numpy.random.randint(0, 2, size=(100, 3, 10)).astype(numpy.float64)
+    y = numpy.random.randint(0, 2, size=(100, 3, 10)).astype(numpy.float64)
+    torch_result = model(x[[0]], y[[0]])
+
+    # Compare with simulation
+    quantized_module = compile_torch_model(
+        model, (x, y), n_bits=4, configuration=default_configuration
+    )
+    qm_result = quantized_module.forward(x[[0]], y[[0]])
```
RomanBredehoft, line 1279: would be great if the following lines could have a bit more spaces and why not comments to explain what we are checking, it's a bit hard to follow what is going on here !

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1259,54 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+    assert isinstance(simulation_result, tuple) and len(simulation_result) == 2
+    for qm_res, sim_res in zip(qm_result, simulation_result):
+        assert isinstance(qm_res, numpy.ndarray)
+        assert isinstance(sim_res, numpy.ndarray)
+        assert (qm_res == sim_res).all()
+    assert isinstance(torch_result, tuple) and len(torch_result) == 2
+    for qm_res, trch_res in zip(qm_result, torch_result):
+        assert isinstance(qm_res, numpy.ndarray)
+        assert isinstance(trch_res, numpy.ndarray)
+        # very low tolerance because we use low-bit-width
```
RomanBredehoft, line 1291: better to have a comment starting with a capitalized letter !

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1259,54 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+        (x, y),
+        n_bits=24,
+    )
+    qm_result = quantized_module.forward(x[[0]], y[[0]])
+    assert isinstance(qm_result, tuple) and len(qm_result) == 2
+    assert isinstance(torch_result, tuple) and len(torch_result) == 2
+    for qm_res, trch_res in zip(qm_result, torch_result):
+        assert isinstance(qm_res, numpy.ndarray)
+        assert isinstance(trch_res, numpy.ndarray)
+        # very low tolerance because we use low-bit-width
```
RomanBredehoft, line 1306: better to have a comment starting with a capitalized letter ! also, better to have a space above a comment

pyproject.toml
```diff
@@ -150,3 +150,6 @@ line-length = 100
 # The tool looks to report 'imported but unused (F401)' while actually the function is used by
 # package which import the __init__.py
 "**/__init__.py" = ["F401"]
+
+[tool.setuptools_scm]
```
RomanBredehoft, line 154: not sure what this is !

Makefile
RomanBredehoft: ok interesting then 1

pyproject.toml
RomanBredehoft: maybe worth a small comment then !

andrei-stoian-zama (changes requested):

pyproject.toml
```diff
@@ -34,7 +34,7 @@ readme = "README.md"
 # Investigate if it is better to fix specific versions or use lower and upper bounds  
 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/2665
 python = ">=3.8.1,<3.11"
-concrete-python = "2.5.0-rc1 "
+#concrete-python = "2023.11.5"
```
RomanBredehoft, line 37: is it right to comment this ? why not keeping the`2.5.0-rc1 ` version to make poetry consider more realistic constraint in the dependencies to install ? this is an open question, not sure of this answer

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -471,8 +479,8 @@ def _fhe_forward(self, *q_x: numpy.ndarray, simulate: bool = True) -> numpy.ndar
             "The quantized module is not compiled. Please run compile(...) first before "
             "executing it in FHE.",
         )
-
-        results_cnp_circuit_list = []
+        results_cnp_circuit_list: List[List[numpy.ndarray]] = []
```
RomanBredehoft, line 482: redundant lines !

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -532,22 +545,30 @@ def quantize_input(self, *x: numpy.ndarray) -> Union[numpy.ndarray, Tuple[numpy.
...
+            q_y_preds (numpy.ndarray): Quantized outputs values.
 
         Returns:
             numpy.ndarray: De-quantized output values of the last layer.
         """
-        y_preds = tuple(
-            output_quantizer.dequant(q_y_preds) for output_quantizer in self.output_quantizers
+        # Make sure that we have as many predictions as quantizers
+        assert len(q_y_preds) == len(self.output_quantizers), (
+            f"{len(q_y_preds)=} != {len(self.output_quantizers)=} "
```
RomanBredehoft, line 561: did not know about this `f"{len(q_y_preds)=}"` trick , nice !

src/concrete/ml/quantization/quantizers.py
```diff
@@ -771,7 +772,7 @@ def quant(self, values: numpy.ndarray) -> numpy.ndarray:
 
         return qvalues.astype(numpy.int64)
 
-    def dequant(self, qvalues: numpy.ndarray) -> Union[Any, numpy.ndarray]:
+    def dequant(self, qvalues: numpy.ndarray) -> numpy.ndarray:
```
RomanBredehoft, line 775: `-> Union[numpy.ndarray, Tracer]` maybe no ? surprised mypy did not complain

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1259,68 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+        assert isinstance(sim_res, numpy.ndarray)
+        assert (qm_res == sim_res).all()
+
+    # Assert that we aren't too far away from torch with low bit width
+    for qm_res, trch_res in zip(qm_result, torch_result):
+        assert isinstance(qm_res, numpy.ndarray)
+        assert isinstance(trch_res, numpy.ndarray)
+
+        # Very low tolerance because we use low-bit-width
+        assert numpy.isclose(qm_res, trch_res, atol=1e-1).all()
```
RomanBredehoft, line 1302: sorry did not see this one before, it's better if you use our utility function `array_allclose_and_same_shape` , we had issues in the past because isclose does not actually check the shape !

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1259,68 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+    # Assert that we the correct number of outputs again
+    assert isinstance(qm_result, tuple) and len(qm_result) == 2
+
+    # Assert that we have the same results as torch with high bit width quantization
+    for qm_res, trch_res in zip(qm_result, torch_result):
+        assert isinstance(qm_res, numpy.ndarray)
+        assert isinstance(trch_res, numpy.ndarray)
+
+        # Very low tolerance because we use low-bit-width
+        assert numpy.isclose(qm_res, trch_res, atol=1e-10).all()
```
RomanBredehoft, line 1321: same !

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1259,68 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+    # Assert that we have the expected number of outputs
+    assert isinstance(qm_result, tuple) and len(qm_result) == 2
+    assert isinstance(simulation_result, tuple) and len(simulation_result) == 2
+    assert isinstance(torch_result, tuple) and len(torch_result) == 2
+
+    # Assert that we are exact between simulation and clear quantized
+    for qm_res, sim_res in zip(qm_result, simulation_result):
+        assert isinstance(qm_res, numpy.ndarray)
+        assert isinstance(sim_res, numpy.ndarray)
+        assert (qm_res == sim_res).all()
```
RomanBredehoft, line 1294: same that below, with no tol I guess

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: so actually this makes me think that it's better to return floatings point for debug_values, or then we move this debug code to `quantized_forward` !

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -337,26 +337,27 @@ def forward(
...
             ] = {}
             for (_, layer) in self.quant_layers_dict.values():
                 layer.debug_value_tracker = debug_value_tracker
-            result = self.quantized_forward(*q_x, fhe="disable")
+            q_y_pred = self.quantized_forward(*q_x, fhe="disable")
             for (_, layer) in self.quant_layers_dict.values():
                 layer.debug_value_tracker = None
-            return result, debug_value_tracker
+            # De-quantize the output predicted values
+            y_pred = self.dequantize_output(*to_tuple(q_y_pred))
```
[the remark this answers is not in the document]
fd0r: I don't get why we didn't dequantize here.

`.forward` should return dequantized values, it just wasn't consistent in debug mode.

See https://github.com/zama-ai/concrete-ml/pull/396/files#diff-84c5dca0cde5235ebd72ca3a3d3ee9e9ca1c4d17e9e28ed7b92885705e5ae346L353

src/concrete/ml/quantization/quantized_module.py
fd0r: ha yeah indeed

src/concrete/ml/quantization/quantized_module.py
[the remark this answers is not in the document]
RomanBredehoft: I agree, either we return float values here or we move the debug code to the `quantized_forward` !

pyproject.toml
fd0r: https://github.com/pypa/setuptools_scm/issues/953

What would you add? A link to the issue or just a message saying that this is to avoid warnings?

src/concrete/ml/quantization/quantizers.py
fd0r: Hum actually no, we never have `Tracer`s in the `dequant` function.

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -461,7 +469,7 @@ def _fhe_forward(self, *q_x: numpy.ndarray, simulate: bool = True) -> numpy.ndar
                 Default to True.
 
         Returns:
-            (numpy.ndarray): Predictions of the quantized model, with integer values.
+            (Tuple[numpy.ndarray, ...]): Predictions of the quantized model, with integer values.
```
RomanBredehoft, line 472: technically it should be `Union[numpy.ndarray, Tuple[numpy.ndarray, ...]]`

tests/torch/test_compile_torch.py
fd0r: Nice utility function indeed.
Changing that!

src/concrete/ml/quantization/quantizers.py
RomanBredehoft: but at the end you have : 
```
      assert isinstance(values, (numpy.ndarray, Tracer))
        return values
```
🤔

pyproject.toml
RomanBredehoft: no yeah I just meant like a `# Tool that enables avoiding warnings` or something like, not very fancy, just to make things easier to understand !

src/concrete/ml/quantization/quantizers.py
fd0r: Ho yeah indeed weird, I'll remove `Tracer` from the assert.

fd0r: Ha no okay we are in a quantizer my bad, I thought it was in `quantized_module.py` yes I'll fix the type hint.

tests/torch/test_compile_torch.py
```diff
@@ -1258,6 +1263,68 @@ def test_fancy_indexing_torch(model_object, default_configuration):
...
+        assert isinstance(sim_res, numpy.ndarray)
+        assert array_allclose_and_same_shape(qm_res, sim_res, atol=1e-30)
+
+    # Assert that we aren't too far away from torch with low bit width
+    for qm_res, trch_res in zip(qm_result, torch_result):
+        assert isinstance(qm_res, numpy.ndarray)
+        assert isinstance(trch_res, numpy.ndarray)
+
+        # Very low tolerance because we use low-bit-width
+        assert array_allclose_and_same_shape(qm_res, trch_res, atol=1e-1)
```
RomanBredehoft, line 1306: I think you mean `1e-10` right ?

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -389,18 +391,19 @@ def quantized_forward(
...
+        self, *q_x: numpy.ndarray
+    ) -> Union[numpy.ndarray, Tuple[numpy.ndarray, ...]]:
         """Forward function for the FHE circuit executed in the clear.
 
         Args:
             *q_x (numpy.ndarray): Input integer values to consider.
 
         Returns:
-            (numpy.ndarray): Predictions of the quantized model, with integer values.
+            (Tuple[numpy.ndarray, ...]): Predictions of the quantized model, with integer values.
```
RomanBredehoft, line 406: sorry same again here `Union[numpy.ndarray, Tuple[numpy.ndarray, ...]]` !

pyproject.toml
fd0r: Not sure how well pypi handles nightly but maybe we could publish the nightlies that we use to pypi if there is no IP blocker.

src/concrete/ml/sklearn/base.py
```diff
@@ -1116,9 +1115,11 @@ def quantize_input(self, X: numpy.ndarray) -> numpy.ndarray:
         assert isinstance(q_X, numpy.ndarray)
         return q_X
 
-    def dequantize_output(self, q_y_preds: numpy.ndarray) -> numpy.ndarray:
+    def dequantize_output(self, *q_y_preds: numpy.ndarray) -> numpy.ndarray:
```
RomanBredehoft, line 1118: just realized here, we have a signature change here : `dequantize_output` is a common method between all built-in models, this means it 'inherits' from the BaseEstimator method, which has the following signature / docstring : 
```
    @abstractmethod
    def dequantize_output(self, q_y_preds: numpy.ndarray) -> numpy.ndarray:
        """De-quantize the output.

        This step ensures that the fit method has been called.

        Args:
            q_y_preds (numpy.ndarray): The quantized output values to de-quantize.

        Returns:
            numpy.ndarray: The de-quantized output values.
        """
```

what should we do then here ? I feel like this signature change is necessary (I guess QNNs could be multi-output right ?) but `dequantize_output` for linear / tree models should stay like it is right now 

I'm actually surprised mypy did not complain here, I thought it would point out that the method's signature has changed .. ? 

so yes, not sure what could be done here, it probably looks fine to just keep it liks this but maybe add a new docstring to better indicate that the signature is a bit different ?

pyproject.toml
RomanBredehoft: I think that's the main issue with nightly, I believe we don't want them to be public, only rc should

tests/torch/test_compile_torch.py
RomanBredehoft: ah ok I was just surprised because you put the same comment (`# Very low tolerance because we use low-bit-width`) here and below (for  `atol=1e-10`, last check)

RomanBredehoft (approved):
looks good ! I only have a few open questions remaining, but would be fine for me if we merge it like this , again, thanks a lot for this integration !

src/concrete/ml/quantization/quantized_module.py
fd0r: > debug_values are integers yes. I see your point about changing forward to return floats, but the forward was returning integers before.. maybe for some good reason, like client/server ?

Pretty sure we don't use the debug mode. Also this change didn't break any test so idk.

pyproject.toml
fd0r: Maybe we should them for a rc once we see that the nightly is working for us?

src/concrete/ml/quantization/quantized_module.py
fd0r: indeed

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -532,22 +547,30 @@ def quantize_input(self, *x: numpy.ndarray) -> Union[numpy.ndarray, Tuple[numpy.
...
+        self, *q_y_preds: numpy.ndarray
+    ) -> Union[numpy.ndarray, Tuple[numpy.ndarray, ...],]:
         """Take the last layer q_out and use its de-quant function.
 
         Args:
-            q_y_preds (numpy.ndarray): Quantized output values of the last layer.
+            q_y_preds (numpy.ndarray): Quantized outputs values.
 
         Returns:
             numpy.ndarray: De-quantized output values of the last layer.
```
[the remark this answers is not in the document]
fd0r: indeed

tests/torch/test_compile_torch.py
fd0r: Actually it's a typo in the second comment

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: actually, we do have a complete test for debugging interediary values : `test_brevitas_intermediary_values` .. looking at it quickly, I'm surprised it didn't break anything, it seems to be expecting some quantized values at some point, so worth to take a quick look at it I would say 😶

RomanBredehoft (dismissed):
Thanks, all is good for me !

RomanBredehoft (dismissed):

andrei-stoian-zama (dismissed):

tests/torch/test_compile_torch.py
jfrery: This has been resolved but has this been done?

tests/quantization/test_compilation.py
```diff
@@ -26,6 +26,7 @@
 INPUT_OUTPUT_FEATURE = [1, 2, 3]
 
 
+@pytest.mark.flaky
```
RomanBredehoft, line 29: please create an issue and add a FXIME here if this PR has detected a new flaky issue

tests/sklearn/test_sklearn_models.py
```diff
@@ -1575,6 +1575,7 @@ def test_fitted_compiled_error_raises(
     check_fitted_compiled_error_raises(model_class, n_bits, x, y)
 
 
+@pytest.mark.flaky
```
RomanBredehoft, line 1578: same, please create an issue and add a FXIME here if this PR has detected a new flaky issue

src/concrete/ml/quantization/quantized_module.py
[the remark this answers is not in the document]
andrei-stoian-zama: you can keep the debug values here as well, the computation is still done on integers

andrei-stoian-zama (approved):

RomanBredehoft (changes requested):
sorry but if we add new flaky marks to pytest, better to create an issue + add a fixme along these marks !

fd0r:
Indeed my bad, I'll create the issue and add the FIXME.

tests/quantization/test_compilation.py
RomanBredehoft: are you sure this is the right issue ? https://github.com/zama-ai/concrete-ml-internal/issues/4169 talks about simulation but this test is mostly about compilation (which made a worker crash in the last CI)

src/concrete/ml/quantization/quantized_module.py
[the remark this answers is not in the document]
fd0r: Not sure what you mean. Do you want me to return integers values?

tests/quantization/test_compilation.py
fd0r: Yes this is the same issue (see https://github.com/zama-ai/concrete-ml/actions/runs/7129808481/job/19414823145#step:27:30139 linked in the issue)

jfrery (approved):
Thanks!

PR #398, opened 2023-11-28
feat: allow QuantizedAdd produces_output_graph

jfrery:
ref https://github.com/zama-ai/concrete-ml-internal/issues/4073
ref https://github.com/zama-ai/concrete-ml-internal/issues/4106
ref https://github.com/zama-ai/concrete-ml-internal/issues/4144

### DIFF
--- a/src/concrete/ml/quantization/quantized_ops.py
+++ b/src/concrete/ml/quantization/quantized_ops.py
@@ -26,7 +26,12 @@ from .base_quantized_op import (
     QuantizedOp,
     QuantizedOpUnivariateOfEncrypted,
 )
-from .quantizers import QuantizationOptions, QuantizedArray, UniformQuantizationParameters
+from .quantizers import (
+    QuantizationOptions,
+    QuantizedArray,
+    UniformQuantizationParameters,
+    UniformQuantizer,
+)
 
 
 def _check_op_input_zero_point(zero_point: Any, op_name: Optional[str]):
@@ -492,7 +497,7 @@ class QuantizedMatMul(QuantizedGemm):
     _impl_for_op_named: str = "MatMul"
 
 
-class QuantizedAdd(QuantizedOp):
+class QuantizedAdd(QuantizedMixingOp):
     """Quantized Addition operator.
 
     Can add either two variables (both encrypted) or a variable and a constant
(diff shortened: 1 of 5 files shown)


src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -554,22 +559,32 @@ def q_impl(
...
+        common_zero_point = 0
+        offset = 0
+
+        output_quant_params = UniformQuantizationParameters(
+            scale=common_scale,
+            zero_point=common_zero_point,
+            offset=offset,
+        )
+
+        quantizer = UniformQuantizer(params=output_quant_params, no_clipping=True)
```
andrei-stoian-zama, line 583: very nice adding `no_clipping`  and re-using the quantizer

jfrery: Got inspired by @RomanBredehoft code on linear models! Yes I think it's best we reuse all the tools we have created for quantizer so that we keep updating them to our needs.

kcelia (dismissed):
Thanks

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -554,22 +559,32 @@ def q_impl(
...
+        # Otherwise we use the output op quantization scale
+        else:
+            common_scale = self.output_quant_params.scale
 
-        rescale_q1 = numpy.rint(
-            q_input_1.quantizer.scale
-            / self.output_quant_params.scale
-            * (q_input_1.qvalues + (-q_input_1.quantizer.zero_point))
-        ).astype(numpy.int64)
+        common_zero_point = 0
```
RomanBredehoft, line 574: is this correct ? I can see that we were using `q_input_0.quantizer.zero_point` and `q_input_1.quantizer.zero_point` above, but they were basically useless ?

RomanBredehoft (approved):
great thanks a lot !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: ah yes of course got it, thanks !

RomanBredehoft (commented):
Thanks !

RomanBredehoft (approved):
thanks !

PR #409, opened 2023-12-14
feat: support Expand Equal ONNX op

jfrery:
ref https://github.com/zama-ai/concrete-ml-internal/issues/4158

I had to implement Equal as well because the `torch.norm(...)` seems to create an Equal node. It's only there to compute constants so it does not appear in the final FHE graph.

### DIFF
--- a/src/concrete/ml/onnx/ops_impl.py
+++ b/src/concrete/ml/onnx/ops_impl.py
@@ -2065,21 +2065,21 @@ def numpy_gather(
 
 
 @onnx_func_raw_args("shape")
-def numpy_expand(x: numpy.ndarray, shape=None) -> Tuple[numpy.ndarray]:
+def numpy_expand(x: numpy.ndarray, shape: Optional[Tuple[int]] = None) -> Tuple[numpy.ndarray]:
     """Apply the expand operator in numpy according to ONNX spec.
 
     See https://github.com/onnx/onnx/blob/main/docs/Operators.md#expand
 
     Args:
         x (numpy.ndarray): Input tensor.
-        shape: Tuple of the new shape.
+        shape (Optional[Tuple[int]]): Tuple of the new shape.
 
     Returns:
         Tuple[numpy.ndarray]: Output tensor.
     """
     target_shape = numpy.array(shape, dtype=int)
-    padding = len(target_shape) - len(x.shape)
+    shape_difference = len(target_shape) - len(x.shape)
 
-    assert_true(padding >= 0, "Target shape cannot have fewer dimensions than input shape")
+    assert_true(shape_difference >= 0, "Target shape cannot have fewer dimensions than input shape")
 
     return (numpy.broadcast_to(x, target_shape),)
--- a/src/concrete/ml/onnx/ops_impl.py
+++ b/src/concrete/ml/onnx/ops_impl.py
@@ -941,6 +941,25 @@ def rounded_numpy_equal_for_trees(
     return (numpy.equal(x, y),)
 
 
+def numpy_equal_float(
+    x: numpy.ndarray,
+    y: numpy.ndarray,
+) -> Tuple[numpy.ndarray]:
+    """Compute equal in numpy according to ONNX spec and cast outputs to floats.
+
+    See https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Equal-13
+
+    Args:
+        x (numpy.ndarray): Input tensor
+        y (numpy.ndarray): Input tensor
+
+    Returns:
+        Tuple[numpy.ndarray]: Output tensor
+    """
+
+    return cast_to_float(numpy_equal(x, y))
+
+
 def numpy_not(
     x: numpy.ndarray,
 ) -> Tuple[numpy.ndarray]:
@@ -2043,3 +2062,24 @@ def numpy_gather(
     slices = tuple(slice(None) if i != axis else indices_list for i in range(x.ndim))
 
     return (x[slices],)
+
+
+@onnx_func_raw_args("shape")
+def numpy_expand(x: numpy.ndarray, shape=None) -> Tuple[numpy.ndarray]:
+    """Apply the expand operator in numpy according to ONNX spec.
+
+    See https://github.com/onnx/onnx/blob/main/docs/Operators.md#expand
+
+    Args:
+        x (numpy.ndarray): Input tensor.
+        shape: Tuple of the new shape.
+
+    Returns:
+        Tuple[numpy.ndarray]: Output tensor.
+    """
+    target_shape = numpy.array(shape, dtype=int)
+    padding = len(target_shape) - len(x.shape)
+
+    assert_true(padding >= 0, "Target shape cannot have fewer dimensions than input shape")
+
+    return (numpy.broadcast_to(x, target_shape),)
--- a/src/concrete/ml/quantization/quantized_ops.py
+++ b/src/concrete/ml/quantization/quantized_ops.py
@@ -2430,3 +2430,87 @@ class ONNXSlice(QuantizedOp):
             bool: False, this operation can not be fused
         """
         return False
+
+
+class QuantizedExpand(QuantizedOp):
+    """Expand operator for quantized tensors."""
+
+    _impl_for_op_named: str = "Expand"
+    quantize_inputs_with_model_outputs_precision = True
+
+    def q_impl(
+        self,
+        *q_inputs: ONNXOpInputOutputType,
+        **attrs,
+    ) -> ONNXOpInputOutputType:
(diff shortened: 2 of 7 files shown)


src/concrete/ml/onnx/ops_impl.py
```diff
@@ -2043,3 +2062,24 @@ def numpy_gather(
     slices = tuple(slice(None) if i != axis else indices_list for i in range(x.ndim))
 
     return (x[slices],)
+
+
+@onnx_func_raw_args("shape")
+def numpy_expand(x: numpy.ndarray, shape=None) -> Tuple[numpy.ndarray]:
```
RomanBredehoft, line 2068: shape parameter is missing a typing (I guess `Optional[Tuple[int]]`) !

src/concrete/ml/onnx/ops_impl.py
```diff
@@ -2043,3 +2062,24 @@ def numpy_gather(
...
+    """Apply the expand operator in numpy according to ONNX spec.
+
+    See https://github.com/onnx/onnx/blob/main/docs/Operators.md#expand
+
+    Args:
+        x (numpy.ndarray): Input tensor.
+        shape: Tuple of the new shape.
```
RomanBredehoft, line 2075: same, missing typing !

src/concrete/ml/onnx/ops_impl.py
```diff
@@ -2043,3 +2062,24 @@ def numpy_gather(
...
+
+    Returns:
+        Tuple[numpy.ndarray]: Output tensor.
+    """
+    target_shape = numpy.array(shape, dtype=int)
+    padding = len(target_shape) - len(x.shape)
```
RomanBredehoft, line 2081: `padding` is a confusing name I would say, maybe `shape_difference` ? not very important anyway

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -2430,3 +2430,87 @@ def can_fuse(self) -> bool:
...
+            op_instance_name,
+            int_input_names,
+            constant_inputs,
+            input_quant_opts,
+            **attrs,
+        )
+
+        # We do not support testing a == b where a,b are encrypted
+        # only comparing to a constant is supported
+        assert_true(constant_inputs is not None and len(constant_inputs) >= 1)
```
RomanBredehoft, line 2516: I'm wondering, shouldn't it be `len(constant_inputs) == 1` (or `len(constant_inputs) == 2`) ? why are we making this check "so loose" ?

tests/quantization/test_quantized_ops.py
```diff
@@ -1146,6 +1148,60 @@ def test_quantized_reshape(shape):
     )
 
 
+@pytest.mark.parametrize(
+    "original_shape, expand_shape",
+    [
```
RomanBredehoft, line 1153: great thorough tests ! But maybe there was a way to automate this ? How can we be sure that all interesting cases are covered ?

tests/quantization/test_quantized_ops.py
```diff
@@ -1193,81 +1249,66 @@ def test_quantized_prelu(n_bits, input_range, input_shape, slope, is_signed, che
     )
 
 
-@pytest.mark.parametrize(
-    "params",
-    [
+@pytest.fixture(scope="module")
+def random_test_data():
+    """Generate data for comparisons operators."""
+    numpy.random.seed(42)  # Fixing the seed for reproducibility
```
RomanBredehoft, line 1255: shouldn't we make it random like other tests : `random_state = numpy.random.randint(0, 2**15)` ? If we fix seeds like this, we'll never encounter some possible corner cases that would need to be fixed !

tests/torch/test_compile_torch.py
```diff
@@ -1144,7 +1147,12 @@ def test_shape_operations_net(
     check_float_array_equal,
 ):
     """Test a pattern of reshaping, concatenation, chunk extraction."""
-    net = ShapeOperationsNet(is_qat)
+    net = model(is_qat)
```
RomanBredehoft, line 1150: small detail but in most of our tests we use `model_class` for the parameter and `model` for the actual instance (instead of `net` here), would be great if you could update this test with that !

tests/torch/test_compile_torch.py
```diff
@@ -1144,7 +1147,12 @@ def test_shape_operations_net(
...
+    if is_qat and not net.is_qat_compatible:
+        pytest.skip(f"Model {model.__name__} is not compatible with is_qat=True")
+
+    # Shape transformation do not support >1 example in the inputset
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3871
     inputset = numpy.random.uniform(size=(1, n_channels, 2, 2))
 
     if is_qat:
```
RomanBredehoft, line 1158: just realized by reading what's following that in case of `is_qat=False`, we never check any r2 score or whatever. Basically, this PR is green only because the forward/compile didn't fail ! There is only this comment bu no FIXME attached to it : is this still expected ? : 
```
    # In QAT quantization options are consistent across all the layers
    # which allows for the elimination of TLUs
    # In PTQ there are TLUs in the graph because Shape/Concat/Transpose
    # must quantize inputs with some default quantization options

    # In QAT testing in FHE is fast since there are no TLUs
    # For PTQ we only test that the model can be compiled and that it can be executed
```

RomanBredehoft (dismissed):
some small details to be cleaned + I have a few questions, but overall looks good !

src/concrete/ml/onnx/ops_impl.py
jfrery: Indeed. But this is true for many of our numpy ops actually (i copied pasted from unsqueeze).

I will add a commit to fix the typings maybe.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Hmm not sure. Again a copy paste here, this assert comes back 5 times in this file.

tests/quantization/test_quantized_ops.py
jfrery: Good question. I don't think we can cover all edge cases. But when we see one we can add it to the list.

tests/quantization/test_quantized_ops.py
jfrery: Yes right I should have removed that. I used a fix seed for debugging. I should probably use pytest args instead. I will remove.

tests/torch/test_compile_torch.py
jfrery: Yes! Let's keep it consistent. I will update.

tests/torch/test_compile_torch.py
jfrery: Well with 3 bits we can't expect the values to be the same. However, we are only soing shape transformation so I guess that only the output shape should be checked. Not sure why we check the actual prediction at all.

src/concrete/ml/onnx/ops_impl.py
jfrery: I can change the name yes

src/concrete/ml/onnx/ops_impl.py
RomanBredehoft: yeah not surprised that some old stuff are not up to date with our typing / doc conventions !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: sure yes, it just seems a bit fishy, I think having a == should e fine (and better) !

tests/torch/test_compile_torch.py
RomanBredehoft: I see, well could you maybe improve that ? because I believe that when `is_qat` is False (which is your case with your added model), we do not check the shape !

src/concrete/ml/onnx/ops_impl.py
jfrery: Done

src/concrete/ml/onnx/ops_impl.py
jfrery: Done

src/concrete/ml/onnx/ops_impl.py
jfrery: Done

tests/quantization/test_quantized_ops.py
jfrery: Done

tests/torch/test_compile_torch.py
jfrery: Done

PR #419, opened 2023-12-27
chore: update concrete-python to 2.5.0-rc2

fd0r:


deps_licenses/licenses_linux_user.txt
```diff
@@ -38,38 +39,39 @@ onnxconverter-common, 1.13.0, MIT License
...
 tqdm, 4.66.1, MIT License; Mozilla Public License 2.0 (MPL 2.0)
-transformers, 4.32.0, Apache Software License
+transformers, 4.36.2, Apache Software License
 typing_extensions, 4.5.0, Python Software Foundation License
-urllib3, 1.26.16, MIT License
+urllib3, 1.26.18, MIT License
 uvicorn, 0.21.1, BSD License
 xgboost, 1.6.2, Apache Software License
-z3-solver, 4.12.3.0, MIT License
+z3-solver, 4.12.2.0, MIT License
```
fd0r, line 76: New package introduced by Concrete Python

pyproject.toml
```diff
@@ -115,6 +115,14 @@ sphinx = "6.1.3"
 tensorflow = {version = "2.12.0", markers="sys_platform != 'darwin'"}
 tensorflow-macos = {version = "2.12.0", markers="sys_platform == 'darwin'"}
 tensorflow-metal = {version = "0.8.0", markers="sys_platform == 'darwin' and platform_machine == 'arm'"}
+tensorflow-io-gcs-filesystem = [
```
fd0r, line 118: We need to force to 0.34.0, since it has a larger support of python versions and os/architectures.

pyproject.toml
```diff
@@ -115,6 +115,14 @@ sphinx = "6.1.3"
...
+  {version = "0.34.0"},
+]
+# tensorflow-io-gcs-filesystem = [
+#   {version = "0.35.0", markers="sys_platform == 'darwin' and platform_machine != 'arm' and python_version >= '3.9'"},
+#   {version = "0.35.0", markers="sys_platform != 'darwin' and python_version >= '3.9'"},
+#   {version = "0.35.0", markers="sys_platform != 'darwin' and python_version == '3.8' and platform_machine != 'arm'"}
+# ]
```
fd0r, line 125: We could also have something like that but for some reason it didn't seem to work.

poetry.lock
```diff
@@ -779,144 +794,183 @@ cron = ["capturer (>=2.4)"]
...
+    {file = "concrete_python-2.5.0rc2-cp39-cp39-manylinux_2_28_x86_64.whl", hash = "sha256:cae76016f85a5dc07bc6da44650107e9874326b195e1f033cc296c278bb80306"},
+]
+
+[package.dependencies]
+importlib-resources = ">=6.1"
+networkx = ">=2.6"
+numpy = ">=1.23"
+scipy = ">=1.10"
+torch = ">=1.13"
+z3-solver = "4.12.2.0"
```
fd0r, line 839: This line was modified manually, because poetry generates `4.12.2` as version number instead of `4.12.2.0` for an unknown reason.

This then breaks `poetry install`.
To keep in mind for further updates.
I'll open an issue to poetry.

bcm-at-zama (approved):
Approved

poetry.lock
umut-sahin: Not sure if this is sustainable. Every time there is an update, you'll need to manually change it :thinking:

fd0r: Yes I know, this is a workaround until poetry fixes that.

bcm-at-zama (approved):

PR #424, opened 2024-01-03
chore: change the way we use gitleaks

fd0r:
Until now we used gitleaks to check the whole history but it makes more sense to only check from the first commit to the current commit.

Otherwise it would break all PRs if someone pushed something that was detected as a leak.
This introduces some issues also when working with multiple remotes.

The commit changes the behavior to check from first commit to current commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -715,9 +715,11 @@ check_supported_ops:
 	git diff docs/deep-learning/onnx_support.md
 	git diff --quiet docs/deep-learning/onnx_support.md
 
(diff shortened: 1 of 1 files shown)


IceTDrinker (approved):
Looks good, this restricts gitleaks to the current branch/PR compared to the old behavior but you were aware of that so should be ok

RomanBredehoft (approved):
I think this is a great idea, thanks !

PR #425, opened 2024-01-04
feat: SGDClassifier training in FHE

fd0r:
* Add SGDClassifier
* Add SGDClassifier FHE training
  - FHE training still has some limitations, mainly: - Binary classification - Single target
* Add option to copy inputs in QuantizedReduceSum
  - For model training in FHE it appeared the need to copy the inputs of the reduce-sum operation to avoid bit-width propagation. This behavior is not always desirable as it adds a PBS that might not be needed for all models. So the solution was to add a flag to do so.
* Fix: fix macos grep usage in our Makefile
  - MacOS has a different grep API than GNU. This crashes one of our script. A simple if-else and change of option name fixes that.

closes https://github.com/zama-ai/concrete-ml-internal/issues/3579


src/concrete/ml/sklearn/linear_model.py
```diff
@@ -103,6 +118,959 @@ def load_dict(cls, metadata: Dict):
...
+        # Generate the weight values to consider for compilation
+        weights_compile_set = numpy.empty((compile_size, x_min.shape[0], n_targets))
+
+        # Generate the bias values to consider for compilation
+        bias_compile_set = numpy.empty((compile_size, 1, n_targets))
+
+        compile_set = (x_compile_set, y_compile_set, weights_compile_set, bias_compile_set)
+
+        # Bound values are hard-coded in order to make sure that the circuit never overflows
+        for index, (label, x_value, coef_value, bias_value) in enumerate(combinations):
```
RomanBredehoft, line 352: I feel like this loop just creates a new `compile_set` from `combinations`, unless I'm missing something, I believe `compile_set`'s above definition (l349) is useless no ?

fd0r: everything is empty at first and the loop populates it, I could define it afterwards but that's not really different to what's being done here

src/concrete/ml/torch/compile.py
```diff
@@ -117,7 +120,8 @@ def build_quantized_module(
     # only work over shape of (1, ., .). For example, some reshape have newshape hardcoded based
     # on the inputset we sent in the NumpyModule.
     quantized_module = post_training_quant.quantize_module(*inputset_as_numpy_tuple)
-
+    if reduce_sum_copy:
```
RomanBredehoft, line 123: I would still add a fixme for https://github.com/zama-ai/concrete-ml-internal/issues/4127 here though !

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: ah yes ok sorry I get it now, you go from a list of n tuples of 4 elements to a tuple of 4 arrays of shape (n, ...)

I feel like there might be a simpler way (and easier to read) of doing so but I don't see it for now so looks good to me

RomanBredehoft (approved):
thanks a lot for all of this !

PR #426, opened 2024-01-04
chore: removing these per-model threshold, they are no more needed

bcm-at-zama:
if pytest is green, let's clean that

closes https://github.com/zama-ai/concrete-ml-internal/issues/4088

### DIFF
--- a/tests/sklearn/test_sklearn_models.py
+++ b/tests/sklearn/test_sklearn_models.py
@@ -186,33 +186,9 @@ def check_correctness_with_sklearn(
         warnings.simplefilter("ignore", category=ConvergenceWarning)
         model, sklearn_model = model.fit_benchmark(x, y)
 
-    # For R2 score measures
-    acceptance_r2scores = {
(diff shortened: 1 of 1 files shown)


fd0r:
Let's just hope we don't create flakies with that

bcm-at-zama:
We can delay this merge, but really, this part of the test is ugly.

bcm-at-zama:
Remark that the only important lines were

```
        "XGBClassifier": 0.7,
        "RandomForestClassifier": 0.8,
```

here

RomanBredehoft:
yes I think it should be fine now ! Note that this test does not consider QNNs because they were flaky / did not make much sense

anyway, you have a conflict btw

fd0r (approved):

bcm-at-zama:
Rebasing

fd0r (approved):
If the CI is green that's fine with me!

RomanBredehoft (approved):
great thanks !

PR #430, opened 2024-01-08
chore: remove last MPS occurence

RomanBredehoft:
refs https://github.com/zama-ai/concrete-ml-internal/issues/3953

### DIFF
--- a/docs/advanced_examples/QuantizationAwareTraining.ipynb
+++ b/docs/advanced_examples/QuantizationAwareTraining.ipynb
@@ -400,11 +400,10 @@
     }
    ],
    "source": [
-    "# You can also train this on GPU for example by setting\n",
-    "# MPS option is supported by macOS with Apple Silicon or AMD GPUs\n",
-    "device = (\n",
-    "    \"cuda\" if torch.cuda.is_available() else \"mps\" if torch.backends.mps.is_available() else \"cpu\"\n",
-    ")\n",
+    "# Add MPS (for macOS with Apple Silicon or AMD GPUs) support when error is fixed. For now, we\n",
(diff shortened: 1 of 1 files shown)


docs/advanced_examples/QuantizationAwareTraining.ipynb
```diff
@@ -400,11 +400,10 @@
...
    "source": [
-    "# You can also train this on GPU for example by setting\n",
-    "# MPS option is supported by macOS with Apple Silicon or AMD GPUs\n",
-    "device = (\n",
-    "    \"cuda\" if torch.cuda.is_available() else \"mps\" if torch.backends.mps.is_available() else \"cpu\"\n",
-    ")\n",
+    "# Add MPS (for macOS with Apple Silicon or AMD GPUs) support when error is fixed. For now, we\n",
+    "# observe a decrease in torch's top1 accuracy when using MPS devices\n",
+    "# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3953\n",
+    "device = \"cuda\" if torch.cuda.is_available() else \"cpu\"\n",
```
fd0r, line 406: Should we have FIXME's in notebooks?

fd0r (approved):
Fine with me but a bit skeptical about having internal fixmes visible in notebooks.

docs/advanced_examples/QuantizationAwareTraining.ipynb
RomanBredehoft: well, we already have a few fixme in some notebooks, I agree it's not awesome but I think it's fine as long as it's only a few

fd0r: alright then

PR #434, opened 2024-01-09
docs: add documentation about scaling

fd0r:


docs/advanced_examples/utils/scaling_comparison_utils.py
```diff
@@ -0,0 +1,109 @@
+import matplotlib.pyplot as plt
```
RomanBredehoft, line 1: not sure what this file is about, it does not seem to be used anywhere !

docs/advanced_examples/LogisticRegression.ipynb
```diff
@@ -450,6 +451,174 @@
     "Decision boundaries from both models are almost identical. This shows executing a logistic regression\n",
     "model in FHE doesn't really impact the performances compared to its scikit-learn equivalent.  "
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "80d8ac5f-a78e-45c7-9251-5a37476d4768",
+   "metadata": {},
+   "source": [
+    "The plot above shows that the quantization introduced by Concrete ML didn't degrade the performance of the model.\n",
```
RomanBredehoft, line 460: this section is awesome ! I would just add a title to it so that we quickly get what the following is about

RomanBredehoft (changes requested):
awesome thanks ! I would just add a title in the notebook for clarity

docs/built-in-models/linear.md
```diff
@@ -28,6 +28,9 @@ The `n_bits` parameter controls the bit-width of the inputs and weights of the l
 
 The `n_bits` parameter can be set to `8` or more bits for models with up to `300` input dimensions. When the input has more dimensions, `n_bits` must be reduced to `6-7`. All performance metrics are preserved down to `n_bits=6`, compared to the non-quantized float models from scikit-learn.
 
+The same quantization parameters (i.e., scale and zero-point) are applied on all features, so it can be beneficial to uniformize the distribution of the different features.
```
bcm-at-zama, line 31: - do we mention using **Scaler directly in the doc? I think it would be interesting to have code block directly with one Scaler example
- I think it would be interesting that also, we have the word 'normalization', such that it's searchable in the doc

fd0r: The goal here is really to make the user take a look at the notebook such that it's clear what the impact of it can be and  they can directly copy-paste the snippet from there.

Could make sense to also have a snippet here but as you want

bcm-at-zama: I think we need to convince them there is a big problem here, else they'll never go to the nb

RomanBredehoft (approved):
thanks !

docs/built-in-models/linear.md
fd0r: I don't think a code snippet will do that tbh

It might not be an issue for datasets where the features have more or less the same range.

bcm-at-zama: it's not to convince them there is a problem with the code snippet: it's to tell them how to not have the problem with code snippet, and to have Scaler and normalization somewhere in the doc

fd0r:
Waiting for @andrei-stoian-zama 's approval to merge.

docs/built-in-models/linear.md
jfrery, line 31: Not sure about the 
```
The same quantization parameters (i.e., scale and zero-point) are applied on all features, so it can be beneficial to uniformize the distribution of the different features.
```

Maybe something like:
```suggestion
The same quantization parameters (i.e., scale and zero-point) are applied on all features, so it can be beneficial to make all feature distribution similar by using standard or min-max normalization.
```

jfrery (changes requested):
Small comment + I had not seen the end of the notebook. I find it very dense Maybe we put the code in a file and have the figure somewhere in the doc instead?

fd0r:
@jfrery so remove from the notebook and just include the figure with some comments in the markdown file?

andrei-stoian-zama (changes requested):
I think it needs a bit more work:

- Introduce the last section
- Remove the utility code or make it smaller
- The graphs need bigger text 
- it's not clear why there is variation in accuracy between no-transform and normalized

docs/built-in-models/linear.md
```diff
@@ -28,6 +28,9 @@ The `n_bits` parameter controls the bit-width of the inputs and weights of the l
 
 The `n_bits` parameter can be set to `8` or more bits for models with up to `300` input dimensions. When the input has more dimensions, `n_bits` must be reduced to `6-7`. All performance metrics are preserved down to `n_bits=6`, compared to the non-quantized float models from scikit-learn.
 
+The same quantization parameters (i.e., scale and zero-point) are applied on all features, so it can be beneficial to make all feature distribution similar by using standard or min-max normalization.
+For a more detailed comparison of the impact of such pre-processing please refer to [the logistic regression notebook](../advanced_examples/LogisticRegression.ipynb).
```
RomanBredehoft, line 32: could be worth to include the graphs from the notebook here for better understanding !

fd0r: They might take too much place in the readme, what do you think about it @andrei-stoian-zama ?

RomanBredehoft: yeah indeed, well as you guys wish

RomanBredehoft (approved):
thanks !

andrei-stoian-zama (approved):
thanks!

fd0r:
> I still find the end of the notebook a bit intense. Could we just keep the graph + explanation and put the pre processing code outside of the notebook? or add some comments?

Do you mean an explanation per processing done on the features?

jfrery (approved):

PR #435, opened 2024-01-09
chore: remove last mentions of Concrete-Numpy

RomanBredehoft:
There is still a method called "cpn_round" in our mixin quantized op but that would create a breaking change, so I added it in https://github.com/zama-ai/concrete-ml-internal/issues/3995 

Also, there are a few (3 I think) links to some concrete-numpy-internal issues left

closes https://github.com/zama-ai/concrete-ml-internal/issues/4074
closes https://github.com/zama-ai/concrete-ml-internal/issues/2250

### DIFF
--- a/src/concrete/ml/sklearn/base.py
+++ b/src/concrete/ml/sklearn/base.py
@@ -15,7 +15,6 @@ from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Set, TextIO, Type, Union
 
 import brevitas.nn as qnn
-import concrete.fhe as cnp
 import numpy
 import onnx
 import sklearn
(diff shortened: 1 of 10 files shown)


src/concrete/ml/sklearn/base.py
```diff
@@ -31,6 +30,9 @@
 from sklearn.linear_model import LinearRegression, LogisticRegression
 from sklearn.utils.validation import check_is_fitted
 
+# pylint: disable-next=ungrouped-imports
+from concrete import fhe as cp
```
fd0r, line 34: why not just use fhe.tag?

fd0r (approved):

src/concrete/ml/sklearn/base.py
RomanBredehoft: because of some pylint magic 🙂 

that's obviously what I've done at the beginning, which requires us to call `from concrete import fhe`. But guess what, since we have this `fhe` parameter in our `predict` methods, pylint was throwing a `W0621: Redefining name 'fhe' from outer scope (line 33) (redefined-outer-name)` 🙂 (https://github.com/zama-ai/concrete-ml/actions/runs/7463742016/job/20309199710)

and I didn't want to have `from concrete.fhe import tag, zeros, array` because calling `zeros(labels.shape)` or `array(arr)` in or code feels a bit ambiguous !

fd0r: Makes sense thanks a lot!

PR #441, opened 2024-01-11
chore: fix poetry version check in sync_env make target

RomanBredehoft:
To make sure everyone has the right poetry version ! cc @jfrery


fd0r:
what was wrong with the previous way of checking the version?

PR #444, opened 2024-01-11
Dont track apidoc

fd0r:


.github/workflows/continuous-integration.yaml
```diff
@@ -510,13 +510,15 @@ jobs:
               - tests/seeding/test_seeding.py
             docs:
               - docs/**
+              - '!docs/developer-guide/api/**.md'
```
RomanBredehoft, line 513: not sure about this one honestly , that would skip `make docs` in case only apidocs were updated (and I don't think that's a good idea)

RomanBredehoft (changes requested):
skipping codeblocks when apidocs is updated seems legit to me ! I would just leave it for make docs

RomanBredehoft (approved):
makes sense, thanks !

RomanBredehoft (approved):

PR #449, opened 2024-01-15
Feat/integrate fhe sum and quantized dict for input and leaves

kcelia:
In this PR: 

- we add the aggregation of outputs from the tree ensembles by setting the 'TREES_USE_FHE_SUM' environment variable, which is disabled by default.
- the opportunity to assign a different quantization to the leaves the quantification of the input and the leaves

Ref https://github.com/zama-ai/concrete-ml-internal/issues/4258
Closes https://github.com/zama-ai/concrete-ml-internal/issues/4167
Closes https://github.com/zama-ai/concrete-ml-internal/issues/451
Closes https://github.com/zama-ai/concrete-ml-internal/issues/4230
~Closes https://github.com/zama-ai/concrete-ml-internal/issues/4095~


kcelia:
The changes that have been made to the notebook might be the result of the rebase. 
I will correct them

src/concrete/ml/quantization/__init__.py
```diff
@@ -1,6 +1,11 @@
 """Modules for quantization."""
 from .base_quantized_op import QuantizedOp
-from .post_training import PostTrainingAffineQuantization, PostTrainingQATImporter, get_n_bits_dict
+from .post_training import (
+    PostTrainingAffineQuantization,
+    PostTrainingQATImporter,
+    get_n_bits_dict,
+    get_n_bits_dict_trees,
```
kcelia, line 7: I added a new get_n_bits for trees, to avoid changing the behavior of the existing method that uses op_weights

jfrery (changes requested):
Thanks! I am having a look at this when the PR is green.

Could you check https://github.com/zama-ai/concrete-ml-internal/issues/4095 as well please? This PR should close it if the output quantization is correct.

src/concrete/ml/sklearn/base.py
```diff
@@ -96,6 +101,9 @@
 # However, for internal testing purposes, we retain the capability to disable this feature
 os.environ["TREES_USE_ROUNDING"] = "1"
 
+# By default, the decision of the tree ensembles is made in clear
+os.environ["TREES_USE_FHE_SUM"] = "0"
```
andrei-stoian-zama, line 105: should't we consider what a user may have configured in their env ? e.g. 

`os.environ["TREES_USE_FHE_SUM"] = os.environ.get("TREES_USE_FHE_SUM", 0)`

andrei-stoian-zama (changes requested):
I'll check soon!

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,55 @@
...
+
+
+def get_n_bits_dict_trees(n_bits: Union[int, Dict[str, int]]) -> Dict[str, int]:
+    """Convert the n_bits parameter into a proper dictionary for tree based-models.
+
+    Args:
+        n_bits (int, Dict[str, int]): number of bits for quantization, can be a single value or
+            a dictionary with the following keys :
+            - "op_inputs" (mandatory)
+            - "op_leaves" (optional)
```
RomanBredehoft, line 36: if it's optional, better to indicate what default value it uses !

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,55 @@
...
+        n_bits_dict (Dict[str, int]): A dictionary properly representing the number of bits to use
+            for quantization.
+    """
+
+    assert_true(
+        isinstance(n_bits, int)
+        or (isinstance(n_bits, Dict) and set(n_bits.keys()).issubset({"op_inputs", "op_leaves"})),
+        "Invalid n_bits, either pass an integer or a dictionary containing integer values for "
+        "the following keys:\n"
+        "- `op_inputs` and `op_leaves` (mandatory)",
```
RomanBredehoft, line 50: this message does not match the docstring ! is `op_leaves` optional or mandatory ?

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,55 @@
...
+            "op_leaves" bits.
+
+    Returns:
+        n_bits_dict (Dict[str, int]): A dictionary properly representing the number of bits to use
+            for quantization.
+    """
+
+    assert_true(
+        isinstance(n_bits, int)
+        or (isinstance(n_bits, Dict) and set(n_bits.keys()).issubset({"op_inputs", "op_leaves"})),
```
RomanBredehoft, line 47: this assert should also check that the dic has the mandatory keys ! Here, you only check that it has at least one of them, you can check `get_n_bits_dict` for ex

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,55 @@
...
+            - "op_leaves" (optional)
+            When using a single integer for n_bits, its value is assigned to "op_inputs" and
+            "op_leaves" bits.
+
+    Returns:
+        n_bits_dict (Dict[str, int]): A dictionary properly representing the number of bits to use
+            for quantization.
+    """
+
+    assert_true(
```
RomanBredehoft, line 45: by the way this lets me think that this should not be an `assert` but rather an ValueError raise (same goes for `get_n_bits_dict_trees`). I don't think asserting a user input is a great practice, but I guess we could make this change in a following PR

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,55 @@
...
+    if isinstance(n_bits, int):
+        n_bits_dict = {
+            "op_inputs": n_bits,
+            "op_leaves": n_bits,
+        }
+
+    elif isinstance(n_bits, Dict):
+        n_bits_dict.update(n_bits)
+
+    assert_true(
```
RomanBredehoft, line 66: same here maybe this should be an exception raise instead of an assert

src/concrete/ml/sklearn/base.py
RomanBredehoft: agree, and same should go `s.environ["TREES_USE_ROUNDING"] = "1"` I guess

src/concrete/ml/sklearn/base.py
```diff
@@ -1279,13 +1287,19 @@ def __init_subclass__(cls):
...
         """Initialize the TreeBasedEstimatorMixin.
 
         Args:
-            n_bits (int): The number of bits used for quantization.
+            n_bits (int, Dict[str, int]): Number of bits to quantize the model. If an int is passed
+                for n_bits, the value will be used for quantizing inputs and leaves. If a dict is
+                passed, then it should contain "op_inputs" and "op_leaves" as keys with
+                corresponding number of quantization bits so that:
+                    - op_inputs : number of bits to quantize the input values
+                    - op_leaves: number of bits to quantize the leaves
```
RomanBredehoft, line 1299: same, this docstring should match the one you used for the `get_n_bits_dict_trees` (indicate what is mandatory or optional, with default value if optional) !

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,55 @@
...
+
+    # If a single integer is passed, we use a default value for the model's input and
+    # output bits
+    if isinstance(n_bits, int):
+        n_bits_dict = {
+            "op_inputs": n_bits,
+            "op_leaves": n_bits,
+        }
+
+    elif isinstance(n_bits, Dict):
```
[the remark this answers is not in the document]
RomanBredehoft: if you don't use them in tree-based models then no need !

src/concrete/ml/quantization/__init__.py
RomanBredehoft: now that I think of it, I definitively get the relevancy behind having an `op_leaves` instead of `op_weights`. However, we are therefore introducing a new name and making this `n_bits` parameter more complex to users (which are already not supposed to be familiar with it) , so maybe we should discuss about that. 

Personally, I think I wouldn't be against keeping the `op_weights` even for trees for that particular reason

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,55 @@
...
+    )
+
+    n_bits_dict: Dict = {}
+
+    # If a single integer is passed, we use a default value for the model's input and
+    # output bits
+    if isinstance(n_bits, int):
+        n_bits_dict = {
+            "op_inputs": n_bits,
+            "op_leaves": n_bits,
```
RomanBredehoft, line 60: now that I think of it, I definitively get the relevancy behind having an op_leaves instead of op_weights. However, we are therefore introducing a new name and making this n_bits parameter more complex to users (which are already not supposed to be familiar with it) , so maybe we should discuss about that.

Personally, I think I wouldn't be against keeping the op_weights even for trees for that particular reason

what do you think @zama-ai/machine-learning ?

src/concrete/ml/sklearn/rf.py
```diff
@@ -124,7 +124,7 @@ def load_dict(cls, metadata: Dict):
             obj.sklearn_model,
             numpy.zeros((len(obj.input_quantizers),))[None, ...],
             framework=obj.framework,
-            output_n_bits=obj.n_bits,
+            output_n_bits=obj.n_bits["op_leaves"] if isinstance(obj.n_bits, Dict) else obj.n_bits,
```
RomanBredehoft, line 127: `obj.n_bits` should only be a dict here, since you call `get_n_bits_dict_trees` in the init, so no need for the isinstance normally

same goes for all below changes

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -141,12 +146,19 @@ def add_transpose_after_last_node(onnx_model: onnx.ModelProto):
     # Get the output node
     output_node = onnx_model.graph.output[0]
 
-    # Create the node with perm attribute equal to (2, 1, 0)
+    # When using FHE sum for tree ensembles, create the node with perm attribute equal to (1, 0)
```
RomanBredehoft, line 149: could you maybe add a comment explaining why this change is needed (and what is this perm attribute) ?

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -277,18 +292,22 @@ def tree_values_preprocessing(
...
     for i, initializer in enumerate(onnx_model.graph.initializer):
+
         # All constants in our tree should be integers.
         # Tree thresholds can be rounded up or down (depending on the tree implementation)
         # while the final probabilities/regression values must be quantized.
         # We extract the value stored in each initializer node into the init_tensor.
         init_tensor = numpy_helper.to_array(initializer)
+        # print(initializer.name, init_tensor.shape)
         if "weight_3" in initializer.name:
+            # print(init_tensor)
```
RomanBredehoft, line 303: this print should be removed !

tests/sklearn/test_sklearn_models.py
```diff
@@ -1871,3 +1946,29 @@ def test_rounding_consistency_for_regular_models(
...
+    verbose=True,
+):
+    """Test that Concrete ML without and with rounding are 'equivalent'."""
+
+    if verbose:
+        print("Run check_fhe_sum_consistency")
+
+    x, y = get_dataset(model_class, parameters, n_bits, load_data, is_weekly_option)
+
+    check_fhe_sum_consistency(
```
RomanBredehoft, line 1968: better to keep the same name between check and tests : `check_fhe_sum_for_tree_based_models` !

tests/sklearn/test_sklearn_models.py
```diff
@@ -1871,3 +1946,29 @@ def test_rounding_consistency_for_regular_models(
...
+@pytest.mark.parametrize("n_bits", [2, 5, 10])
+def test_fhe_sum_for_tree_based_models(
+    model_class,
+    parameters,
+    n_bits,
+    load_data,
+    is_weekly_option,
+    verbose=True,
+):
+    """Test that Concrete ML without and with rounding are 'equivalent'."""
```
RomanBredehoft, line 1961: wrong mention of rounding here !

tests/sklearn/test_sklearn_models.py
```diff
@@ -1834,7 +1909,7 @@ def test_linear_models_have_no_tlu(
 # Additional tests for this purpose should be added in future updates
 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4179
 @pytest.mark.parametrize("model_class, parameters", get_sklearn_tree_models_and_datasets())
-@pytest.mark.parametrize("n_bits", [2, 5, 11])
+@pytest.mark.parametrize("n_bits", [2, 5, 10])
```
RomanBredehoft, line 1912: why the change ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1658,7 +1729,11 @@ def test_p_error_simulation(
     The test checks that models compiled with a large p_error value predicts very different results
     with simulation or in FHE compared to the expected clear quantized ones.
     """
-    n_bits = get_n_bits_non_correctness(model_class)
+
+    if os.getenv("TREES_USE_FHE_SUM") == "1":
+        n_bits = 4
```
RomanBredehoft, line 1734: if only works for 4 bits ? also maybe better to include this in directly in `get_n_bits_non_correctness` and be sure to check that this is only applied for tree-based models (using `model_class`) !

tests/sklearn/test_dump_onnx.py
```diff
@@ -222,8 +221,15 @@ def test_dump(
...
   %/_operators.0/MatMul_1_output_0 = MatMul(%_operators.0.weight_3, %/_operators.0/Reshape_2_output_0)
   %/_operators.0/Reshape_3_output_0 = Reshape[allowzero = 0](%/_operators.0/MatMul_1_output_0, %/_operators.0/Constant_3_output_0)
-  %transposed_output = Transpose[perm = [2, 1, 0]](%/_operators.0/Reshape_3_output_0)
-  return %transposed_output
+  """
+        + (
+            """%/_operators.0/ReduceSum_output_0 = ReduceSum[keepdims = 0](%/_operators.0/Reshape_3_output_0, %onnx::ReduceSum_22)
+  %transposed_output = Transpose[perm = [1, 0]](%/_operators.0/ReduceSum_output_0)
+  """
+            if os.getenv("TREES_USE_FHE_SUM") == "1"
```
RomanBredehoft, line 229: the problem is that the test will only run for one value, not both right ?

src/concrete/ml/sklearn/rf.py
kcelia: `get_n_bits_dict_trees` is not in the init but in the fit. 
I had "test_pipline" and "grid_search" tests failing because of that.

tests/sklearn/test_dump_onnx.py
kcelia: yes only one

src/concrete/ml/quantization/post_training.py
kcelia: updated

tests/sklearn/test_dump_onnx.py
RomanBredehoft: well it's not great for coverage no ?

tests/sklearn/test_sklearn_models.py
RomanBredehoft: ok got it !

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: yes, do the same for that one too!

andrei-stoian-zama:
Please add the issues you close, for example https://github.com/zama-ai/concrete-ml-internal/issues/4167 ?

src/concrete/ml/quantization/post_training.py
kcelia: In our first discussion with @andrei-stoian-zama on this subject, we thought about quantifying op_leaves only when TREES_USE_FHE_SUM is enabled. 

Today, it is disabled by default and we intend to enabled it when all the necessary security tools are ready. 

Then, after consideration, I think it's a great feature to add a dict for n_bits for tree. 
Even by reducing the precision of the leaves, the accuracy is preserved for the classifiers, a little less for the regressors, and on top of that, having it allows us to go further with the trees (more trees, more features, more examples than before). 

so if we validate the integration of the dict for trees, we'll have to agree on the name op_leaves, op_output, or op_weights. 
and we'll have to explain it the doc:

src/concrete/ml/quantization/post_training.py
kcelia: I agree.

I'll do it in this PR for get_n_bits_trees, and do it for the other function in another PR

src/concrete/ml/sklearn/rf.py
RomanBredehoft: ah yes I see, pretty annoying then but oh well

src/concrete/ml/quantization/post_training.py
RomanBredehoft: yes I was only talking about adding this `op_leaves` naming through `get_n_bits_dict_trees` instead of keeping things simple as they already are with `get_n_bits_dict` (like linear models, which don't use `model_*` attributes)

src/concrete/ml/quantization/post_training.py
RomanBredehoft: yes great ! here's the issue for it : https://github.com/zama-ai/concrete-ml-internal/issues/4232

src/concrete/ml/quantization/__init__.py
RomanBredehoft: by the way I would start to make this fund private (will do the same for `get_n_bits_dict` later : https://github.com/zama-ai/concrete-ml-internal/issues/3995)

src/concrete/ml/quantization/post_training.py
RomanBredehoft: you mean you'll do it in a following PR like you said above right ?

src/concrete/ml/sklearn/base.py
```diff
@@ -1410,10 +1433,13 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
         # Sum all tree outputs
         # Remove the sum once we handle multi-precision circuits
         # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/451
-        y_preds = numpy.sum(y_preds, axis=-1)
+        if os.getenv("TREES_USE_FHE_SUM") == "0":
```
RomanBredehoft, line 1436: the main problem I see with this sort of behavior is in a client-server setting : if the client's env variables are not matching the ones from the server, we'll have some issues here .. 

but I'm not able to find any proper solutions yet. I thought of a `self._run_sum_in_fhe = os.getenv("TREES_USE_FHE_SUM")` in the `__init__` for ex (which would fix the problem I mentionned) but I'm really not sure it's a good practice

RomanBredehoft: any thoughts on it @zama-ai/machine-learning

src/concrete/ml/sklearn/rf.py
RomanBredehoft, line 127: ideally, https://github.com/zama-ai/concrete-ml-internal/issues/4234 would avoid this kind of things, but we'll see that later I think

src/concrete/ml/quantization/post_training.py
kcelia: done

src/concrete/ml/quantization/post_training.py
kcelia: yes

src/concrete/ml/sklearn/rf.py
kcelia: You will probably encounter, this error if you define a dict in the init:

`RuntimeError: Cannot clone object DecisionTreeClassifier(n_bits=3), as the constructor either does not set or modifies parameter n_bits`

src/concrete/ml/sklearn/tree_to_numpy.py
kcelia: done

tests/sklearn/test_sklearn_models.py
```diff
@@ -1200,6 +1201,76 @@ def check_rounding_consistency(
...
+    x,
+    y,
+    n_bits,
+    is_weekly_option,
+):
+    """Test that Concrete ML without and with FHE sum are 'equivalent'."""
+
+    # Run the test with more samples during weekly CIs
+    if is_weekly_option:
+        fhe_test = get_random_samples(x, n_sample=5)
```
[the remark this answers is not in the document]
kcelia: Except `test_p_error_simulation` test all other fhe runs are executed in the weekly CI.

tests/sklearn/test_sklearn_models.py
kcelia: I reduced the precision because with the sum in fhe the bitwidth is larger and I have an OOM on my machine.

kcelia: done

tests/sklearn/test_dump_onnx.py
kcelia: It"s not a problem for the coverage, because it's the src that needs to be covered.

on the other hand, it would be really good to relaunch the weekly ci with TREES_USE_FHE_SUM enabled.

kcelia:
@zama-ai/machine-learning, 2 open questions are unresolved:

1- https://github.com/zama-ai/concrete-ml/pull/449/files#r1455654854
2 - Would it be better to relaunch the weekly CI with TREE_USE_SUM_FHE enabled ?

tests/sklearn/test_sklearn_models.py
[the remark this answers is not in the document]
RomanBredehoft: shouldn't we at least run a single FHE inference per regular CI ? iirc the only exceptions we've made for that for now are if fhe tests are way too long !

tests/sklearn/test_dump_onnx.py
RomanBredehoft: yeah I meant coverage in a general sense indeed, in this current state we are not testing with `TREES_USE_FHE_SUM` activated, which is a problem I think (and since we're not executing anything , only compiling, it's fine if we do that in regular CIs I think)

src/concrete/ml/sklearn/rf.py
RomanBredehoft: I'm not sure, iirc we implemented `get_params` and `get_sklearn_params` to fix this specific error 🤔  but we'll see

RomanBredehoft:
> 2 - Would it be better to relaunch the weekly CI with TREE_USE_SUM_FHE enabled ?

not sure why ? we are testing the feature with `TREE_USE_SUM_FHE=1` in your new test right ? and for the dump onnx imo we should tests both in regular CIs

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,13 +1290,23 @@ def __init_subclass__(cls):
...
+                    - op_inputs (mandatory): number of bits to quantize the input values
+                    - op_leaves (optional): number of bits to quantize the leaves
+                Default to 6.
         """
-        self.n_bits: int = n_bits
+
+        # Check if 'n_bits' is a valid value
+        _inspect_tree_n_bits(n_bits)
+
+        self.n_bits: Union[int, Dict[str, int]] = n_bits
```
RomanBredehoft, line 1309: why not using `_get_n_bits_dict_trees` directly here (and then remove the use of `_inspect_tree_n_bits(n_bits)` ?

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,97 @@
...
+
+    _inspect_tree_n_bits(n_bits)
+
+    # If a single integer is passed, we use a default value for the model's input and leaves
+    if isinstance(n_bits, int):
+        return {"op_inputs": n_bits, "op_leaves": n_bits}
+
+    # Default `op_leaves` to `op_inputs` if not specified
+    if "op_leaves" not in n_bits:
+        n_bits["op_leaves"] = n_bits["op_inputs"]  # pragma: no cover
```
RomanBredehoft, line 113: why this `# pragma: no cover` here ? we should test this ! we actually have a `test_n_bits.py` in our quantization tests so maybe that's the place where we could add something

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,97 @@
...
+        "`op_leaves` bits.\n"
+    )
+
+    error_message = ""
+
+    if isinstance(n_bits, int):
+        if n_bits <= 0:
+            error_message = "n_bits must be a non-null, positive integer"
+    elif isinstance(n_bits, dict):
+        if "op_inputs" not in n_bits.keys() or set(n_bits.keys()) - {"op_leaves", "op_inputs"}:
```
RomanBredehoft, line 68: small detail but I would maybe separate this into two branches, with `if "op_inputs" not in n_bits.keys()` having an error message like "key 'op_inputs' is mandatory" , which is more precise

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,97 @@
...
+            error_message = "All values in `n_bits` dictionary must be non-null, positive integers"
+
+        elif n_bits.get("op_leaves", 0) > n_bits.get("op_inputs", 0):
+            error_message = "`op_leaves` must be less than or equal to `op_inputs`"
+    else:
+        error_message = "n_bits must be either an integer or a dictionary"
+
+    if len(error_message) > 0:
+        raise ValueError(
+            f"{error_message}. Got `{type(n_bits)}` and `{n_bits}` value.\n{detailed_message}"
```
RomanBredehoft, line 83: small detail but I would rephrase `Got {type(n_bits)} and {n_bits} value.` to `Got {n_bits=} ({type(n_bits)=} ).`

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,97 @@
...
+            - "op_inputs" (mandatory): number of bits to quantize the input values
+            - "op_leaves" (optional): number of bits to quantize the leaves, must be less than or
+                equal to `op_inputs`. defaults to the value of "op_inputs" if not specified.
+
+    Raises:
+        ValueError: If 'n_bits' does not conform to the required format or value constraints.
+    """
+
+    detailed_message = (
+        "Invalid `n_bits`, either pass a non-null positive integer or a dictionary containing "
```
RomanBredehoft, line 53: back quotes / backicks (`) have a special meaning in python when printing them and it's supposed to be deprecated in python 3.0: https://stackoverflow.com/questions/1673071/what-do-backticks-mean-to-the-python-interpreter-example-num

in our case I think we just want to use single quotes (') instead here as we only want to refer to the name ! this goes for all or our error messages ect I've created an issue for it as I think we already have some other cases in the code : https://github.com/zama-ai/concrete-ml-internal/issues/4235

kcelia:
> > 2 - Would it be better to relaunch the weekly CI with TREE_USE_SUM_FHE enabled ?
> 
> not sure why ? we are testing the feature with `TREE_USE_SUM_FHE=1` in your new test right ? and for the dump onnx imo we should tests both in regular CIs

@RomanBredehoft , it's disabled.

src/concrete/ml/sklearn/base.py
kcelia: Because I got this error: `RuntimeError: Cannot clone object DecisionTreeClassifier(n_bits=3), as the constructor either does not set or modifies parameter n_bits` in test_pipline

kcelia: I assumed it was because of this error, that hadn't put it in the init

src/concrete/ml/quantization/post_training.py
kcelia: Thanks i wans't aware of this file.

I'll move the n_bits validity there

Thanks

src/concrete/ml/quantization/post_training.py
kcelia: ok

src/concrete/ml/quantization/post_training.py
kcelia: ok

tests/sklearn/test_sklearn_models.py
```diff
@@ -1871,3 +1949,63 @@ def test_rounding_consistency_for_regular_models(
...
+
+
+@pytest.mark.parametrize(
+    "n_bits, error_message",
+    [
+        (0, "n_bits must be a non-null, positive integer"),
+        (-1, "n_bits must be a non-null, positive integer"),
+        (
+            {"op_inputs": 4, "op_leaves": 2, "op_weights": 2},
+            "Invalid keys in `n_bits` dictionary. Only 'op_inputs' \\(mandatory\\) and 'op_leaves' "
```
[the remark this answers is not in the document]
kcelia: > are the "\" necessary ?
Yes other it doesn't catch the message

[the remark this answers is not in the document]
kcelia: > also, we have a test_n_bits.py (in quantization tests) file for such tests, maybe it's better to move that there (and use a similar workflow) !

Yes, as I said in a previous message, I'll move them there.

src/concrete/ml/quantization/post_training.py
kcelia: ok thanks

src/concrete/ml/quantization/post_training.py
kcelia: I just remembered an old PR by @bcm-at-zama  where he had to remove the '=' symbol, although I'm not entirely sure of the reason, it could have been related to a version compatibility issue.

tests/sklearn/test_sklearn_models.py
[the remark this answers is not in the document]
kcelia: It extend the regular CI runtime.
I'll bring this up at our next weekly meeting @zama-ai/machine-learning

src/concrete/ml/quantization/post_training.py
RomanBredehoft: ah yes probably with the 3.7 version, but we're not supporting it anymore !

src/concrete/ml/sklearn/base.py
RomanBredehoft: ah yes I see, pretty inconvenient convention from sklearn : https://stackoverflow.com/questions/24510510/python-scikit-learn-cannot-clone-object-as-the-constructor-does-not-seem-to ..

RomanBredehoft: thanks !

src/concrete/ml/sklearn/rf.py
RomanBredehoft: actually yes you're right , I've updated the issue in consequence !

src/concrete/ml/quantization/post_training.py
kcelia: done

tests/sklearn/test_sklearn_models.py
```diff
@@ -1200,6 +1207,76 @@ def check_rounding_consistency(
...
+    if is_weekly_option:
+        non_fhe_sum_predict_fhe = predict_method(fhe_test, fhe="execute")
+
+    with pytest.MonkeyPatch.context() as mp_context:
+
+        # Enable the FHE summation of tree ensemble outputs
+        mp_context.setenv("TREES_USE_FHE_SUM", "1")
+
+        # Check that the summation of tree ensemble outputs is enabled
+        fhe_sum_enabled = os.environ.get("TREES_USE_FHE_SUM") == "1"
```
RomanBredehoft, line 1253: same here

RomanBredehoft:
@kcelia yes by default but we do `mp_context.setenv("TREES_USE_FHE_SUM", "1")` in our tests to avoid that right ?

kcelia:
> @kcelia yes by default but we do mp_context.setenv("TREES_USE_FHE_SUM", "1") in our tests to avoid that right ?

I don't get your question, did you reply to a specific message ?

RomanBredehoft:
yes sorry I was answering to your latest message : 

> > > 2 - Would it be better to relaunch the weekly CI with TREE_USE_SUM_FHE enabled ?
> > 
> > 
> > not sure why ? we are testing the feature with `TREE_USE_SUM_FHE=1` in your new test right ? and for the dump onnx imo we should tests both in regular CIs
> 
> @RomanBredehoft , it's disabled.

src/concrete/ml/sklearn/base.py
```diff
@@ -96,7 +102,10 @@
 # Enable rounding feature for all tree-based models by default
 # Note: This setting is fixed and cannot be altered by users
 # However, for internal testing purposes, we retain the capability to disable this feature
-os.environ["TREES_USE_ROUNDING"] = "1"
+os.environ["TREES_USE_ROUNDING"] = os.environ.get("TREES_USE_ROUNDING", "1")
+
+# By default, the decision of the tree ensembles is made in clear
+os.environ["TREES_USE_FHE_SUM"] = os.environ.get("TREES_USE_FHE_SUM", "0")
```
fd0r, line 108: I thought we tried to avoid env variables? 🤔

kcelia: It's our third env variables =)

src/concrete/ml/sklearn/base.py
fd0r: Yes we should store this as an attribute in the object.

I'm not sure why we went for an env var and not a parameter in the object ?

fd0r (commented):
Thanks a lot for this feature @kcelia 🙏🏼 

I'm not sure about this whole env-var mechanism, why not have it as an option? We already introduced `fit_encrypted` for encrypted training, why not add an option here too?

@RomanBredehoft 's comment is very important, we need to check that we stay consitent between training/development and deployment.

src/concrete/ml/sklearn/base.py
kcelia: We don't want these options to be used easily by the users, but in other hands,  we want to test them internally. 

- For `TREES_USE_ROUNDING` we want to spot any changes during concrete python updates
- For `TREES_USE_FHE_SUM` it's disabled now, but it will be enabled in the future to enhance our model's security once all the necessary tools are in place.

andrei-stoian-zama (changes requested):

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: I think you are both right !

jfrery: The choice of the env variable was because this is not something we want people to use as it's more of an experimental feature. But now we want to make it compatible with the deployment. I find this a bit sketchy. If we want this to be an actual option then yes we should go for a parameter instead of a env variable. But the feature brought some performance loss in some cases thus why it's not the new default.

fd0r (changes requested):
Didn't we want to remove the need for this env var?

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,19 +1290,65 @@ def __init_subclass__(cls):
...
+    def use_fhe_sum(self) -> bool:
+        """Property getter for `use_fhe_sum`.
+
+        Returns:
+            bool: The current setting of the `_use_fhe_sum` attribute.
+        """
+        return self._use_fhe_sum
+
+    @use_fhe_sum.setter
+    def use_fhe_sum(self, value) -> None:
```
kcelia, line 1330: I added a property just to raise a message to the user, and notify the user about the need to refit the model when this feature is enabled.

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -252,7 +274,7 @@ def tree_onnx_graph_preprocessing(
         # sklearn models apply the reduce sum before the transpose.
         # To have equivalent output between xgboost in sklearn,
         # apply the transpose before returning the output.
-        add_transpose_after_last_node(onnx_model)
+        add_transpose_after_last_node(onnx_model, use_fhe_sum)
```
kcelia, line 277: this function is only used for trees.
So it won't alter the behavior.

tests/sklearn/test_dump_onnx.py
```diff
@@ -20,108 +19,19 @@
 # pylint: disable=line-too-long
 
 
-def check_onnx_file_dump(model_class, parameters, load_data, str_expected, default_configuration):
+def check_onnx_file_dump(model_class, parameters, load_data, default_configuration, use_fhe_sum):
```
kcelia, line 22: I had to move  `str_expected`  in `check_onnx_file_dump`

src/concrete/ml/sklearn/base.py
```diff
@@ -96,7 +102,10 @@
 # Enable rounding feature for all tree-based models by default
 # Note: This setting is fixed and cannot be altered by users
 # However, for internal testing purposes, we retain the capability to disable this feature
-os.environ["TREES_USE_ROUNDING"] = "1"
+os.environ["TREES_USE_ROUNDING"] = os.environ.get("TREES_USE_ROUNDING", "1")
```
fd0r, line 105: Still an environ variable here 🙂

fd0r (changes requested):
I see a remove env var commit but I still see `os.environ` usages 👀

src/concrete/ml/sklearn/base.py
```diff
@@ -96,7 +102,10 @@
...
+
+# By default, the decision of the tree ensembles is made in clear
+TREES_USE_FHE_SUM = False
```
RomanBredehoft, line 108: I think this is not needed anymore right ?

src/concrete/ml/sklearn/base.py
RomanBredehoft: yes good idea !

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -132,21 +136,31 @@ def assert_add_node_and_constant_in_xgboost_regressor_graph(onnx_model: onnx.Mod
     )
 
 
-def add_transpose_after_last_node(onnx_model: onnx.ModelProto):
+def add_transpose_after_last_node(onnx_model: onnx.ModelProto, use_fhe_sum: bool):
     """Add transpose after last node.
 
     Args:
         onnx_model (onnx.ModelProto): The ONNX model.
+        use_fhe_sum (bool): This parameter is exclusively used to tree-based models.
```
RomanBredehoft, line 144: `used to tree-based models` -> `used for tree-based models`

src/concrete/ml/sklearn/tree_to_numpy.py
RomanBredehoft: we are in the `tree_to_numpy` file (where this function only appears) so no worries yes

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -316,6 +341,8 @@ def tree_to_numpy(
         x (numpy.ndarray): The input data.
         use_rounding (bool): This parameter is exclusively used to tree-based models.
             It determines whether the rounding feature is enabled or disabled.
+        use_fhe_sum (bool): This parameter is exclusively used to tree-based models.
+            It determines whether the sum of the trees' outputs is computed in FHE.
```
RomanBredehoft, line 345: same, could you specifyc that the default is false ?

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -316,6 +341,8 @@ def tree_to_numpy(
         x (numpy.ndarray): The input data.
         use_rounding (bool): This parameter is exclusively used to tree-based models.
             It determines whether the rounding feature is enabled or disabled.
```
RomanBredehoft, line 343: good time to also add that the default is true here I guess 😅

tests/sklearn/test_sklearn_models.py
```diff
@@ -143,10 +144,16 @@ def preamble(model_class, parameters, n_bits, load_data, is_weekly_option):
...
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3979
     if get_model_name(model_class) == "KNeighborsClassifier":
-        # KNN can only be compiled with small quantization bit numbers for now
-        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3979
         n_bits = 2
+
+    # Adjust the quantization precision for tree-based model based on `TREES_USE_FHE_SUM` setting.
+    # When enabled, the circuit's bitwidth increases, potentially leading to Out-of-Memory issues.
+    # Therefore, the maximum quantization precision is 4 bits in this case.
+    elif model_class in _get_sklearn_tree_models() and os.environ.get("TREES_USE_FHE_SUM") == "1":
```
RomanBredehoft, line 155: this should be updated

tests/sklearn/test_sklearn_models.py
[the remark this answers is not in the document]
RomanBredehoft: just realized we haven't talked about this ! I still think we should at least do 1 sample in fhe in the regular CI

if we start doing tests only in simulation in regular CIs, then in that case why not doing it for all regular CIs ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1871,3 +1921,87 @@ def test_rounding_consistency_for_regular_models(
...
+)
+@pytest.mark.parametrize("model_class", _get_sklearn_tree_models())
+def test_invalid_n_bits_setting(model_class, n_bits, error_message):
+    """Check if the model instantiation raises an exception with invalid 'n_bits' settings."""
+
+    with pytest.raises(ValueError, match=f"{error_message}. Got '{type(n_bits)}' and '{n_bits}'.*"):
+        instantiate_model_generic(model_class, n_bits=n_bits)
+
+
+@pytest.mark.parametrize("n_bits", [5, {"op_inputs": 5}, {"op_inputs": 2, "op_leaves": 1}])
```
RomanBredehoft, line 1985: could you add a fixme indicating that this test should be extended to all built-ins please ? using this issue :  https://github.com/zama-ai/concrete-ml-internal/issues/4234

tests/sklearn/test_sklearn_models.py
```diff
@@ -1871,3 +1921,87 @@ def test_rounding_consistency_for_regular_models(
...
+        ),
+        (
+            {"op_inputs": -2, "op_leaves": -5},
+            "All values in 'n_bits' dictionary must be non-null, positive integers",
+        ),
+        ({"op_inputs": 2, "op_leaves": 5}, "'op_leaves' must be less than or equal to 'op_inputs'"),
+        (0.5, "n_bits must be either an integer or a dictionary"),
+    ],
+)
+@pytest.mark.parametrize("model_class", _get_sklearn_tree_models())
```
RomanBredehoft, line 1977: same as below

tests/sklearn/test_dump_onnx.py
```diff
@@ -457,4 +402,105 @@ def test_dump(
...
+            module__power_of_two_scaling=False,
+            max_epochs=1,
+            verbose=0,
+            callbacks="disable",
+        )
+
+    check_onnx_file_dump(
+        model_class, parameters, load_data, default_configuration, use_fhe_sum=False
+    )
+    check_onnx_file_dump(
```
RomanBredehoft, line 504: running the check with `use_fhe_sum=True` is only necessary for trees, could you add a `if model_class in _get_sklearn_tree_models()` before that run ?

src/concrete/ml/sklearn/base.py
kcelia: Oh yes, forgot to remove it.
Thanks

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -213,6 +230,8 @@ def tree_onnx_graph_preprocessing(
         framework (str): The framework from which the ONNX model is generated.
             (options: 'xgboost', 'sklearn')
         expected_number_of_outputs (int): The expected number of outputs in the ONNX model.
+        use_fhe_sum (bool): This parameter is exclusively used to tree-based models.
+            It determines whether the sum of the trees' outputs is computed in FHE.
```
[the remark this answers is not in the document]
kcelia: Of course yes.

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,19 +1290,65 @@ def __init_subclass__(cls):
...
+        """
+
+        assert isinstance(value, bool), "Value must be a boolean type"
+
+        if value is True:
+            warnings.simplefilter("always")
+            warnings.warn(
+                "Enabling `use_fhe_sum` computes the sum of the ouputs of tree ensembles in FHE.\n"
+                "This may slow down the computation and increase the maximum bitwidth.\n"
+                "To optimize performance, consider reducing the quantization leaf precision.\n"
```
jfrery, line 1344: Can they really decrease the output quantization precision? I don't think we have made this feature available

src/concrete/ml/sklearn/tree_to_numpy.py
jfrery, line 234: I don't think we need to specify that the parameter is for tree based model since it's in the tree...py file. Also, what happens for decision tree models that have only 1 tree?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1171,8 +1177,7 @@ def check_rounding_consistency(
         mp_context.setenv("TREES_USE_ROUNDING", "0")
 
         # Check that rounding is disabled
-        rounding_disabled = os.environ.get("TREES_USE_ROUNDING") == "0"
-        assert rounding_disabled
+        assert os.environ.get("TREES_USE_ROUNDING") == "0", "'TREES_USE_ROUNDING' is not disabled"
```
jfrery, line 1180: Didn't we say we remove those env variable?

jfrery (changes requested):
Thanks! I have a few comments.

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,19 +1290,65 @@ def __init_subclass__(cls):
...
+
+    @use_fhe_sum.setter
+    def use_fhe_sum(self, value) -> None:
+        """Property setter for `use_fhe_sum`.
+
+        Args:
+            value (int): Whether to enable or disable the feature.
+        """
+
+        assert isinstance(value, bool), "Value must be a boolean type"
```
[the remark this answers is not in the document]
kcelia: why ?

src/concrete/ml/sklearn/base.py
kcelia: > I don't think we have made this feature available

Yes, I added it in this PR, see: src/concrete/ml/quantization/post_training.py::_get_n_bits_dict_trees

tests/sklearn/test_dump_onnx.py
kcelia: ok

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,19 +1290,65 @@ def __init_subclass__(cls):
...
         #: The model's inference function. Is None if the model is not fitted.
         self._tree_inference: Optional[Callable] = None
 
+        #: Wether to perform the sum of the output's tree ensembles in FHE or not.
+        self._use_fhe_sum = False
+
         BaseEstimator.__init__(self)
 
+    @property
+    def use_fhe_sum(self) -> bool:
```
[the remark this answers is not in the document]
andrei-stoian-zama: maybe `fhe_ensembling`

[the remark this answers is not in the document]
jfrery: sounds good!

src/concrete/ml/sklearn/base.py
jfrery: oh awesome! thanks

tests/sklearn/test_sklearn_models.py
jfrery: Ah yeah my bad. Let's not mix everything in one PR.

src/concrete/ml/sklearn/base.py
```diff
@@ -96,7 +102,7 @@
 # Enable rounding feature for all tree-based models by default
 # Note: This setting is fixed and cannot be altered by users
 # However, for internal testing purposes, we retain the capability to disable this feature
-os.environ["TREES_USE_ROUNDING"] = "1"
+os.environ["TREES_USE_ROUNDING"] = os.environ.get("TREES_USE_ROUNDING", "1")
```
kcelia, line 105: This env variable is for ROUNDING not for the sum of the outputs of the tree ensembles in fhe

src/concrete/ml/sklearn/base.py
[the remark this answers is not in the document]
kcelia: Thanks, I'll keep it.

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,99 @@
...
+            - "op_inputs" (mandatory): number of bits to quantize the input values
+            - "op_leaves" (optional): number of bits to quantize the leaves, must be less than or
+                equal to 'op_inputs. defaults to the value of 'op_inputs if not specified.
+
+    Raises:
+        ValueError: If 'n_bits' does not conform to the required format or value constraints.
+    """
+
+    detailed_message = (
+        "Invalid 'n_bits', either pass a non-null positive integer or a dictionary containing "
```
andrei-stoian-zama, line 53: `a non-null positive integer` is not very clear. Simply a "strictly positive integer" should be better. there are several occurrences.

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,19 +1287,67 @@ def __init_subclass__(cls):
...
+        """Property setter for `fhe_ensembling`.
+
+        Args:
+            value (bool): Whether to enable or disable the feature.
+        """
+
+        assert isinstance(value, bool), "Value must be a boolean type"
+
+        if value is True:
+            print("LAA")
```
andrei-stoian-zama, line 1338: this print not needed

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,19 +1287,67 @@ def __init_subclass__(cls):
...
 
+        #: Wether to perform the sum of the output's tree ensembles in FHE or not.
+        # By default, the decision of the tree ensembles is made in clear.
+        self._fhe_ensembling = False
+
         BaseEstimator.__init__(self)
 
+    @property
+    def fhe_ensembling(self) -> bool:
+        """Property getter for `_fhe_ensembling`.
```
andrei-stoian-zama, line 1320: I don't understand why there is a "setter". This value should be set in the constructor, like `n_bits`. It should also be serialized

andrei-stoian-zama (changes requested):

src/concrete/ml/sklearn/tree_to_numpy.py
kcelia: Decision trees without the sum in FHE:
image

kcelia: @jfrery @andrei-stoian-zama, we have a new node for decision tree with 1 tree. 
We are like 1.1 slower

src/concrete/ml/sklearn/base.py
kcelia: As discussed in https://zama-ai.slack.com/archives/C0455QDGT6C/p1706087454383859

If the user alters a value of a private member outside of a class,  undefined behaviors are expected.

tests/sklearn/test_dump_onnx.py
```diff
@@ -457,4 +405,106 @@ def test_dump(
...
+            module__power_of_two_scaling=False,
+            max_epochs=1,
+            verbose=0,
+            callbacks="disable",
+        )
+
+    check_onnx_file_dump(model_class, parameters, load_data, default_configuration)
+
+    # Additional tests exclusively dedicated for tree ensemble models.
+    if model_class in _get_sklearn_tree_models()[2:]:
```
RomanBredehoft, line 507: I'm not sure what is this `[2:]` ? Is it in order to avoid decision tree models ? In that cas could you do `_get_sklearn_tree_models(select=["RandomForest", "XGB"])` instead ? it's more explicit that way

kcelia: thanks, it was a mistake !

kcelia: thanks, it was a mistake !

andrei-stoian-zama (approved):
good for me

tests/sklearn/test_sklearn_models.py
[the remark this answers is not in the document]
RomanBredehoft: I still think this is an important question !! for me, either : 
- we do like always in this test, ie we do 1 inference in FHE in the regular CI and 5 in the weekly CI
- we only run in simulation in the regular CI and 5 FHE runs in the weekly CI, but in that case why not doing so for the other tests ? I've summarized the situation here : https://github.com/zama-ai/concrete-ml-internal/issues/4248

tests/sklearn/test_sklearn_models.py
```diff
@@ -1871,3 +1917,91 @@ def test_rounding_consistency_for_regular_models(
...
+        model,
+        x,
+        y,
+        predict_method,
+        is_weekly_option,
+    )
+
+
+# This test should be extended to all built-in models.
+# FIXME: https://github.com/zama-ai/concrete-ml-internal#4234
```
RomanBredehoft, line 1955: I think you wanted to put `https://github.com/zama-ai/concrete-ml-internal/issues/4234`

tests/sklearn/test_sklearn_models.py
```diff
@@ -1871,3 +1917,91 @@ def test_rounding_consistency_for_regular_models(
...
+@pytest.mark.parametrize("model_class", _get_sklearn_tree_models())
+def test_invalid_n_bits_setting(model_class, n_bits, error_message):
+    """Check if the model instantiation raises an exception with invalid `n_bits` settings."""
+
+    with pytest.raises(ValueError, match=f"{error_message}. Got '{type(n_bits)}' and '{n_bits}'.*"):
+        instantiate_model_generic(model_class, n_bits=n_bits)
+
+
+# This test should be extended to all built-in models.
+# FIXME: https://github.com/zama-ai/concrete-ml-internal#4234
```
RomanBredehoft, line 1984: same : `https://github.com/zama-ai/concrete-ml-internal/issues/4234`

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -132,21 +136,31 @@ def assert_add_node_and_constant_in_xgboost_regressor_graph(onnx_model: onnx.Mod
...
 
 
-def add_transpose_after_last_node(onnx_model: onnx.ModelProto):
+def add_transpose_after_last_node(onnx_model: onnx.ModelProto, fhe_ensembling: bool):
     """Add transpose after last node.
 
     Args:
         onnx_model (onnx.ModelProto): The ONNX model.
+        fhe_ensembling (bool): Determines whether the sum of the trees' outputs is computed in FHE.
+            Default to False.
```
RomanBredehoft, line 145: there is no default in this function

RomanBredehoft (changes requested):
a few very little remaining comments and we're good !

tests/sklearn/test_sklearn_models.py
kcelia: damn bad copy paste

src/concrete/ml/sklearn/base.py
```diff
@@ -1281,17 +1287,33 @@ def __init_subclass__(cls):
...
+                    - op_inputs (mandatory): number of bits to quantize the input values
+                    - op_leaves (optional): number of bits to quantize the leaves
+                Default to 6.
         """
-        self.n_bits: int = n_bits
+
+        # Check if 'n_bits' is a valid value.
+        _inspect_tree_n_bits(n_bits)
+
+        #: The number of bits to quantize the model.
```
jfrery, line 1306: What are these `#:` ? Do they mean something special?

src/concrete/ml/quantization/post_training.py
```diff
@@ -23,6 +23,99 @@
...
+
+
+def _inspect_tree_n_bits(n_bits):
+    """Validate the 'n_bits' parameter for tree-based models.
+
+    This function checks whether 'n_bits' is a valid integer or dictionary.
+    - If 'n_bits' is an integer, it must be a non-null positive, its value is assigned to
+        'op_inputs' and 'op_leaves' bits
+    - If it is a dictionary, it should contain integer values for keys 'op_leaves' and 'op_inputs',
+        where 'op_leaves' should not exceed 'op_inputs'.
```
jfrery, line 36: Not sure about the namings here. What about `n_bits_input` `n_bits_output`?

jfrery (changes requested):

src/concrete/ml/quantization/post_training.py
andrei-stoian-zama: NNs have `op_inputs` `op_outputs`. It keeps that style, I think it's fine like this.

src/concrete/ml/sklearn/tree_to_numpy.py
andrei-stoian-zama: ok, no big deal, as long as it produces correct results

src/concrete/ml/sklearn/base.py
kcelia: Good question,  it's a comment style, for class attribute.
You can find them all over our code.

src/concrete/ml/sklearn/tree_to_numpy.py
kcelia: I think that we shouldn't bother with that now, especially since the feature is currently disabled

src/concrete/ml/sklearn/base.py
RomanBredehoft: it's for sphinx (since we are still supporting it), it enables documenting instances attributes that are not defined by the init  (the docstring only documents init parameters) : 
- https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoproperty
- https://stackoverflow.com/questions/8649105/how-to-show-instance-attributes-in-sphinx-doc

works also for class attributes, although in this case we can simply have a `Attributes:` section in the class' docstring (ex: `_is_a_public_cml_model` in `BaseEstimator`) 

although you are right, in this case it shouldn't really appear as the n_bits is defined in the init's docstring so the comment could be removed

src/concrete/ml/quantization/post_training.py
RomanBredehoft: @jfrery the reason is that QNNs have `model_inputs`, `model_outputs` , `op_inputs` and `op_weights`, so you need to differentiate both. As Andrei said, here, it's taking the same convention

although I agree for trees it's getting q bit misleading without much context but we don't have much choice as we should keep our namings coherent with one another

my main comment on this naming was about `op_leave` (https://github.com/zama-ai/concrete-ml/pull/449#discussion_r1453237293), this introduces a new name into our convention. Although I get why it's more relevant to tree models, I felt keeping `op_weights` was fine imo

RomanBredehoft (changes requested):
a few remaining comments we but we're close !

src/concrete/ml/quantization/post_training.py
jfrery: Well I find op_ not ideal for tree based models. Also if we want to keep consistency then it should be model_inputs instead of op_inputs.

I am really not a fan of having op_leave for many reasons. Most users won't know what leave refer to. Same for op.

The why not `model_input` and `model_output`? It makes sense for the input but would require a change for `model_output`. This will help us maintain the output at a given precision which is not easy with op_leave since this targets every tree instead of the final combination.

RomanBredehoft: I agree with you I'm not a fan of `op_leave`. but about having `model_input` and `model_output` only, how are you going to handle the n_bits to use for  quantizing the leaves ? you'll either need an additional parameter similar to the suggested `op_leave` or a way to infer its value based on `model_output` but not sure how this could easily/efficiently be done 🤔

kcelia: @zama-ai/machine-learning, please let's agree on a proper naming.
I suggest: tree_input_bits/tree_leaf_bits or quant_input_bits/quant_leaf_bits

-> model_ouput why not, but doesn't ring like it's the leaves that we are quantizing
-> op_weight, really not a big fan, we are quantizing leaves not weights.

RomanBredehoft: I think it's best if you ask directly in the slack channel !

tests/sklearn/test_sklearn_models.py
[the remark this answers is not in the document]
RomanBredehoft: so based on what we've discussed here (https://zama-ai.slack.com/archives/C0455QDGT6C/p1706176888681709), it seems that it's preferable to continue running at least a single inference in FHE in the regular CI until we enable the new simulation everywhere

src/concrete/ml/sklearn/base.py
fd0r: Still do we want to keep this env var?

src/concrete/ml/sklearn/base.py
```diff
@@ -1319,7 +1345,7 @@ def fit(self, X: Data, y: Target, **fit_parameters):
         # Check that the underlying sklearn model has been set and fit
         assert self.sklearn_model is not None, self._sklearn_model_is_not_fitted_error_message()
 
-        # Convert the tree inference with Numpy operators
+        # Enable rounding feature
         enable_rounding = os.environ.get("TREES_USE_ROUNDING", "1") == "1"
```
fd0r, line 1349: Still env var based

tests/sklearn/test_sklearn_models.py
fd0r: When did we decide that?
I'm fine if this is done in another PR but would prefer to not have to rely on env vars.

fd0r (approved):
Fine with me but we should get rid of the env var used for rounding in a follow-up PR

src/concrete/ml/sklearn/base.py
kcelia: this one, is for rounding, not fhe_sum.

If we decide to remove it as well, it should be in another PR.

RomanBredehoft:
@kcelia it would be better if you could re-run the CI once more to make sure we have full coverage btw !

RomanBredehoft (approved):
Huge work, thanks a lot !!

fd0r (approved):
Thanks a lot, great work!
Let's remove the other env-var in a follow-up PR 😉

kcelia:
Hi @jfrery, could you approve it please ?
We have been waiting for you to merge it =)

jfrery (approved):
Yes approved thanks!

PR #453, opened 2024-01-18
docs: document n_bits for compile torch functions

andrei-stoian-zama:
The `n_bits` configuration for `compile_torch_model` did not explain that weights&activations can have different bitwidths. I added this explanation. 

closes https://github.com/zama-ai/concrete-ml-internal/issues/2365

### DIFF
--- a/src/concrete/ml/torch/compile.py
+++ b/src/concrete/ml/torch/compile.py
@@ -158,7 +158,12 @@ def _compile_torch_or_onnx_model(
         artifacts (DebugArtifacts): Artifacts object to fill during compilation
         show_mlir (bool): if set, the MLIR produced by the converter and which is going
             to be sent to the compiler backend is shown on the screen, e.g., for debugging or demo
-        n_bits: the number of bits for the quantization
+        n_bits: number of bits for quantization, can be a single value or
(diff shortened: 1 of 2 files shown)


src/concrete/ml/torch/compile.py
```diff
@@ -158,7 +158,12 @@ def _compile_torch_or_onnx_model(
         artifacts (DebugArtifacts): Artifacts object to fill during compilation
         show_mlir (bool): if set, the MLIR produced by the converter and which is going
             to be sent to the compiler backend is shown on the screen, e.g., for debugging or demo
-        n_bits: the number of bits for the quantization
+        n_bits: number of bits for quantization, can be a single value or
```
RomanBredehoft, line 161: better to write `n_bits (Union[int, Dict[str, int]])` for all. Also, maybe indicate that the default is 8 bits

RomanBredehoft (changes requested):
great thanks ! needs a bit more details though 

also, I think this PR closes this issue : https://github.com/zama-ai/concrete-ml-internal/issues/2365

kcelia (approved):

RomanBredehoft (approved):
thanks !

PR #462, opened 2024-01-23
chore: refacto quantized module init + handle no input onnx

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4242

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,6 +143,8 @@ filterwarnings = [
     "ignore:non-integer arguments to randrange\\(\\) have been deprecated since Python 3\\.10 and will be removed in a subsequent version:DeprecationWarning",
     "ignore:Deprecated call to `pkg_resources.declare_namespace:DeprecationWarning",
     "ignore:The --rsyncdir command line argument and rsyncdirs config variable are deprecated.:DeprecationWarning",
+    "ignore:Converting a tensor to a NumPy array might cause the trace to be incorrect.",
+    "ignore:torch.from_numpy results are registered as constants in the trace.",
 ]
 
 [tool.semantic_release]
--- a/src/concrete/ml/onnx/convert.py
+++ b/src/concrete/ml/onnx/convert.py
@@ -12,6 +12,7 @@ import onnxoptimizer
 import torch
 from onnx import checker, helper
 
+from ..common.debugging import assert_true
 from .onnx_utils import (
     IMPLEMENTED_ONNX_OPS,
     execute_onnx_with_numpy,
@@ -149,6 +150,14 @@ def get_equivalent_numpy_forward_from_torch(
         input_names=arguments,
     )
     equivalent_onnx_model = onnx.load_model(str(output_onnx_file_path))
+
+    # Check if the inputs are present in the model's graph
+    for input_name in arguments:
+        assert_true(
+            any(input_name == node.name for node in equivalent_onnx_model.graph.input),
+            f"Input '{input_name}' is not present in the ONNX model. Please check the onnx graph.",
+        )
+
     # Remove the tempfile if we used one
     if use_tempfile:
         output_onnx_file_path.unlink()
--- a/src/concrete/ml/onnx/convert.py
+++ b/src/concrete/ml/onnx/convert.py
@@ -155,7 +155,8 @@ def get_equivalent_numpy_forward_from_torch(
     for input_name in arguments:
         assert_true(
             any(input_name == node.name for node in equivalent_onnx_model.graph.input),
-            f"Input '{input_name}' is not present in the ONNX model. Please check the onnx graph.",
+            f"Input '{input_name}' is missing in the ONNX graph after export. "
+            "Verify the forward pass for issues.",
         )
 
     # Remove the tempfile if we used one
--- a/src/concrete/ml/onnx/ops_impl.py
+++ b/src/concrete/ml/onnx/ops_impl.py
@@ -1479,10 +1479,9 @@ def numpy_pad(
 def numpy_cast(data: numpy.ndarray, *, to: int) -> Tuple[numpy.ndarray]:
     """Execute ONNX cast in Numpy.
 
-    For traced values during compilation, it supports booleans and floats,
-    which are converted to float. For raw values (used in constant folding or shape computations),
+    This function supports casting to booleans, floats, and double for traced values,
+    converting them accordingly. For raw values (used in constant folding or shape computations),
     any cast is allowed.
-
     See: https://github.com/onnx/onnx/blob/main/docs/Operators.md#Cast
 
     Args:
@@ -1496,8 +1495,12 @@ def numpy_cast(data: numpy.ndarray, *, to: int) -> Tuple[numpy.ndarray]:
     if isinstance(data, RawOpOutput):
         return (data.astype(onnx.helper.tensor_dtype_to_np_dtype(to)).view(RawOpOutput),)
 
-    # Allow both bool and float types
-    assert_true(to in (onnx.TensorProto.BOOL, onnx.TensorProto.FLOAT))
+    allowed_types = (onnx.TensorProto.BOOL, onnx.TensorProto.FLOAT, onnx.TensorProto.DOUBLE)
+    assert to in allowed_types, (
+        f"Invalid 'to' data type: {onnx.TensorProto.DataType.Name(to)}. "
+        f"Only {', '.join(onnx.TensorProto.DataType.Name(t) for t in allowed_types)}"
+        "are allowed for casting."
+    )
 
     # Will be used for traced values
     return (data.astype(numpy.float64),)
--- a/src/concrete/ml/onnx/ops_impl.py
+++ b/src/concrete/ml/onnx/ops_impl.py
@@ -1479,8 +1479,9 @@ def numpy_pad(
 def numpy_cast(data: numpy.ndarray, *, to: int) -> Tuple[numpy.ndarray]:
     """Execute ONNX cast in Numpy.
 
-    For traced values during compilation, it supports only booleans, which are converted to float.
-    For raw values (used in constant folding or shape computations), any cast is allowed.
+    For traced values during compilation, it supports booleans and floats,
+    which are converted to float. For raw values (used in constant folding or shape computations),
+    any cast is allowed.
 
     See: https://github.com/onnx/onnx/blob/main/docs/Operators.md#Cast
 
@@ -1495,7 +1496,8 @@ def numpy_cast(data: numpy.ndarray, *, to: int) -> Tuple[numpy.ndarray]:
     if isinstance(data, RawOpOutput):
         return (data.astype(onnx.helper.tensor_dtype_to_np_dtype(to)).view(RawOpOutput),)
 
-    assert_true(to == onnx.TensorProto.BOOL)
+    # Allow both bool and float types
+    assert_true(to in (onnx.TensorProto.BOOL, onnx.TensorProto.FLOAT))
 
     # Will be used for traced values
     return (data.astype(numpy.float64),)
--- a/src/concrete/ml/quantization/quantized_module.py
+++ b/src/concrete/ml/quantization/quantized_module.py
@@ -99,28 +99,39 @@ class QuantizedModule:
         # Set base attributes for API consistency. This could be avoided if an abstract base class
         # is created for both Concrete ML models and QuantizedModule
         # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/2899
-        self.fhe_circuit = None
+
+        all_or_none_params = [
+            ordered_module_input_names,
+            ordered_module_output_names,
+            quant_layers_dict,
+        ]
+        assert_true(
+            all(v is None or v == {} for v in all_or_none_params)
+            or not any(v is None or v == {} for v in all_or_none_params),
+            "All of ordered_module_input_names, ordered_module_output_names, "
+            "and quant_layers_dict must be provided if any one of them is provided.",
+        )
+        self.ordered_module_input_names = (
+            tuple(ordered_module_input_names) if ordered_module_input_names else ()
+        )
+        self.ordered_module_output_names = (
+            tuple(ordered_module_output_names) if ordered_module_output_names else ()
+        )
+        self.quant_layers_dict = (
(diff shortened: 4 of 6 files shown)


pyproject.toml
```diff
@@ -143,6 +143,8 @@ filterwarnings = [
     "ignore:non-integer arguments to randrange\\(\\) have been deprecated since Python 3\\.10 and will be removed in a subsequent version:DeprecationWarning",
     "ignore:Deprecated call to `pkg_resources.declare_namespace:DeprecationWarning",
     "ignore:The --rsyncdir command line argument and rsyncdirs config variable are deprecated.:DeprecationWarning",
+    "ignore:Converting a tensor to a NumPy array might cause the trace to be incorrect.",
```
RomanBredehoft, line 146: I wonder if ignoring these warnings for the whole project is a good thing

src/concrete/ml/onnx/ops_impl.py
```diff
@@ -1495,7 +1496,8 @@ def numpy_cast(data: numpy.ndarray, *, to: int) -> Tuple[numpy.ndarray]:
     if isinstance(data, RawOpOutput):
         return (data.astype(onnx.helper.tensor_dtype_to_np_dtype(to)).view(RawOpOutput),)
 
-    assert_true(to == onnx.TensorProto.BOOL)
+    # Allow both bool and float types
+    assert_true(to in (onnx.TensorProto.BOOL, onnx.TensorProto.FLOAT))
```
andrei-stoian-zama, line 1500: please add an assert message

andrei-stoian-zama (changes requested):

src/concrete/ml/onnx/convert.py
```diff
@@ -149,6 +150,14 @@ def get_equivalent_numpy_forward_from_torch(
         input_names=arguments,
     )
     equivalent_onnx_model = onnx.load_model(str(output_onnx_file_path))
+
+    # Check if the inputs are present in the model's graph
+    for input_name in arguments:
+        assert_true(
+            any(input_name == node.name for node in equivalent_onnx_model.graph.input),
+            f"Input '{input_name}' is not present in the ONNX model. Please check the onnx graph.",
```
RomanBredehoft, line 158: I would maybe:
- say `Input '{input_name}' (found in the forward pass)` to be more precise 
- detail a bit more `Please check the onnx graph.` : what should the user do ? if he just builds a torch model, saves it to onnx but then encounters this problem, what could likely be the reason for it ?

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -99,28 +99,39 @@ def __init__(
         # Set base attributes for API consistency. This could be avoided if an abstract base class
         # is created for both Concrete ML models and QuantizedModule
         # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/2899
```
RomanBredehoft, line 101: better to move this fixme right above `self.input_quantizers` because it talks about these attributes, no the `self.ordered_module_input_names` and similar

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -99,28 +99,39 @@ def __init__(
...
+
+        all_or_none_params = [
+            ordered_module_input_names,
+            ordered_module_output_names,
+            quant_layers_dict,
+        ]
+        assert_true(
+            all(v is None or v == {} for v in all_or_none_params)
+            or not any(v is None or v == {} for v in all_or_none_params),
+            "All of ordered_module_input_names, ordered_module_output_names, "
```
RomanBredehoft, line 111: I would rephrase this by "Please either set all three 'ordered_module_input_names', 'ordered_module_output_names' and 'quant_layers_dict' or none of them" 

also, we should avoid putting `assert` or `assert_true` for user inputs like this and instead add ValueError raises + an associate small test (cf https://github.com/zama-ai/concrete-ml-internal/issues/4232)

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -99,28 +99,39 @@ def __init__(
...
-        assert all([ordered_module_input_names, ordered_module_output_names, quant_layers_dict])
-        self.ordered_module_input_names = tuple(ordered_module_input_names)
-        self.ordered_module_output_names = tuple(ordered_module_output_names)
-
-        assert quant_layers_dict is not None
-        self.quant_layers_dict = copy.deepcopy(quant_layers_dict)
-
-        self.output_quantizers = self._set_output_quantizers()
+        # Initialize output quantizers based on quant_layers_dict
+        if self.quant_layers_dict:
```
RomanBredehoft, line 131: I don't understand the modifications in this file, what happens if we don't give any of the inputs like `quant_layers_dict` ? are the `output_quantizers` built later ? I don't see where this is used in the new test 😅

tests/torch/test_compile_torch.py
```diff
@@ -1373,3 +1373,32 @@ def test_mono_parameter_rounding_warning(
...
+            """Forward pass."""
+            zeros = numpy.zeros_like(x)
+            x = x + zeros
+            x = (x > self.threshold).to(torch.float32)
+            x = self.fc1(x)
+            return x
+
+    model = SimplifiedNet()
+
+    with pytest.raises(AssertionError) as excinfo:
```
RomanBredehoft, line 1401: as said above, this should be a ValueError then !

src/concrete/ml/onnx/ops_impl.py
RomanBredehoft, line 1500: yes not sure where this is used or tested !

pyproject.toml
jfrery: It's for pytest only

src/concrete/ml/quantization/quantized_module.py
jfrery: This is just some refacto so that we actually have default values for all attributes. It fixes nothing though. The test would be to check that attribute at indeed initialized to something but I find this pretty irrelevant.

tests/torch/test_compile_torch.py
jfrery: Oh yeah I forgot about this. Will fix.

src/concrete/ml/onnx/ops_impl.py
jfrery: Done

src/concrete/ml/onnx/convert.py
jfrery: > say Input '{input_name}' (found in the forward pass) to be more precise

```
f"'{input_name}' is missing in the ONNX graph after export. Verify the forward pass for issues."
```

But  about

> detail a bit more Please check the onnx graph. : what should the user do ? if he just builds a torch model, saves it to onnx but then encounters this problem, what could likely be the reason for it ?

I have no idea. For the test we added it was to not use numpy in the forward but I am sure there are tons of different ways to make an onnx input disappear and so as much reasons.

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -99,28 +99,39 @@ def __init__(
...
+        )
+        self.ordered_module_output_names = (
+            tuple(ordered_module_output_names) if ordered_module_output_names else ()
+        )
+        self.quant_layers_dict = (
+            copy.deepcopy(quant_layers_dict) if quant_layers_dict is not None else {}
+        )
+        self.input_quantizers: List[UniformQuantizer] = []
+        self.output_quantizers: List[UniformQuantizer] = []
+        self.fhe_circuit: Union[None, Circuit] = None
```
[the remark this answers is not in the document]
jfrery: Yes sure

src/concrete/ml/quantization/quantized_module.py
jfrery: Sure I rephrased.

About the second part, it's not really user inputs here. Users use the QuantizedModule through the PTQ. I can add a test anyway yes.

andrei-stoian-zama (approved):
perfect!

pyproject.toml
RomanBredehoft: ah yes indeed, but still these warnings seem pretty important no ?

src/concrete/ml/onnx/ops_impl.py
RomanBredehoft: but this test itself is meant for another feature. Could you maybe add a specific test that only does this `to(float)` ?

jfrery: I have a test that check different cast yes.

tests/torch/test_compile_torch.py
jfrery: Should be good.

src/concrete/ml/onnx/ops_impl.py
RomanBredehoft: ah yes sorry didn't see, great then thanks !

src/concrete/ml/quantization/quantized_module.py
jfrery: The only different is that now, we initialize all attributes. Before we were just returning so some attributes were not present. It was just bad practice. This refacto should not change anything. Testing that attributes are properly initialized could be done but then why not doing it everywhere.

src/concrete/ml/onnx/convert.py
RomanBredehoft: ah yes I see, well ok we can keep it that way I guess. anyway yes for the first point I like your suggestion

src/concrete/ml/onnx/ops_impl.py
```diff
@@ -1495,7 +1495,12 @@ def numpy_cast(data: numpy.ndarray, *, to: int) -> Tuple[numpy.ndarray]:
...
-    assert_true(to == onnx.TensorProto.BOOL)
+    allowed_types = (onnx.TensorProto.BOOL, onnx.TensorProto.FLOAT, onnx.TensorProto.DOUBLE)
+    assert to in allowed_types, (
+        f"Invalid 'to' data type: {onnx.TensorProto.DataType.Name(to)}. "
+        f"Only {', '.join(onnx.TensorProto.DataType.Name(t) for t in allowed_types)}"
+        "are allowed for casting."
+    )
 
     # Will be used for traced values
     return (data.astype(numpy.float64),)
```
RomanBredehoft, line 1506: not sure what is happening here then : we check that `to` is a bool, float or double, but we only cast to double without talking `to` into account ?

pyproject.toml
jfrery: Yes! IMO these warning should be runtime errors. Not sure why torch allows exporting to onnx with broken forward.

But in pytest, warnings == failures. So we have to filter these.

RomanBredehoft (changes requested):
a few remaining comments from my side !

tests/torch/test_compile_torch.py
RomanBredehoft: actually you were right not to change anything here this was a legit assert my bad

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: it's still a public class that we even can serialize like any built-in models. But actually I meant "inputs" more than 'user inputs", asserts should not really be used to check input value formats

src/concrete/ml/onnx/ops_impl.py
jfrery: Yes it's pretty weird indeed. And before it was a cast to bool and then we casted to float64 so even weirder.

I think the idea is that we want to be able to support cast (i.e. not break because of a cast in the torch forward) but we don't want to change the FHE semantic. Casting to anything else than float64 is dangerous as it could be interpreted by concrete-python improperly and close a LUT.

Actually as @andrei-stoian-zama stated casting to float64 could be potentially dangerous as well if inputs where integers as it would mean opening a LUT (while the cast actually as no real effect in FHE).

tests/torch/test_compile_torch.py
jfrery: not sure I understood? I actually think you were right? Why not using the feature from pytest.raises instead of creating a new line with an assert?

src/concrete/ml/onnx/ops_impl.py
RomanBredehoft: ah yes actually now that I think of it, I remember already having some issues with that `numpy_cast` in the past

it feels like there's something we should do about your last point though !

src/concrete/ml/onnx/convert.py
jfrery: I changed the message. Should be good.

src/concrete/ml/onnx/ops_impl.py
jfrery: But would this kind of check integer be for the numpy cast only? I think this is a more general concern. Also we have function that do work on integers (tree specific numpy method) and some function that work on both (numpy_gemm).

pyproject.toml
RomanBredehoft: but if these warnings should be considered warnings, then maybe it's best if we add a `with pytest.raise` for them instead of ignoring them no ? 🤔

src/concrete/ml/onnx/ops_impl.py
RomanBredehoft: yes yes for sure this is more of a general concern, maybe we should summarize this in a dedicated issue ?

jfrery: Will do

pyproject.toml
fd0r: This isn't a warnings raised from CML so I see no issue in just ignoring it.
Using `with pytest.raise` would mean that we want the warning to be raised.

src/concrete/ml/onnx/ops_impl.py
fd0r: We could maybe have something that does:
```
if not int: return float
if int: return int
```
or smth like that

fd0r (approved):
Fine with me as is

pyproject.toml
RomanBredehoft: yes ok you're right, fine  for me then !

RomanBredehoft (approved):
fine with we thanks a lot ! just dpn't forget to create an issue for the `numpy_cast` part !

PR #471, opened 2024-01-29
chore: update commit rule for CI

RomanBredehoft:
I believe the rule could be simplified as we only use this pattern


kcelia (dismissed):
Thanks

.github/workflows/continuous-integration.yaml
```diff
@@ -425,10 +425,10 @@ jobs:
         if: ${{ fromJSON(env.IS_PR) && steps.install-deps.outcome == 'success' && !cancelled() }}
         uses: gsactions/commit-message-checker@16fa2d5de096ae0d35626443bcd24f1e756cafee
         with:
-          pattern: '^((feat|fix|chore|refactor|style|test|docs)(\((bounds|helpers|data_types|debugging|extensions|fhe_circuit|mlir|graph|optimization|representation|tracing|values|benchmarks|ci|scripts|compilation|execution|deps)\))?(\!)?\:) .+$'
+          pattern: '^((feat|fix|chore|docs)\:) .+$'
```
fd0r, line 428: Not a big fan of this tbh.
I agree with removing the check of what is in parantheses though

fd0r (changes requested):
I'd rather keep feat|fix|chore|refactor|style|test|docs for now even if we don't use it that often.

Btw we should probably discuss with other teams on standardization across projects.

.github/workflows/continuous-integration.yaml
RomanBredehoft: out of curiosity, why do you think that ? 

I'm not sure why we decided on these types specifically as the link we give in our contributing section leads to : 
```
[
  'build',
  'chore',
  'ci',
  'docs',
  'feat',
  'fix',
  'perf',
  'refactor',
  'revert',
  'style',
  'test'
];
```
which has even more types that what we've defined

fd0r: Well I would be fine with the full list too 🤷🏼 
But removing some of the one we have now feels a bit restrictive imho.

RomanBredehoft: ok let's conform then !

fd0r (approved):

PR #477, opened 2024-01-30
chore: custom cp for cifar bench

rudy-6-4:
This helps to launch the bench using an alternative wheel from concrete CI and alternative concrete python branch

### DIFF
--- a/.github/workflows/cifar_benchmark.yaml
+++ b/.github/workflows/cifar_benchmark.yaml
@@ -7,6 +7,16 @@ on:
         default: "main"
         required: true
         type: string
+      alternative-cp-wheel-artifact-id:
+        description: Alternative Concrete-Python Wheel Artifact-ID (see https://github.com/zama-ai/concrete/actions/workflows/concrete_python_release.yml)
+        default: "none"
+        required: true
+        type: string
+      alternative-cp-branch:
+        description: Alternative Concrete-Python Branch
+        default: "none"
+        required: true
+        type: string
       benchmark:
         description: Benchmark to run (cifar-10-8b or cifar-10-16b)
         default: "cifar-10-16b"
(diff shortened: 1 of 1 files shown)


fd0r:
Nice to have but I would verify that we log some hash of Concrete Python to be able to identify which version of CP was used.

IIRC we currently log the git hash of the repo but this change means that the version used won't be the one in `poetry.lock` thus not tracked in the git hash.

RomanBredehoft:
also could you rename your commit by "feat: custom cp for cifar branch" ? since you've opened your PR we've removed the scopes in commit messages and I worry that it will make the CI fail when merging into main ! thanks

fd0r:
Actually @RomanBredehoft this is more a `chore` than `feat` imo since this doesn't add anything to the end-user.

fd0r (changes requested):
Two comments/questions:
- IMO this isn't a `feat` but a `chore`
- What would be the result of `    from importlib.metadata import version; version("concrete-python")` ?

RomanBredehoft:
yes you're right @fd0r I did not realize that

rudy-6-4:
> Two comments/questions:
> 
>     * IMO this isn't a `feat` but a `chore`
> 
>     * What would be the result of `from importlib.metadata import version; version("concrete-python")` with a custom build? We use it to track the improvements in the benchmark database so I'm afraid it will mess things up.
>       
>       * If we have some hash that identifies the version using `version("concrete-python")` then that's fine with me
>       * Otherwise we should find a way to detect it or just skip the upload to the DB step if we are using a custom version of Concrete

I disabled the upload for such cases as it could be arbitrary wip branch & wheel that are not easy to identify.

fd0r (approved):

RomanBredehoft (approved):

PR #484, opened 2024-02-05
feat: support conv1d operator

RomanBredehoft:
Adds a workaround in Concrete ML for supporting torch's `conv1d` using `conv2d` + some `unsqueeze`/`squeeze` ops, until it is handled in Concrete Python (https://github.com/zama-ai/concrete-ml-internal/issues/4117) 

closes https://github.com/zama-ai/concrete-ml-internal/issues/4109


fd0r (approved):
Awesome thanks!

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -911,14 +930,22 @@ def q_impl(
         else:
             numpy_q_out = conv_wx
 
+        # Workaround for handling torch's Conv1d operator until it is supported by Concrete Python
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4117
+        if is_conv1d:
+            numpy_q_out = numpy.squeeze(numpy_q_out, axis=-2)
```
jfrery, line 936: Is this axis=-2 going to work in all cases?

jfrery (approved):
Nice feature! Thanks

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: what do you have in mind by "all cases" ? What I can say about this is that : 
- we check that the kernel shape is either 1d or 2d, and we act accordingly to this (ie we quickly raise an error if inputs are n-d with n >= 3)
- for 1d conv, torch's convention is that inputs is (N, C, L) (https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html) while it's (N, C, H, W) for conv2d (https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html). This `-2` makes the 1d input of (N, C, L) go to (N, C, 1, L), but since we modify the kernel's shape (+ strides / pads / dilation) accordingly (ie, we put 1 in it's height dimension), I think all should be good. Haven't tried but putting `-1` and setting the kernel shape to `(kernel_shape[0], 1)` is probably equivalent. Do you want me to add a comment in the code ? 
- 
@jfrery

andrei-stoian-zama (approved):
Really great work here!

src/concrete/ml/quantization/quantized_ops.py
jfrery: All good. I was just checking if there was no obvious cases with time series where this wouldn't work but I guess we are fine.

PR #486, opened 2024-02-06
docs: update operator list in torch support's documentation section

RomanBredehoft:
A user pointed that that this list was not up to date. Turns out we indeed added/changed quite some things since we've updated this section.

closes https://github.com/zama-ai/concrete-ml-internal/issues/4268

### DIFF
--- a/docs/deep-learning/torch_support.md
+++ b/docs/deep-learning/torch_support.md
@@ -155,62 +155,93 @@ Concrete ML supports a variety of PyTorch operators that can be used to build fu
 
 ### Operators
 
-#### univariate operators
+#### Univariate operators
 
-- [`torch.abs`](https://pytorch.org/docs/stable/generated/torch.abs.html)
+- [`torch.nn.identity`](https://pytorch.org/docs/stable/generated/torch.nn.Identity.html)
 - [`torch.clip`](https://pytorch.org/docs/stable/generated/torch.clip.html)
+- [`torch.clamp`](https://pytorch.org/docs/stable/generated/torch.clamp.html)
+- [`torch.round`](https://pytorch.org/docs/stable/generated/torch.round.html)
+- [`torch.floor`](https://pytorch.org/docs/stable/generated/torch.floor.html)
+- [`torch.min`](https://pytorch.org/docs/stable/generated/torch.min.html)
+- [`torch.max`](https://pytorch.org/docs/stable/generated/torch.max.html)
+- [`torch.abs`](https://pytorch.org/docs/stable/generated/torch.abs.html)
+- [`torch.neg`](https://pytorch.org/docs/stable/generated/torch.neg.html)
+- [`torch.sign`](https://pytorch.org/docs/stable/generated/torch.sign.html)
+- [`torch.logical_or, torch.Tensor operator ||`](https://pytorch.org/docs/stable/generated/torch.logical_or.html)
+- [`torch.logical_not`](https://pytorch.org/docs/stable/generated/torch.logical_not.html)
+- [`torch.gt, torch.greater`](https://pytorch.org/docs/stable/generated/torch.gt.html)
+- [`torch.ge, torch.greater_equal`](https://pytorch.org/docs/stable/generated/torch.ge.html)
+- [`torch.lt, torch.less`](https://pytorch.org/docs/stable/generated/torch.lt.html)
+- [`torch.le, torch.less_equal`](https://pytorch.org/docs/stable/generated/torch.le.html)
+- [`torch.eq`](https://pytorch.org/docs/stable/generated/torch.eq.html)
+- [`torch.where`](https://pytorch.org/docs/stable/generated/torch.where.html)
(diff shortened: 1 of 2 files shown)


docs/deep-learning/torch_support.md
```diff
@@ -155,62 +155,93 @@ Concrete ML supports a variety of PyTorch operators that can be used to build fu
...
-- [`torch.gt`](https://pytorch.org/docs/stable/generated/torch.gt.html)
-- [`torch.clamp`](https://pytorch.org/docs/stable/generated/torch.clamp.html)
+- [`torch.pow`](https://pytorch.org/docs/stable/generated/torch.pow.html)
+- [`torch.sum`](https://pytorch.org/docs/stable/generated/torch.sum.html)
 - [`torch.mul, torch.Tensor operator *`](https://pytorch.org/docs/stable/generated/torch.mul.html)
 - [`torch.div, torch.Tensor operator /`](https://pytorch.org/docs/stable/generated/torch.div.html)
-- [`torch.nn.identity`](https://pytorch.org/docs/stable/generated/torch.nn.Identity.html)
 - [`torch.nn.BatchNorm2d`](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm2d.html)
+- [`torch.nn.BatchNorm3d`](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm3d.html)
+- [`torch.erf, torch.special.erf`](https://pytorch.org/docs/stable/special.html#torch.special.erf)
```
RomanBredehoft, line 186: I have a doubt in putting these two here (erf + pad), but I think it's ok

docs/deep-learning/torch_support.md
```diff
@@ -155,62 +155,93 @@ Concrete ML supports a variety of PyTorch operators that can be used to build fu
...
+- [`torch.abs`](https://pytorch.org/docs/stable/generated/torch.abs.html)
+- [`torch.neg`](https://pytorch.org/docs/stable/generated/torch.neg.html)
+- [`torch.sign`](https://pytorch.org/docs/stable/generated/torch.sign.html)
+- [`torch.logical_or, torch.Tensor operator ||`](https://pytorch.org/docs/stable/generated/torch.logical_or.html)
+- [`torch.logical_not`](https://pytorch.org/docs/stable/generated/torch.logical_not.html)
+- [`torch.gt, torch.greater`](https://pytorch.org/docs/stable/generated/torch.gt.html)
+- [`torch.ge, torch.greater_equal`](https://pytorch.org/docs/stable/generated/torch.ge.html)
+- [`torch.lt, torch.less`](https://pytorch.org/docs/stable/generated/torch.lt.html)
+- [`torch.le, torch.less_equal`](https://pytorch.org/docs/stable/generated/torch.le.html)
+- [`torch.eq`](https://pytorch.org/docs/stable/generated/torch.eq.html)
```
RomanBredehoft, line 176: side note: torch.eq is not the same as torch.equal

docs/deep-learning/torch_support.md
```diff
@@ -155,62 +155,93 @@ Concrete ML supports a variety of PyTorch operators that can be used to build fu
...
+- [`torch.conv1d`, `torch.nn.Conv1D`](https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html)
+- [`torch.conv2d`, `torch.nn.Conv2D`](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html)
+- [`torch.nn.AvgPool2d`](https://pytorch.org/docs/stable/generated/torch.nn.AvgPool2d.html)
+- [`torch.nn.MaxPool2d`](https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html)
 
-Concrete ML supports these operators but also the QAT equivalents from Brevitas.
+Concrete ML also supports some of their QAT equivalents from Brevitas.
 
 - `brevitas.nn.QuantLinear`
+- `brevitas.nn.QuantConv1d`
```
RomanBredehoft, line 216: I have a doubt here : is supporting `torch.conv1d` enough for supporting `brevitas.nn.QuantConv1d` ?

fd0r: I think so yeah

fd0r (approved):

docs/deep-learning/torch_support.md
kcelia: I think we support all Brevitas operators that are equivalent to those already supported in Torch. 
So the first sentence seems more appropriate. 

Should we add: to this list: qnn.QuantLinear, qnn.QuantIdentity, qnn.QuantReLU

RomanBredehoft: `qnn.QuantLinear` is in this list and `qnn.QuantIdentity` is below in the "Quantizers" sub-section

but for other like activation functions, @andrei-stoian-zama pointed out to me that we should not suggest users to use them as : 
- it's less certain that we indeed do support them
- we most likely never need them after all, as activation functions don't require quantization (done with a TLU), we can just use `torch.relu` 

so I don't think there's much to add here !

fd0r (approved):

Issue #428, opened 2024-01-04
Unknown attributes: count_include_pad when compiling with compile_torch_model


NgesBrian:
## Summary
When develop a custom neural network and try to compile it using MNIST Dataset through compile_torch_model, I get an assertion error :
"AssertionError: Got the following unknown attributes: count_include_pad. Accepted attributes: ceil_mode, kernel_shape, pads, strides" 

I was expecting compilation and generation of the circuit which I will then use to train my model. 

- versions affected:
- python version: 3.10.12
- config : x86_64
- workaround (optional): remove the pooling layers but you are likely to run into other issues and above all, affect the accuracy of the network. 

Step by step procedure someone should follow to trigger the bug:

RomanBredehoft:
Hello @NgesBrian ,
Indeed it seems that our AvgPool2D operator is missing support for the `count_include_pad` parameter. We are planning on fixing this in the near future and we will let you know when this is done 😉 

Thanks for the report !

RomanBredehoft:
This is just to let you known that we have fixed the issue and it will be included in our upcoming release ! Besides, for now, we only support the `count_include_pad=1` behavior.

PR #489, opened 2024-02-06
fix: add support to AvgPool's missing parameters

RomanBredehoft:
`auto_pad` and `count_include_pad` parameters are described in [ONNX's documentation](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#AveragePool-11) and I believe : 
- that we cannot support `auto_pad` (ie, we force explicit padding)
- we force `count_include_pad` to be True (ie, we include padded values when calculating values for the edges)

tell me what you think about it. I also took the liberty to add checks on our input asserts, as we already do for the conv operator 

closes https://github.com/zama-ai/concrete-ml/issues/428
closes https://github.com/zama-ai/concrete-ml-internal/issues/4203

### DIFF
--- a/src/concrete/ml/onnx/ops_impl.py
+++ b/src/concrete/ml/onnx/ops_impl.py
@@ -1337,10 +1337,12 @@ def numpy_conv(
 def numpy_avgpool(
     x: numpy.ndarray,
     *,
-    ceil_mode: int,
     kernel_shape: Tuple[int, ...],
-    pads: Tuple[int, ...] = None,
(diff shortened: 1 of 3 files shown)


andrei-stoian-zama (approved):

fd0r (approved):

jfrery (approved):
Thanks. Would there be a way to catch if we miss some parameter vs onnx?

RomanBredehoft:
made an issue for it @jfrery , not sure it's trivial

PR #492, opened 2024-02-08
chore: update Concrete Python to 2.5.1

fd0r:


RomanBredehoft (approved):
looks good ! I'm just surprised about these "float, int, bool" changes, why would mypy detect that, I thought we always returned arrays .. but maybe some shape operators like flatten don't 🤔

RomanBredehoft (dismissed):

andrei-stoian-zama (approved):

Makefile
```diff
@@ -19,7 +19,7 @@ OPEN_PR="true"
 # Force the installation of a Concrete Python version, which is very useful with nightly versions
 # /!\ WARNING /!\: This version should NEVER be a wildcard as it might create some
 # issues when trying to run it in the future.
-CONCRETE_PYTHON_VERSION="concrete-python==2.5.0"
+CONCRETE_PYTHON_VERSION="concrete-python==2.5.1"
```
kcelia, line 22: You did this change in another PR: https://github.com/zama-ai/concrete-ml/pull/493/files

fd0r: Yes the other PR targets the release branch and this one main, otherwise the PRs are the same

PR #493, opened 2024-02-08
chore: update Concrete Python to 2.5.1

fd0r:


andrei-stoian-zama (approved):

src/concrete/ml/sklearn/qnn.py
```diff
@@ -141,7 +141,8 @@ def __init__(
         virtual_params = kwargs.pop("virtual_params_", {})
 
         self._kwargs = kwargs
-        vars(self).update(kwargs)
```
kcelia, line 144: Out of curiosity, why did you change it ? 
It's the same, and the first one is more concise.

fd0r: I had to update mypy that wasn't pleased with this since `vars` doesn't return a `dict` but a `mapping` that in the spec doesn't have an `update` method.

PR #497, opened 2024-02-08
chore: improve flaky rerun step in CI (pytest crash, appended coverage)

RomanBredehoft:
This PR fixes two small issues related to https://github.com/zama-ai/concrete-ml/actions/runs/7817775612/job/21326578294 : 
- when rerunning flaky tests, since we now append the coverage, we also need to make sure that it still fails under 100% coverage + avoids printing covered lines for all files
- in the specific case where pytest crashes (syntax error in a test file), make sure we do no falsely detect a flaky test and trigger the rerun step (this was triggered if pytest failed + no non-flaky tests was found, but we now also make sure we do have at least a flaky test)


kcelia (approved):
Thanks

PR #498, opened 2024-02-08
chore: update tensorflow to avoid invalid escape sequence error

RomanBredehoft:
Following https://github.com/zama-ai/concrete-ml-internal/issues/4217, turns out Keras also raised an `invalid escape sequence` error that I've missed in the CI (https://github.com/zama-ai/concrete-ml/actions/runs/7817775612/job/21326578294). I'm unfortunately not able to reproduce it locally. 

But after digging in it a bit, I realized that the escape character was removed in keras/tensorflow 2.13 , so hopefully this fixes the issue. If it does not, will go for https://github.com/zama-ai/concrete-ml/pull/496 and investigate further


jfrery (approved):
Thanks!

fd0r (approved):

PR #511, opened 2024-02-14
fix: make skorch import fail without error

fd0r:
Skorch is not available in the step of the release that launches the import of skorch.
This makes the import fail without error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,7 @@ setup_env:
 	echo "Installing $(CONCRETE_PYTHON_VERSION)" && \
 	poetry run python -m pip install -U --pre "$(CONCRETE_PYTHON_VERSION)"
 	"$(MAKE)" fix_omp_issues_for_intel_mac
-
-	poetry run python -c "import skorch" # Details above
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -57,7 +57,7 @@ setup_env:
 	poetry run python -m pip install -U --pre "$(CONCRETE_PYTHON_VERSION)"
 	"$(MAKE)" fix_omp_issues_for_intel_mac
 
-	poetry run python -c "import skorch" # Details above
+  poetry run python -c "import skorch" || true # Details above
```
RomanBredehoft, line 60: not well indented

RomanBredehoft (changes requested):
the new line is not well indented 

also, although this would probably fix the issue, I believe the initial error is different (problem with poetry + the clone install script), so this should be temporary (not worth a fixme I'll try to fix it today)

RomanBredehoft (approved):

RomanBredehoft (approved):

kcelia (approved):

PR #512, opened 2024-02-14
fix: make skorch import fail without error

fd0r:
Skorch is not available in the step of the release that launches the import of skorch.
This makes the import fail without error.


RomanBredehoft (approved):

RomanBredehoft:
looks like you got a flaky test, but it was not re-run because in this case the worker crashes (and honestly I'm not sure we can easily handle this case for re-run)

PR #515, opened 2024-02-14
chore: fix link to license file in the readme

RomanBredehoft:
The link was always wrong (ex: https://github.com/zama-ai/concrete-ml/actions/runs/7887169029/job/21521900852), but it looks like in some systems like macos, pathlib's `exists()` function is not case sensitive 🙂


fd0r (approved):

kcelia (dismissed):
Thanks

jfrery (approved):

PR #517, opened 2024-02-15
Redo the test seeding system

bcm-at-zama:
The seeding system is completely changed and simplified: now, you launch a test with `pytest`, which is going to use a randomly seed (either set by you or picked randomly), and this randomly seed is all what you need: the seeds used for all tests are going to be derived from this master seed, whatever the order of the tests. Such that, if there is a crash, you don't need a print of some subseed to reproduce.

More precisely: the used seed for a test of a function `f_i` (eg, test_compute_bits_precision) on a configuration `c_j` (eg, x0-8) of a test file `t_k` (eg, tests/common/test_utils.py) is computed as some `hash(f_i, c_j, t_k, randomly-seed)`

Or, in other words, if for some randomly-seed, a test crash for `(f_i, c_j, t_k)`, you can reproduce the same seed by directly calling: `python -m pytest t_k::f_i[c_j] --randomly-seed=XXX`, eg `python -m pytest -xsvv "tests/common/test_utils.py::test_compute_bits_precision[x0-8]" --randomly-seed=1508248188`.


bcm-at-zama:
I did that for our issue in the CI. However, I was not able to reproduce this exact issue, so I wouldn't be able if it perfectly works. However, I'm pretty confident it's the real seed (and it will be nice in the future to try the system and check it).

bcm-at-zama:
Hum, the fact that it says 

--forcing_random_seed 12679042460344433083 --randomly-dont-reset-seed

too much is not a good sign, maybe I should have another look.

bcm-at-zama:
It will never work like this, random's are picked in the test so we don't simulate this by aborting at the beginning. We would need a major change in the seeding system to make it work as expected, let me see

conftest.py
```diff
@@ -49,15 +50,6 @@ def pytest_addoption(parser):
         help="To dump pytest-cov term report to a text file.",
     )
 
-    parser.addoption(
-        "--forcing_random_seed",
```
bcm-at-zama, line 53: --forcing_random_seed is no more needed! the master seed is the randomly seed, which is unique and always printed at the beginning

conftest.py
```diff
@@ -176,37 +168,61 @@ def function_to_seed_torch(seed):
...
+    assert (
+        absolute_path.count("concrete-ml/test") == 1
+    ), f"the absolute path is not as expected {absolute_path=}"
+
+    relative_file_path = absolute_path[
+        absolute_path.find("concrete-ml/test") + len("concrete-ml/") :
+    ]
+
+    # Derive the sub_seed from the randomly_seed and the test name
+    derivation_string = f"{relative_file_path} # {str(request.node.name)} # {randomly_seed}"
```
bcm-at-zama, line 191: this is the derivation of the subseed: for a test of a function f_i (eg, test_compute_bits_precision) on a configuration c_j (eg, x0-8) of a test file t_k (eg, tests/common/test_utils.py) is computed as some hash(f_i, c_j, t_k, randomly-seed)

conftest.py
```diff
@@ -176,37 +168,61 @@ def function_to_seed_torch(seed):
...
+    seed = sub_seed
 
-    # Python
-    random.seed(seed)
-    print("\nForcing seed to random.seed to ", seed)
     print(
-        f"\nRelaunch the tests with --forcing_random_seed {seed} "
-        + "--randomly-dont-reset-seed to reproduce. Remark that adding --randomly-seed=... "
-        + "is needed when the testcase uses randoms in pytest parameters"
+        f"\nRelaunch the tests with --randomly_seed {randomly_seed} "
```
bcm-at-zama, line 206: repro directly a subtest by:
- forcing --randomly_seed
- setting the subtest you want
and that's it

script/make_utils/check_pytest_determinism.sh
```diff
@@ -40,32 +40,25 @@ set -e
...
 echo "Successful determinism check"
 
-# Now, check --forcing_random_seed, i.e., check that one can reproduce conditions of a bug in a single file
-# and test without having to relaunch the full pytest, by just picking the right --forcing_random_seed
+# Now, check that one can reproduce conditions of a bug in a single file
+# and test without having to relaunch the full pytest
 
 # All lines that start with "tests/seeding/test_seeding.py::" represent a test that passed
 LIST_FILES=$(grep "tests/seeding/test_seeding.py::" "${OUTPUT_DIRECTORY}/one.txt")
-LIST_SEED=()
```
bcm-at-zama, line 48: no more need to find --forcing_random_seed in the log files! it was especially boring since sometimes, they were not there if a crash occurred.

conftest.py
```diff
@@ -176,37 +168,61 @@ def function_to_seed_torch(seed):
...
-def autoseeding_of_everything(record_property, request):
+def autoseeding_of_everything(request):
     """Seed everything we can, for determinism."""
-    main_seed = request.config.getoption("--forcing_random_seed", default=None)
+    randomly_seed = request.config.getoption("--randomly-seed", default=None)
 
-    if main_seed is None:
-        main_seed = random.randint(0, 2**64 - 1)
+    if randomly_seed is None:
+        raise ValueError("why isn't --randomly-seed set?")
```
RomanBredehoft, line 176: would prefer a message like `"Please set --randomly-seed set with an integer"`

conftest.py
```diff
@@ -176,37 +168,61 @@ def function_to_seed_torch(seed):
...
 
-    seed = main_seed
-    record_property("main seed", main_seed)
+    # We need to find the relative file path of the test file. It does not look native with request,
+    # so we recompute it.
+    absolute_path = str(request.fspath)
+
+    assert (
+        absolute_path.count("concrete-ml/test") == 1
```
RomanBredehoft, line 183: not sure what this is, and the associated error remains pretty vague !

conftest.py
```diff
@@ -176,37 +168,61 @@ def function_to_seed_torch(seed):
...
+    ), f"the absolute path is not as expected {absolute_path=}"
+
+    relative_file_path = absolute_path[
+        absolute_path.find("concrete-ml/test") + len("concrete-ml/") :
```
RomanBredehoft, line 187: not sure what this does either

conftest.py
bcm-at-zama: But they don't have to set this themselves, it's done by the system. In other words, it's normally impossible to raise this error

bcm-at-zama: I can change the sentence if you want, but not to what you mentioned because it's misleading

conftest.py
RomanBredehoft: why not indeed ! is `request.node.name` the actual test + config ? like `tests/common/test_utils.py::test_compute_bits_precision[True-False-5]` ?

conftest.py
bcm-at-zama: I have an absolute path, eg `/Users/benoitchevalliermames/Documents/Zama/Git/concrete-ml/tests/common/test_utils.py`
I want to compute `tests/common/test_utils.py` from it
It will work, as soon as the guy making the cloning didn't set a crazy path like

`/something/concrete-ml/tests/something-else/concrete-ml/tests/`

in which case I want to abort because it's not good.

conftest.py
```diff
@@ -176,37 +168,61 @@ def function_to_seed_torch(seed):
...
+
+    # Derive the sub_seed from the randomly_seed and the test name
+    derivation_string = f"{relative_file_path} # {str(request.node.name)} # {randomly_seed}"
+
+    hash_object = hashlib.sha256()
+    hash_object.update(b"{derivation_string}")
+    hash_object.digest()
+    hash_value = hash_object.hexdigest()
+
+    sub_seed = int(hash_value, 16) % 2**64
```
RomanBredehoft, line 198: why is this needed ?

conftest.py
bcm-at-zama: If you really want, I can make a change to find the _last_ occurence of `concrete-ml/tests`, instead of the first one. But except if the guys want to bother us, never the assert will fail

conftest.py
bcm-at-zama: explained in

+    # We need to find the relative file path of the test file. It does not look native with request,
+    # so we recompute it.

two lines before. Not sufficient, or a better way to do it?

conftest.py
RomanBredehoft: what do you mean by "setting the subtest you want" ? isn't running `pytest tests/common/test_utils.py::test_compute_bits_precision[True-False-5] --randomly_seed 78975873 --randomly-dont-reset-seed` enough for example ?

conftest.py
bcm-at-zama: why _what_ is needed? which part?

conftest.py
```diff
@@ -176,37 +168,61 @@ def function_to_seed_torch(seed):
...
+        f"\nRelaunch the tests with --randomly_seed {randomly_seed} "
+        + "--randomly-dont-reset-seed to reproduce."
     )
     print(
         "Remark that potentially, any option used in the pytest call may have an impact so in "
         + "case of problem to reproduce, you may want to have a look to `make pytest` options"
     )
 
+    # Python
+    random.seed(seed)
```
RomanBredehoft, line 215: are you sure numpy does not already do that ?

deps_licenses/licenses_mac_intel_user.txt.md5
```diff
@@ -1 +1 @@
-f709427468f5be4ee6836603495fea72
+a923947bfb17b658ab8efe61d5cafe96
```
RomanBredehoft, line 1: you should run the licence update action and target your branch to also update linux + M1 licenses : https://github.com/zama-ai/concrete-ml/actions/workflows/update_licenses.yaml

conftest.py
bcm-at-zama: if we have two files t_k in two different directories with the same filename: it would give the same hash and so subseed. I wanted to keep the "all the tests have different seed" properties, so I included the path as well. But I don't want absolute path, such that, whatever our config (eg, mine or yours), the seeds are the same. In other words, if you clone CML in a dir R and I do it in a dir B, our seeds are the same

conftest.py
bcm-at-zama: Now with the changes, yes `pytest tests/common/test_utils.py::test_compute_bits_precision[True-False-5] --randomly_seed 78975873 --randomly-dont-reset-see` is enough but that's exactly the change. Before, it was NOT sufficient, we have to use `--forcing_random_seed ` option

RomanBredehoft (changes requested):
thanks for that ! I made a few comments ! Also : 
- I would add a few comments explaining a bit more what you are doing (notably where I asked some questions)
- could you also update the licenses using the dedicated action ? 
- is there anything to change in the makefile as well ?

conftest.py
bcm-at-zama: Tried to re-explain what it does in this thread

conftest.py
bcm-at-zama: - it's not new, it was already there
- and what's the problem with seeding: I want to be sure of the conditions in the `(f_i, c_j, t_k)` test, so I force the seeds here

deps_licenses/licenses_mac_intel_user.txt.md5
bcm-at-zama: the deps have not changed, I was said by @fd0r there is a racing problem

conftest.py
RomanBredehoft: hmm, so you're saying running `pytest ... --randomly-seed` with nothing after won't trigger this issue ? 

if so ok but yes maybe put it more like `--randomly-seed has not been set properly` then

bcm-at-zama: no it fails, but it's already caught by the system, much before, directly in randomly-seed or pytest:

```
python -m pytest -xsvv "tests/common/test_utils.py::test_compute_bits_precision[x0-8]" --randomly-seed     
/Users/benoitchevalliermames/Documents/Zama/Git/concrete-ml/.venv/lib/python3.9/site-packages/_pytest/config/__init__.py:330: PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown.
Plugin: helpconfig, Hook: pytest_cmdline_parse
UsageError: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: argument --randomly-seed: expected one argument
For more information see https://pluggy.readthedocs.io/en/stable/api_reference.html#pluggy.PluggyTeardownRaisedWarning
  config = pluginmanager.hook.pytest_cmdline_parse(
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: argument --randomly-seed: expected one argument
```

conftest.py
bcm-at-zama: ok

conftest.py
RomanBredehoft: ah yes I see, ok looks fine actually, I was first suspicious of this `len("concrete-ml/")` but I get it now, it's like doing some kind of regex manually

conftest.py
RomanBredehoft: yes yes I got that, it's just taht your above comment says it computes  " hash(f_i, c_j, t_k, randomly-seed)" but in the code I only see 3 variables

but now I guess  `str(request.node.name)` is actually the `f_i, c_j,` part. `relative_file_path` being `t_k`

conftest.py
bcm-at-zama: yes, it's possible you guys have a cleaner way to do that

conftest.py
bcm-at-zama: done, ok to resolve?

conftest.py
```diff
@@ -176,37 +168,77 @@ def function_to_seed_torch(seed):
...
 
 @pytest.fixture(autouse=True)
-def autoseeding_of_everything(record_property, request):
+def autoseeding_of_everything(request):
     """Seed everything we can, for determinism."""
-    main_seed = request.config.getoption("--forcing_random_seed", default=None)
 
-    if main_seed is None:
-        main_seed = random.randint(0, 2**64 - 1)
+    # Explanations on the seeding system:
```
bcm-at-zama, line 174: new comments @RomanBredehoft

conftest.py
RomanBredehoft: well, the line I comment on : `    sub_seed = int(hash_value, 16) % 2**64`

is it because the has value (in hexa) can go above 2**64 b ? I would jut add a comment around this part (and the above lines) to add a simple `# Get the subseed by getting the hash value and retrieving an int64 value`

also : 
- for seeding I thought numpy prefered 32b values (in the code we often use % 2**31 for seed iirc)
- I guess the risk of collisiions because of this %64 are minimal, right  ?

conftest.py
RomanBredehoft: okok got it you were commenting the actual workflow, not what users was supposed to do my bad !

conftest.py
bcm-at-zama: yes: eg, if you launch `python -m pytest -xsvv "tests/common/test_utils.py::test_compute_bits_precision[x0-8]"`, you'll see

```
Using randomly_seed=2216371280
Using derivation_string='tests/common/test_utils.py # test_compute_bits_precision[x0-8] # 2216371280'
Using sub_seed=16202174809221803539
```

so everything is here: the relative test file, the config, the function and the randomly seed.

conftest.py
RomanBredehoft: okok if it's already there all is good

deps_licenses/licenses_mac_intel_user.txt.md5
RomanBredehoft: aah ok my bad yes it's because mac intel licenses were not updated last time got it !

conftest.py
bcm-at-zama: - the hash is a SHA256, so it's 256b. And random.seed wants a 64b seed and numpy.random.seed wants a 32 seed; so I reduce to 64b and I have `numpy.random.seed(seed % 2**32)`
- using % 2**31 reduces to 31b, there is one bit loss here
- we have collisions on 64b in O(2**32) but it's not a problem here [what's the attack] and it's not because of us [it's python and numpy which want reduced seeds]

conftest.py
bcm-at-zama: yes, it just moved by few lines

deps_licenses/licenses_mac_intel_user.txt.md5
bcm-at-zama: not sure, I'll let you fix that guys, it's not related to this PR and apparently you know the problem, CC @fd0r

conftest.py
RomanBredehoft: yeah got it thanks !  well I would just add a little line to simply explain that, nothing fancy

conftest.py
RomanBredehoft: got it, so yeah maybe just update the message and that should be enough !

conftest.py
RomanBredehoft: yes thanks !

conftest.py
bcm-at-zama: Adding a line.

conftest.py
RomanBredehoft: yes great thanks  !

conftest.py
RomanBredehoft: I was more referring to my second suggestion : `--randomly-seed has not been set properly` as clearly it's not a user problem !

conftest.py
RomanBredehoft: ok great !

conftest.py
RomanBredehoft: great !

conftest.py
bcm-at-zama: but the user doesn't have to set --randomly-seed
to me, it's more an internal issue, like "I arrive in an impossible situation", we should have an issue like that

RomanBredehoft: yes I know, if you prefer maybe `--randomly-seed has not been properly configured internally.`

bcm-at-zama: if you want: done

RomanBredehoft (dismissed):
thanks a lot for this !

bcm-at-zama:
Fixing conformance issues.

RomanBredehoft (approved):

kcelia (approved):
Thanks

RomanBredehoft:
are merging this @bcm-at-zama ?

bcm-at-zama:
Yes please, I've let you manage this guys!

PR #519, opened 2024-02-16
feat: let's have compressed keys as an option [WIP]

bcm-at-zama:
[WIP for now]

let's check how it behaves, is there issues sometimes?


bcm-at-zama:
eg, `python -m pytest -xsvv "tests/sklearn/test_sklearn_models.py::test_separated_inference[6-fhe-RandomForestRegressor]" --forcing_random_seed 4939467286315543969 --randomly-dont-reset-seed` fails

BourgerieQuentin:
please retry experiments with coming nightly (fixed in https://github.com/zama-ai/concrete/pull/691)

bcm-at-zama:
Rebasing + using the new CP (which comes tomorrow, and has a fix for key compression)

bcm-at-zama:
Trying to run https://github.com/zama-ai/concrete-ml/actions/workflows/update_licenses.yaml workflow even if Mac distribution is not there: I wonder if we need Mac for regular builds, I would say No

bcm-at-zama:
The action fails because it tries to install CP for Mac

bcm-at-zama:
Tests are green now!

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -632,6 +632,7 @@ def compile(
         global_p_error: Optional[float] = None,
         verbose: bool = False,
         inputs_encryption_status: Optional[Sequence[str]] = None,
+        compress_evaluation_keys: bool = True,
```
andrei-stoian-zama, line 635: I would rather we set this through the configuration and keep only ML related params here

andrei-stoian-zama (changes requested):
- I think we should only set "compress_keys" through the Configuration if that's possible
- We should find out if there is any runtime latency difference with / without compression
- We should set the proper default based on the previous question

src/concrete/ml/quantization/quantized_module.py
bcm-at-zama: then, would be awesome to have a single place / class for crypto-related configs, and we would change just one place and it would apply to all models we support (linear, tree, qnn, torch). Maybe already the case!

andrei-stoian-zama:
Superseded by #726

PR #530, opened 2024-03-05
chore: set new simulation as default

melanciani:
to check if tests are still much slower with the new simulation


RomanBredehoft:
this was made for testing it, closing it for now as it's still too slow to integrate

PR #536, opened 2024-03-11
feat: add encrypted data-frame API

RomanBredehoft:
Adding support for encrypted data-frame through Pandas-like API 

refs https://github.com/zama-ai/concrete-ml-internal/issues/4342
closes https://github.com/zama-ai/concrete-ml-internal/issues/4211
closes https://github.com/zama-ai/concrete-ml-internal/issues/4276
closes https://github.com/zama-ai/concrete-ml-internal/issues/4247


RomanBredehoft:
I've taken most of your comments into account @andrei-stoian-zama , but some things are still remaining / to be discussed 

- minor things : 
    - I mistakenly made the API pass the evaluation key in the `merge` / `join` operator, instead of in `encrypt_from_pandas`, I'll fix that 
    - use a real data set instead of a synthetic custom one
    - add tests
    - clean and pass make pcc
- medium things : 
    - add quant / de-quant : not sure `quant_cols` is useful actually, couldn't we just : 
        - always quantize floats 
        - quantize integers if they are out of bound (lower than min / greater than max allowed) ?
    - add string mapping
    - add dtype mapping (so that we don't always have floats in the final DF: we can store dtypes in pre-processing and cast in post-processing)
- major things : 
    - I'm currently not sure how we can make the client "operator-agnostic" in anticipation to strong composition (as I'm not sure how it will work)
        - I'm expecting that we'll have several circuits to handle, one per supported operator
        - issue: for encryption, we need to provide what "input position" to use, but that requires to know which circuit operator to use right ? (ie: for left/right join, the circuit expects 4 inputs, but only the first or second one are relevant when encrypting the input users and need to run: `enc_input, _, _, _ = client.encrypt(input, None, None, None)`)
         - or maybe we'll be able to use any circuit anyway as they will be compiled in a way that any operator can be used as the first operator ? in that case we could randomly/arbitrarily chose a circuit for encryption
    - what do we do for circuits ?
        - do we store the supported circuits in CML's source (using client/server zip files, like done currently) ?
        - do we ask the dev to run its own compile (and then save the client/server zip files on its own) ? 
        - do we make the user compile the circuit each time he generates the keys for the first time ?
    - do we set 4 bits as a global constant for now (like done in the PR) and remove it in q2 with radix ? 
        - it's related to the items right above (ie: if we allow other n_bits, should CML compile all possible circuits or are users/devs going to do it ?)
- things I've done differently from your suggestion : 
    - `generate_keys` does not exit, as we don't use keys to encrypt / run the fhe exec, instead client/server instances from CP (in particular, there's no way to retrieve the private keys in the API). In the notebook, the equivalent is therefore `get_client_and_eval_keys` and `load_server` (name to be confirmed)
    - I did not expose `n_bits` and `quant_cols` in `encrypt_from_pandas` because it didn't feel relevant (see medium/major items)
    - `serialize` method is named `to_json`  (could be `dump` as well, unsure)
    - `load_encrypted` is named `load_encrypted_dataframe` 
    - I added `encrypt_from_csv` func and `decrypt_to_csv` method , we can remove if not necessary

use_case_examples/dataframe/encrypted_pandas.ipynb
```diff
@@ -0,0 +1,321 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "client_1, evaluation_keys_1 = get_client_and_eval_keys(OPERATOR, keys_path=KEYS_DIR / \"keys\")\n",
+    "client_2, evaluation_keys_2 = get_client_and_eval_keys(OPERATOR, keys_path=KEYS_DIR / \"keys\")"
```
jfrery, line 120: Maybe a comment about what's going on here. Currently client share the key. But could be based on a public key + TD in the future.

src/concrete/ml/pandas/_operators.py
```diff
@@ -0,0 +1,241 @@
...
+    sort=False,
+    copy=None,
+    indicator=False,
+    validate="1:1",
+):
+    """Execute a left join using encrypted data-frames with Pandas kwargs.
+
+    For now, only a left and right join merge is supported. Additionally, default values for
+    parameter 'validate' is different than in Pandas.
+    """
```
jfrery, line 51: No params description in the docstring?

src/concrete/ml/pandas/_operators.py
```diff
@@ -0,0 +1,241 @@
...
+    on=None,
+    left_on=None,
+    right_on=None,
+    left_index=False,
+    right_index=False,
+    sort=False,
+    copy=None,
+    indicator=False,
+    validate="1:1",
+):
```
jfrery, line 46: What's the output data type? Pandas, Numpy? We can probably add some typing here. Same for the args.

src/concrete/ml/pandas/client_server.py
```diff
@@ -0,0 +1,122 @@
...
+
+def _get_encrypt_config():
+    return _PANDAS_OPS_TO_CIRCUIT_CONFIG["left_right_join"]["encrypt_config"]
+
+
+def _get_min_max_allowed():
+    return (1, _get_left_right_join_max_value(_N_BITS_PANDAS))
+
+
+def _save_client_server(config, force_save=True):
```
jfrery, line 92: This doesn't look like to be used or accessible by users?

jfrery (commented):
Nice implementation. Really like the API.

Some notes:
- I am a bit scared of having deployment files in our src. They are dependent lots of things (CML / CP version, python version, os?). Is this all handled? Can we find another way? Probably let the user generate them?
-  let's add typing where we can so that vscode isnt' lost at least.

src/concrete/ml/pandas/__init__.py
```diff
@@ -0,0 +1,32 @@
+from .client_server import get_client_and_eval_keys, load_client
```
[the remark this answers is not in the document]
RomanBredehoft: this is the __init__ so it's normal to have these

src/concrete/ml/pandas/client_server.py
RomanBredehoft: it's not supposed to be indeed

andrei-stoian-zama (approved):
Looks great!

andrei-stoian-zama (approved):
Great work!

jfrery (commented):
Very nice API! I have a very small comment on the notebook maybe:

```python
df_joined_enc_server = df_left_enc.merge(df_right_enc, how=HOW, on=ON)
```

HOW and ON should be a bit more explicit at this point rather than making reader go back all the way up.

PR #537, opened 2024-03-13
chore: fix seeding system path + typo

jfrery:


andrei-stoian-zama (dismissed):

bcm-at-zama:
What was the issue, here?

RomanBredehoft (dismissed):
thanks ! 

@bcm-at-zama in docker, there is not "concrete-ml folder, only src, tests, etc .. which made this fail !

RomanBredehoft (approved):

PR #538, opened 2024-03-14
chore: remove cifar with splitting

andrei-stoian-zama:
And additionally:
- change deployment use case to cifar 2b
- move perror search to cifar 2b

Closes https://github.com/zama-ai/concrete-ml-internal/issues/4314
Closes https://github.com/zama-ai/concrete-ml-internal/issues/3773


use_case_examples/cifar/cifar_brevitas_training/p_error_search.py
jfrery, line 1: Introducing this here probably requires some changes in the readme + have you run it? What is the final `p_error` found? It's likely not going to be 0.1

jfrery (changes requested):
Thanks for tackling this.

use_case_examples/cifar/cifar_brevitas_training/p_error_search.py
andrei-stoian-zama: No, I didn't run it, it's way too long.

We can run it later with the new improved simulation

jfrery (approved):
Looks good to me. Could you please create the related issue to reintroduce p_error search for cifar with new simulation?

PR #539, opened 2024-03-15
chore: update CP to nightly with approx rounding (24.3.14)

andrei-stoian-zama:


jfrery (dismissed):
Awesome! Thanks. Do you think we should build a script to do the steps you did here?

RomanBredehoft (approved):
still suspicious of commenting CP in toml though

bcm-at-zama:
agreed with @RomanBredehoft

fd0r:
> still suspicious of commenting CP in toml though

We always worked like that afaik

> Awesome! Thanks. Do you think we should build a script to do the steps you did here?

Maybe @andrei-stoian-zama can do some feedback on this but it's quite fast to do by hand and we don't really do this that much.

fd0r (approved):

bcm-at-zama:
>> still suspicious of commenting CP in toml though
> We always worked like that afaik

Well, not me: https://github.com/zama-ai/concrete-ml/pull/519/files :D

jfrery:
If the new simulation is added here, could you make this two different commits, one for CP update and one for the new simulation?

jfrery:
xgboost multi threading was interfering with CP simulation so:
- I set `n_jobs = 1`  by default in the xgboost models. Clearly not ideal for users. I am trying to see if we can do that in our tests easily.
- I commented out the low bit-width rounding. It's flaky with the new simulation. Will do a proper issue on this.
- We sill have the `Simulating TFHE failed` in some case see https://github.com/zama-ai/concrete-ml-internal/issues/4333

conftest.py
```diff
@@ -147,6 +147,25 @@ def default_configuration():
         enable_unsafe_features=True,
         use_insecure_key_cache=True,
         insecure_key_cache_location="ConcreteNumpyKeyCache",
+        fhe_simulation=False,  # Simulation compilation is done lazilly on circuit.simulate
```
kcelia, line 150: off-topic ; Will an error be raised if both fhe_simulation and fhe_execution flags are set to true?

src/concrete/ml/pytest/utils.py
```diff
@@ -388,6 +388,11 @@ def instantiate_model_generic(model_class, n_bits, **parameters):
         model_name (str): The type of the model as a string.
         model (object): The model instance.
     """
+    # Force multi threaded models to be single threaded as it is not working
```
kcelia, line 391: Maybe add an issue here

conftest.py
jfrery: > off-topic ; Will an error be raised if both fhe_simulation and fhe_execution flags are set to true?

No you can set both to true. By default, the simulation one is off but if you trigger circuit.simulate, it will compile for simulation.



> Maybe add a comment, saying that the default config is fhe_execution in this docstring function

Actually default configuration is both of them.

tests/torch/test_compile_torch.py
```diff
@@ -421,9 +421,11 @@ def accuracy_test_rounding(
...
         x_test, quantized_numpy_module_round_high_precision, simulate=simulate
     )
-    check_is_good_execution_for_cml_vs_circuit(
-        x_test, quantized_numpy_module_round_low_precision, simulate=simulate
-    )
+    # low bit-width rounding is not behaving as expected with new simulation
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4331
+    # check_is_good_execution_for_cml_vs_circuit(
+    #     x_test, quantized_numpy_module_round_low_precision, simulate=simulate
+    # )
```
kcelia, line 428: > It doesn't work as expected at low bit widths. 

Do you know the low bit precision ? If so, perhaps you could add it to the comment.

kcelia (approved):

src/concrete/ml/pytest/utils.py
jfrery: Not sure what kind of follow up issue we could make. xgboost,  pytest and new simulation are all trying to access all threads so we have to let one of them on all threads and set the rest single thread.

tests/torch/test_compile_torch.py
jfrery: Not sure what's the threshold. Here we use 2 bits.

conftest.py
RomanBredehoft: > Actually default configuration is both of them.

what do you mean ? in CP (and here) fhe simu is not enabled by default right ?

RomanBredehoft, line 150: we try to avoid in-line comment so better to move it above please !

src/concrete/ml/common/utils.py
```diff
@@ -38,7 +38,7 @@
 # when simulating FHE executions
 # Set 'USE_OLD_VL' to False by default once the new simulation is fixed
 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4091
```
RomanBredehoft, line 40: also time to remove the fixme + comment above !

tests/sklearn/test_sklearn_models.py
```diff
@@ -1710,6 +1710,10 @@ def test_p_error_simulation(
     # Check if model is linear
     is_linear_model = is_model_class_in_a_list(model_class, _get_sklearn_linear_models())
 
+    # Do not run the test for linear models since there is no PBS (i.e. p_error has no impact)
+    if is_linear_model:
```
RomanBredehoft, line 1714: you should take the time to simplify the test here : we have no more reason to test linear models, so instead of having this `is_linear_model` and skip the test, you should simply change the 
```
@pytest.mark.parametrize("model_class, parameters", MODELS_AND_DATASETS)
```
to 
```
@pytest.mark.parametrize(
    "model_class, parameters",
    get_sklearn_linear_models_and_datasets() + get_sklearn_tree_models_and_datasets() + get_sklearn_neighbors_models_and_datasets()
)
```

we should keep `skip` for tests that will one day be re-enabled (after some fix or else)

conftest.py
```diff
@@ -147,6 +147,25 @@ def default_configuration():
...
+        fhe_execution=True,
+    )
+
+
+@pytest.fixture
+def simulation_configuration():
```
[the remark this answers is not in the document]
jfrery: Though I don't use this simulation_configuration in our test for now. I have another branch with this but it doesn't seem to change much.

conftest.py
jfrery: sure. I moved it locally. Will push it if ever I have something important to push. Otherwise it will go in one of the other PR.

src/concrete/ml/common/utils.py
jfrery: Good catch thanks

src/concrete/ml/pytest/utils.py
```diff
@@ -388,6 +388,11 @@ def instantiate_model_generic(model_class, n_bits, **parameters):
...
+    # properly with pytest multi-threading
+    if "n_jobs" in model_class().get_params():
```
RomanBredehoft, line 393: I would prefer to make this less general like , isn't this parameter only for xgb ? if so I would just filter here "if xgb then n_jobs=1" as it's easier to maintain and debug in case we later have some issues related to this !

RomanBredehoft (changes requested):
thanks a lot, I have a few comments !

tests/sklearn/test_sklearn_models.py
jfrery: Indeed! I have not paid attention to the semantic my bad. 

You mean this right? No more linear models

```python
@pytest.mark.parametrize(
    "model_class, parameters",
    get_sklearn_neighbors_models_and_datasets()
    + get_sklearn_tree_models_and_datasets()
    + get_sklearn_neural_net_models_and_datasets(),
)
```

src/concrete/ml/pytest/utils.py
jfrery: No it's also for knn and random forest I believe. We really do not want any inteference with pytest multi threading. We don't really care about parallel training we only have very small dataset in our tests. It won't change anything.

tests/sklearn/test_sklearn_models.py
jfrery: Actually, the test makes sure linear models are not impacted by the p_error. We should keep that, right @RomanBredehoft ?

jfrery (approved):

tests/sklearn/test_sklearn_models.py
RomanBredehoft: ah yes indeed, but yes in that case don't skip it ! just modify the way we check in simulation , good catch

src/concrete/ml/pytest/utils.py
RomanBredehoft: ok sure why not then !

RomanBredehoft (approved):

tests/sklearn/test_sklearn_models.py
jfrery: Yes so this is done in #562.

src/concrete/ml/common/utils.py
jfrery: Done in #562

conftest.py
jfrery: Done in #562

PR #547, opened 2024-03-18
fix: reinstate apidoc generated tags

andrei-stoian-zama:
The tags were removed by accident by editing directly on Gitbook

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -41,6 +41,7 @@
 
 ## References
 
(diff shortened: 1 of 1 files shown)


fd0r (dismissed):

bcm-at-zama (commented):
If I were you, I would also change a bit the script which generates the list: instead of having the '-' lists as of today, I would suggest you to have the '*' as in GB lists. It avoids useless diffs

RomanBredehoft (approved):

PR #549, opened 2024-03-20
chore: add another test that our seeding system is good

bcm-at-zama:
by checking that the seed depends on the file name, function name and parameter 

so:
- it was already working, we had already a seed which was different from one file/function/param to the other
- but I've added tests to prove / check it

Principle of the test:
- we have a few functions in two different files doing the same exact call to random picking
- we run them and put everything in a text file
- then, we check that the first line of this text file is different from others, to check that:
       (i) generated numbers are different
       (ii) sub_seed is different
which proves that, even if you copy a test in different files and functions and parameters, their PRNG is going to behave differently, thanks to the sub_seed system which is different per us

closes #https://github.com/zama-ai/concrete-ml-internal/issues/4325

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,9 +210,9 @@ spcc_internal: $(SPCC_DEPS)
 # -svv disables capturing of stdout/stderr and enables verbose output
 # --count N is to repeate all tests N times (with different seeds). Default is to COUNT=1.
 # --randomly-dont-reorganize is to prevent Pytest from shuffling the tests' order
-# --randomly-dont-reset-seed is to make sure that, if we run the same test several times (with
-# @pytest.mark.repeat(3)), different seeds are used, even if things are still deterministic using 
-# the main seed
+# --randomly-dont-reset-seed is important: if it was not there, the randomly package would reset
+# seeds to the same value, for all tests, resulting in same random's being taken in the tests, which
+# reduces a bit the impact / coverage of our tests
 # --capture=tee-sys is to make sure that, in case of crash, we can search for "Forcing seed to" in 
 # stdout in order to be able to reproduce the failed test using that seed
 # --cache-clear is to clear all Pytest's cache at before running the tests. This is done in order to 
--- a/script/make_utils/check_pytest_determinism.sh
+++ b/script/make_utils/check_pytest_determinism.sh
@@ -70,4 +70,46 @@ echo ""
 echo "diff:"
 echo ""
 diff -u "${OUTPUT_DIRECTORY}/one.modified.txt" "${OUTPUT_DIRECTORY}/three.txt" --ignore-all-space --ignore-blank-lines --ignore-space-change
+
+# Run an execution and then check that the sub_seeds depends on the file name and the function name and parameters
+RANDOMLY_SEED=$RANDOMLY_SEED TEST=tests/seeding/test_seeding_system_file_a.py make pytest_one_single_cpu > "${OUTPUT_DIRECTORY}/seeds.txt"
+
+# This would not be readable
+# SC2181: Check exit code directly with e.g., 'if mycmd;', not indirectly with $?.
+# shellcheck disable=SC2181
+if [ $? -ne 0 ]
+then
+    echo "The commandline failed with:"
+    cat  "${OUTPUT_DIRECTORY}/seeds.txt"
+    exit 255
+fi
+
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -212,7 +212,8 @@ spcc_internal: $(SPCC_DEPS)
 # --randomly-dont-reorganize is to prevent Pytest from shuffling the tests' order
 # --randomly-dont-reset-seed is to make sure that, if we run the same test several times (with
 # @pytest.mark.repeat(3)), different seeds are used, even if things are still deterministic using 
-# the main seed
+# the main seed. --randomly-dont-reset-seed is also important to make that all seeds are different,
```
RomanBredehoft, line 215: there's already a comment for `--randomly-dont-reset-seed` 2 lines above, so you can remove this

script/make_utils/check_pytest_determinism.sh
```diff
@@ -70,4 +70,37 @@ echo ""
...
+    echo "The commandline failed with:"
+    cat  "${OUTPUT_DIRECTORY}/seeds.txt"
+    exit 255
+fi
+
+
+grep "Output" "${OUTPUT_DIRECTORY}/seeds.txt" > "${OUTPUT_DIRECTORY}/seeds_outputs.txt"
+grep "sub_seed" "${OUTPUT_DIRECTORY}/seeds.txt" > "${OUTPUT_DIRECTORY}/seeds_sub_seed.txt"
+
+python script/make_utils/check_first_line_is_different_from_other_lines.py --file "${OUTPUT_DIRECTORY}/seeds_outputs.txt"
```
RomanBredehoft, line 103: could you just explain what you are checking here and why it's important that it passes ?

RomanBredehoft: also why just the first line and not the whole file ?

RomanBredehoft (changes requested):
thanks, I still have a few questions and a small changes !

Makefile
bcm-at-zama: the full sentence is 

```
# --randomly-dont-reset-seed is to make sure that, if we run the same test several times (with
# @pytest.mark.repeat(3)), different seeds are used, even if things are still deterministic using 
# the main seed. --randomly-dont-reset-seed is also important to make that all seeds are different,
# from one test to the other
```

I don't think it's a duplicate, it explains several things

script/make_utils/check_pytest_determinism.sh
bcm-at-zama: adding a 

```
# In the following test, we:
#       log all Output's in seeds_outputs.txt, in a file of N lines: then, we'll check that
#           the first Output is different from the 2nd to N-th once
#       log all sub_seed's in seeds_sub_seed.txt, in a file of N lines: then, we'll check that
#           the first sub_seed is different from the 2nd to N-th once
#
# The goal of these test is to check that, really, seeds and outputs are different, ie that the
# filename or the function name or the parameters were really used in the seeding. Ideally, we 
# should check that all the outputs and seeds are different from each other, but it would be a long
# test (ie, being in N**2 [or ideally N log(N)] instead of N), so it's not worth it
```

It should answer all your questions, does it?

bcm-at-zama (commented):
Answering @RomanBredehoft 's comments

bcm-at-zama:
@RomanBredehoft : tell me what you think with the changes, please

Makefile
RomanBredehoft: well then I'm not sure what you are referring to with your new comment, in what circumstances does `--randomly-dont-reset-seed` makes seeds different "from one test to the other" ? maybe elaborate a bit more in the comment

also I think you are missing a `sure` in `important to make [sure] that all seeds` !

script/make_utils/check_pytest_determinism.sh
RomanBredehoft: ok great thanks !

RomanBredehoft (dismissed):
would just elaborate a bit moer in the makefile comment but apart from that looks good !

bcm-at-zama:
Better now @RomanBredehoft ?

Makefile
bcm-at-zama: new text

RomanBredehoft (approved):
thanks !

PR #557, opened 2024-03-25
chore: add simulation configuration to avoid double compilation

jfrery:


jfrery:
Double compilation doesn't seem to be the main problem. I will come back to this if needed.

PR #562, opened 2024-03-28
chore: speed up our tests

jfrery:


tests/sklearn/test_sklearn_models.py
```diff
@@ -1923,16 +1929,21 @@ def test_rounding_consistency_for_regular_models(
 
 @pytest.mark.parametrize("model_class, parameters", get_sklearn_tree_models_and_datasets())
 @pytest.mark.parametrize("n_bits", [2, 5, 10])
-def test_fhe_sum_for_tree_based_models(
+@pytest.mark.parametrize("execution_mode", ["simulate", "fhe"])
```
RomanBredehoft, line 1932: small detail but I find it more straigthforward to instead define a
```
@pytest.mark.parametrize("execute_in_fhe", [True, False])
```

src/concrete/ml/pytest/utils.py
```diff
@@ -333,36 +333,31 @@ def get_sklearn_all_models_and_datasets(
...
-        )
-        + get_sklearn_neighbors_models_and_datasets(
-            regressor=regressor,
-            classifier=classifier,
-            unique_models=unique_models,
-            select=select,
-            ignore=ignore,
-        )
-    )
+        for (model, dataset_params), _, _ in models_and_datasets:
```
RomanBredehoft, line 352: this is wrong, we should not do this : we have models where we intentionally have several tests for them (using different parameters: tweedie, classifiers for binary or multi classif, ect)

if you want a single test per model, you should use `UNIQUE_MODELS_AND_DATASETS` instead of `MODELS_AND_DATASETS` 

and if you need to call a specific type of model (like trees for example), then you should use the `unique_models` parameters like `get_sklearn_tree_models_and_datasets(unique_models=True)`

RomanBredehoft (changes requested):
I think there's a few changes to be made here

src/concrete/ml/pytest/utils.py
jfrery: > if you want a single test per model, you should use UNIQUE_MODELS_AND_DATASETS instead of MODELS_AND_DATASETS (basically in this PR you make MODELS_AND_DATASETS the same as UNIQUE_MODELS_AND_DATASETS iiuc)

Yep that's not working this is the point of this fix. We are currently checking multiple times each model class. But I went too fast here I didn't make the change at the right place it seems. Thanks will correct.

tests/sklearn/test_sklearn_models.py
jfrery: Yes will update.

src/concrete/ml/pytest/utils.py
jfrery: Let me check without this. This test is 5min so it was a 2m win maybe not the most relevant.

tests/sklearn/test_sklearn_models.py
jfrery: Done

src/concrete/ml/pytest/utils.py
RomanBredehoft: not sure what you mean by "it's not working" 🤔

jfrery: This is what I see. Here it's for random forest but every model is being tested twice. That's not good.

```
| Folder        | File                   | Test               | Test Options                  |   Duration |
|:--------------|:-----------------------|:-------------------|:------------------------------|-----------:|
| tests/sklearn | test_sklearn_models.py | test_input_support | torch-RandomForestClassifier  |      7.901 |
| tests/sklearn | test_sklearn_models.py | test_input_support | numpy-RandomForestClassifier  |      7.901 |
| tests/sklearn | test_sklearn_models.py | test_input_support | list-RandomForestClassifier   |      8.289 |
| tests/sklearn | test_sklearn_models.py | test_input_support | pandas-RandomForestClassifier |      7.877 |
| tests/sklearn | test_sklearn_models.py | test_input_support | pandas-RandomForestClassifier |      2.314 |
| tests/sklearn | test_sklearn_models.py | test_input_support | list-RandomForestClassifier   |      2.261 |
| tests/sklearn | test_sklearn_models.py | test_input_support | numpy-RandomForestClassifier  |      2.313 |
| tests/sklearn | test_sklearn_models.py | test_input_support | torch-RandomForestClassifier  |      2.367 |
```

src/concrete/ml/onnx/ops_impl.py
```diff
@@ -340,7 +340,7 @@ def numpy_sigmoid(
     Returns:
         Tuple[numpy.ndarray]: Output tensor
     """
-    return (1.0 / (1.0 + numpy.exp(-x)),)
+    return (numpy.exp(-numpy.logaddexp(0, -x)),)
```
andrei-stoian-zama, line 343: This improves the dynamic range handling of the sigmoid right ?

andrei-stoian-zama (approved):

src/concrete/ml/onnx/ops_impl.py
jfrery: It's more stable with large values which happens in some of our tests. I wasn't sure if was going to fuse properly but it seems that it does.

src/concrete/ml/pytest/utils.py
RomanBredehoft: are you showing tests of `UNIQUE_MODELS_AND_DATASETS` (unexpected) or `MODELS_AND_DATASETS` (expected) ? @jfrery

tests/sklearn/test_sklearn_models.py
```diff
@@ -241,7 +241,10 @@ def check_correctness_with_sklearn(
 
     # If the model is a classifier, check that accuracies are similar
     if is_classifier_or_partial_classifier(model):
-        check_accuracy(y_pred_sklearn, y_pred_fhe, threshold=threshold_accuracy)
+        # Skip if SGDClassifier
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4344
+        if get_model_name(model_class) != "SGDClassifier":
```
RomanBredehoft, line 246: would have prefered a lower accuracy threshold instead of a skip honestly, but how bad is this ?

RomanBredehoft (approved):
small detail with the SGD but we can tackle this later

PR #573, opened 2024-04-02
chore: refacto compile to handle rounding_threshold_bits

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4351

### DIFF
--- a/src/concrete/ml/common/utils.py
+++ b/src/concrete/ml/common/utils.py
@@ -608,7 +608,7 @@ def array_allclose_and_same_shape(
     return a.shape == b.shape and numpy.allclose(a, b, rtol, atol, equal_nan)
 
 
-def check_rounding_threshold(rounding_threshold_bits):
+def process_rounding_threshold_bits(rounding_threshold_bits):
     """Check and process the rounding_threshold_bits parameter.
 
     Args:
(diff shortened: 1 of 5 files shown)


RomanBredehoft (changes requested):
could you also add the call to `check_rounding_threshold` in : 
- `build_quantized_module` (this method can be called separately from `_compile_torch_or_onnx_model`)
- `ONNXConverter`'s init (just to be safe)

jfrery:
> could you also add the call to `check_rounding_threshold` in :
> 
>     * `build_quantized_module` (this method can be called separately from `_compile_torch_or_onnx_model`)
> 
>     * `ONNXConverter`'s init (just to be safe)

Yes good point about `build_quantized_module`. I have done the changes.

src/concrete/ml/common/utils.py
```diff
@@ -605,3 +606,50 @@ def array_allclose_and_same_shape(
     assert isinstance(b, numpy.ndarray)
 
     return a.shape == b.shape and numpy.allclose(a, b, rtol, atol, equal_nan)
+
+
+def check_rounding_threshold(rounding_threshold_bits):
```
RomanBredehoft, line 611: did not realize that this was not only checking but also "building" the object. I would rename it to something else then, like `process_rounding_threshold_bits`

jfrery: Sure!

RomanBredehoft (approved):
great thanks a lot !

PR #575, opened 2024-04-02
chore: fix sync repository action (bis)

RomanBredehoft:
it's still failing : https://github.com/zama-ai/concrete-ml/actions/runs/8522402034/job/23342541783

the issue is that the external action does not provide any LFS options

also, it looks like we were cloning the repo twice, so I remove the external action and simplified to these steps

### DIFF
--- a/.github/workflows/sync_on_push.yaml
+++ b/.github/workflows/sync_on_push.yaml
@@ -31,10 +31,17 @@ jobs:
           name: repo-archive
           path: '.'
 
+      # - name: Sync internal repository
+      #   run: |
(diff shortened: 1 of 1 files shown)


RomanBredehoft:
I'm usually not a fan of doing this when we are doing things related to main, but in this case I think it should be good indeed, will do thanks !

RomanBredehoft:
I can't find a way to make this work, it's pretty weird

jfrery (approved):
Thanks for fixing this!

RomanBredehoft:
finally working 🎉

jfrery (approved):

PR #576, opened 2024-04-02
docs: add dataframe documentation

andrei-stoian-zama:

PR #580, opened 2024-04-03
chore: remove update SUMMARY in make apidocs step

RomanBredehoft:
it's not necessary anymore, and keeping makes `make apidocs` fail


andrei-stoian-zama (approved):

RomanBredehoft:
bceause os this @bcm-at-zama : https://github.com/zama-ai/concrete-ml/pull/576#discussion_r1549212458

bcm-at-zama:
ok look like we are speaking about it at two different places, let's continue in the first place

PR #582, opened 2024-04-03
chore: improve save/load methods for encrypted data-frames

RomanBredehoft:
this should make the save/load methods faster

### DIFF
--- a/src/concrete/ml/pandas/_utils.py
+++ b/src/concrete/ml/pandas/_utils.py
@@ -134,7 +134,7 @@ def deserialize_elementwise(array: numpy.ndarray) -> numpy.ndarray:
 
 
 def serialize_evaluation_keys(evaluation_keys: fhe.EvaluationKeys) -> bytes:
-    """Serialize the evaluation keys into a string of hexadecimal numbers.
+    """Serialize the evaluation keys into bytes.
 
     Args:
         evaluation_keys (fhe.EvaluationKeys): The evaluation keys to serialize.
--- a/src/concrete/ml/pandas/_utils.py
+++ b/src/concrete/ml/pandas/_utils.py
@@ -133,28 +133,28 @@ def deserialize_elementwise(array: numpy.ndarray) -> numpy.ndarray:
     return numpy.vectorize(deserialize_value)(array)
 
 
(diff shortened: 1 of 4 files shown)


src/concrete/ml/pandas/_utils.py
```diff
@@ -133,28 +133,28 @@ def deserialize_elementwise(array: numpy.ndarray) -> numpy.ndarray:
     return numpy.vectorize(deserialize_value)(array)
 
 
-def serialize_evaluation_keys(evaluation_keys: fhe.EvaluationKeys) -> str:
+def serialize_evaluation_keys(evaluation_keys: fhe.EvaluationKeys) -> bytes:
```
andrei-stoian-zama, line 136: why do you need hex? it requires processing that is slow

src/concrete/ml/pandas/dataframe.py
```diff
@@ -318,9 +320,16 @@ def save(self, path: Union[Path, str]):
...
-        with path.open("w", encoding="utf-8") as file:
-            json.dump(encrypted_df_dict, file)
+        if path.suffix != ".zip":
+            path = path.with_suffix(".zip")
+
+        encrypted_df_dict, evaluation_keys = self._to_dict_and_eval_keys()
+
+        encrypted_df_json_bytes = json.dumps(encrypted_df_dict).encode(encoding="utf-8")
+
+        with zipfile.ZipFile(path, "w") as zip_file:
```
andrei-stoian-zama, line 330: can you make sure there is no compression. it's useless to compress the eval key which looks random and it just slows down the saving process

andrei-stoian-zama (changes requested):
Thanks for the quick PR. Can you clarify a few things please?

src/concrete/ml/pandas/_utils.py
RomanBredehoft: I removed it actually, just forgot to update the docstring as well good catch

initially it was for making it serializable in json

src/concrete/ml/pandas/dataframe.py
RomanBredehoft: yes

andrei-stoian-zama (approved):

PR #583, opened 2024-04-03
chore: add timing in data-frame notebook example

RomanBredehoft:


andrei-stoian-zama (approved):

deps_licenses/licenses_mac_silicon_user.txt
```diff
@@ -2,14 +2,14 @@ Name, Version, License
 GitPython, 3.1.41, BSD License
 PyYAML, 6.0.1, MIT License
 anyio, 3.7.1, MIT License
-boto3, 1.34.72, Apache Software License
-botocore, 1.34.72, Apache Software License
+boto3, 1.34.75, Apache Software License
```
andrei-stoian-zama, line 5: any idea why this gets updated ?

RomanBredehoft: yes, the action that updates the licenses do not handle macos arm, only macos intel , so the first time I make conformance locally I need to push them as well

PR #585, opened 2024-04-03
fix: SGDClassifier post-processing with multi-class and improve linear models' predict method

RomanBredehoft:
first step towards a green weekly CI 

So there are a few things to note in this PR : 
- the weekly CI was failing because SGD classifier's `post_processing` method was not properly integrated : instead, everything was done in `predict_proba`, which interferes with the way we build and test our API (notably the `test_separated_inference`)
- I found a more general "issue" related to linear models' `predict` method: while trees / qnns usually do `predict_proba` + `argmax`, for linear models sklearn does `decision_function` + `argmax`. In theory this should not change anything as the `predict_proba` basically does a sigmoid/softmax or normalization, but in practice it made the argmax behave differently because of slight floating points errors (basically the same as in https://github.com/zama-ai/concrete-ml-internal/issues/3369)
- while debugging this, I also encountered https://github.com/zama-ai/concrete-ml-internal/issues/4029 and decided to fix it once and for all, else I more or less had to make linear classifiers ouputs' shape not coherent with sklearn which seemed going backward

Why are we discovering this only recently ? A few reasons : 
- for linear classifiers: it's been quite a long time now that we almost never test our classifier's `predict` method because of the float issues mentioned above : we consider that the most important fact is making sure that quantized == quantized. The only paces where this seem to happen is the hyper parameter test, which is where @jfrery detected the issue  
- for sgd's post processing: the way we were doing post-processing was good, the issue was only when running the inference in separated ways like in `test_separated_inference`. Still, our custom post_processing function was good for 1D array (tested in regular CIs) with "log_loss" loss (default). Only the weekly tests 2D arrays, which is where we found the error

I also took the liberty to clean a bit some parts/tests (no breaking changes) 

closes https://github.com/zama-ai/concrete-ml-internal/issues/4344
closes https://github.com/zama-ai/concrete-ml-internal/issues/4252
closes https://github.com/zama-ai/concrete-ml-internal/issues/4029

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -13,6 +13,7 @@ from concrete.fhe import Graph as CPGraph
 from concrete.fhe.compilation import Circuit, Configuration
 from concrete.fhe.mlir.utils import MAXIMUM_TLU_BIT_WIDTH
 from sklearn.datasets import make_classification, make_regression
+from sklearn.metrics import accuracy_score
 
 from concrete.ml.common.utils import (
     SUPPORTED_FLOAT_TYPES,
@@ -420,7 +421,7 @@ def check_accuracy():
     """Fixture to check the accuracy."""
 
     def check_accuracy_impl(expected, actual, threshold=0.9):
-        accuracy = numpy.mean(expected == actual)
+        accuracy = accuracy_score(expected, actual)
         assert accuracy >= threshold, f"Accuracy of {accuracy} is not high enough ({threshold})."
 
     return check_accuracy_impl
--- a/src/concrete/ml/sklearn/base.py
+++ b/src/concrete/ml/sklearn/base.py
@@ -1783,13 +1783,17 @@ class SklearnLinearClassifierMixin(
         y_proba = self.post_processing(y_logits)
         return y_proba
 
-    # In scikit-learn, the argmax is done on the scores directly, not the probabilities
+    # In scikit-learn, the argmax is done on the logits directly, not the probabilities
     def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
         # Compute the predicted scores
-        y_proba = self.decision_function(X, fhe=fhe)
+        y_logits = self.decision_function(X, fhe=fhe)
 
         # Retrieve the class with the highest score
-        y_preds = numpy.argmax(y_proba, axis=1)
+        # If there is a single dimension, only compare the scores to 0
+        if y_logits.ndim == 1 or y_logits.shape[1] == 1:
+            y_preds = (y_logits > 0).astype(int)
+        else:
+            y_preds = numpy.argmax(y_logits, axis=1)
 
         return self.classes_[y_preds]
 
--- a/src/concrete/ml/sklearn/base.py
+++ b/src/concrete/ml/sklearn/base.py
@@ -808,11 +808,9 @@ class BaseClassifier(BaseEstimator):
             # If the prediction array is 1D, transform the output into a 2D array [1-p, p],
             # with p the initial output probabilities
             # This is similar to what is done in scikit-learn
-            if y_preds.ndim == 1:
-                y_preds = numpy.vstack([1 - y_preds, y_preds]).T
-
-            elif y_preds.shape[1] == 1:
-                y_preds = numpy.concatenate((1 - y_preds, y_preds), axis=1)
+            if y_preds.ndim == 1 or y_preds.shape[1] == 1:
+                y_preds = y_preds.flatten()
+                return numpy.vstack([1 - y_preds, y_preds]).T
 
         # Else, apply the softmax operator
         else:
--- a/src/concrete/ml/sklearn/base.py
+++ b/src/concrete/ml/sklearn/base.py
@@ -694,6 +694,8 @@ class BaseEstimator:
         Returns:
             numpy.ndarray: The post-processed predictions.
         """
+        assert isinstance(y_preds, numpy.ndarray), "Output predictions must be an array."
+
         return y_preds
 
 
@@ -805,7 +807,11 @@ class BaseClassifier(BaseEstimator):
 
             # If the prediction array is 1D, transform the output into a 2D array [1-p, p],
             # with p the initial output probabilities
-            if y_preds.ndim == 1 or y_preds.shape[1] == 1:
+            # This is similar to what is done in scikit-learn
+            if y_preds.ndim == 1:
+                y_preds = numpy.vstack([1 - y_preds, y_preds]).T
+
+            elif y_preds.shape[1] == 1:
                 y_preds = numpy.concatenate((1 - y_preds, y_preds), axis=1)
 
         # Else, apply the softmax operator
@@ -1387,8 +1393,13 @@ class BaseTreeEstimatorMixin(BaseEstimator, sklearn.base.BaseEstimator, ABC):
     def dequantize_output(self, q_y_preds: numpy.ndarray) -> numpy.ndarray:
         self.check_model_is_fitted()
 
-        q_y_preds = self.output_quantizers[0].dequant(q_y_preds)
-        return q_y_preds
+        y_preds = self.output_quantizers[0].dequant(q_y_preds)
+
+        # If the preds have shape (n, 1), squeeze it to shape (n,) like in scikit-learn
+        if y_preds.ndim == 2 and y_preds.shape[1] == 1:
+            return y_preds.ravel()
+
+        return y_preds
 
     def _get_module_to_compile(self) -> Union[Compiler, QuantizedModule]:
         assert self._tree_inference is not None, self._is_not_fitted_error_message()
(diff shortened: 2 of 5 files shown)


src/concrete/ml/sklearn/linear_model.py
```diff
@@ -254,27 +254,6 @@ def __init__(
                     f"({fit_encrypted=}). Got {parameters_range=}"
                 )
 
-    def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
```
RomanBredehoft, line 257: we don't need it as this is essentially a (bad) copy paste of the linea models' `post_processing` method, which is inherited and used anyway

tests/sklearn/test_sklearn_models.py
```diff
@@ -1605,31 +1605,11 @@ def test_predict_correctness(
 
 
 @pytest.mark.parametrize("model_class, parameters", MODELS_AND_DATASETS)
-@pytest.mark.parametrize(
```
RomanBredehoft, line 1608: there's no point of simulating here (we are testing `encrypt` + `run` + `decrypt`)

tests/sklearn/test_sklearn_models.py
```diff
@@ -1605,31 +1605,11 @@ def test_predict_correctness(
...
-    [
-        pytest.param(False, id="fhe"),
-    ],
-)
-# N_BITS_LINEAR_MODEL_CRYPTO_PARAMETERS bits is currently the
-# limit to find crypto parameters for linear models
-# make sure we only compile below that bit-width.
-# Additionally, prevent computations in FHE with too many bits
-@pytest.mark.parametrize(
-    "n_bits",
```
RomanBredehoft, line 1619: there no real point of testing multiple n_bits values, as we are only testing the api + comparing fhe vs simulation

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -254,27 +254,6 @@ def __init__(
...
-
-            # If the prediction array is 1D, transform the output into a 2D array [1-p, p],
-            # with p the initial output probabilities
-            if y_preds.ndim == 1 or y_preds.shape[1] == 1:
-                y_preds = numpy.concatenate((1 - y_preds, y_preds), axis=1)
-
-        # Else, apply the softmax operator
-        else:
-            y_preds = numpy_sigmoid(y_preds)[0]
-            y_preds = y_preds / y_preds.sum(axis=1)
```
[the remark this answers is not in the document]
jfrery: How did this pass our tests?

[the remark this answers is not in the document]
RomanBredehoft: actually I was wrong, this was not the issue my bad

src/concrete/ml/sklearn/base.py
```diff
@@ -1735,38 +1735,6 @@ class SklearnLinearRegressorMixin(SklearnLinearModelMixin, sklearn.base.Regresso
     """
 
 
-class SklearnSGDRegressorMixin(SklearnLinearRegressorMixin):
```
RomanBredehoft, line 1738: just moved it for better readability

src/concrete/ml/sklearn/base.py
```diff
@@ -1815,6 +1783,48 @@ def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) ->
         y_proba = self.post_processing(y_logits)
         return y_proba
 
+    # In scikit-learn, the argmax is done on the scores directly, not the probabilities
+    def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
```
RomanBredehoft, line 1787: this fixes https://github.com/zama-ai/concrete-ml-internal/issues/4344 (reason is explained in main comment, but basically this is how sklearn does)

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -253,27 +252,12 @@ def __init__(
                     "Setting 'parameter_range' is mandatory if FHE training is enabled "
                     f"({fit_encrypted=}). Got {parameters_range=}"
                 )
-
-    def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
```
RomanBredehoft, line 257: this post_processing was wrong + I moved it below

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -835,61 +819,24 @@ def partial_fit(
...
-            return numpy.vstack([1 - prob, prob]).T
-
-        # OvR normalization, like LibLinear's predict_probability
-        prob /= prob.sum(axis=1).reshape((prob.shape[0], -1))
-
-        return prob
-
-    def predict_proba(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.ndarray:
-        """Probability estimates.
+    def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
```
RomanBredehoft, line 822: basically we don't need to define the `predict_proba` method as it is in sklearn, we only need to re-define `post_processing` with sklearn's implem

tests/sklearn/test_fhe_training.py
```diff
@@ -133,27 +133,23 @@ def test_fit_error_if_non_binary_targets(n_bits, max_iter, parameter_min_max):
         model.partial_fit(x, y, fhe="disable")
 
 
-@pytest.mark.parametrize("loss", ["log_loss", "modified_huber"])
```
RomanBredehoft, line 136: here we are testing error raises for specific arguments, we don't need all these inputs

tests/sklearn/test_sklearn_models.py
```diff
@@ -651,7 +648,11 @@ def check_separated_inference(model, fhe_circuit, x, check_float_array_equal):
         is_classifier_or_partial_classifier(model)
         and get_model_name(model) != "KNeighborsClassifier"
     ):
-        y_pred = numpy.argmax(y_pred, axis=-1)
+        # For linear classifiers, the argmax is done on the scores directly, not the probabilities
+        if is_model_class_in_a_list(model, _get_sklearn_linear_models()):
```
RomanBredehoft, line 652: as mentioned in the main comment, let's do it like sklearn

jfrery (approved):
Awesome! Thanks a lot for the fixes + cleaning. Great explanations as well.

conftest.py
```diff
@@ -420,7 +421,7 @@ def check_accuracy():
     """Fixture to check the accuracy."""
 
     def check_accuracy_impl(expected, actual, threshold=0.9):
-        accuracy = numpy.mean(expected == actual)
+        accuracy = accuracy_score(expected, actual)
```
RomanBredehoft, line 424: this is because we have https://github.com/zama-ai/concrete-ml-internal/issues/4029, which is already a fixme + skip in the modified tests below

basically we have a shape difference with sklearn for outputs for may models (we output (100,1) while sklearn gives (100,), so the previous formula was failing). Let's fix this is another PR !

RomanBredehoft, line 424: better imo

src/concrete/ml/sklearn/base.py
```diff
@@ -805,8 +807,12 @@ def post_processing(self, y_preds: numpy.ndarray) -> numpy.ndarray:
 
             # If the prediction array is 1D, transform the output into a 2D array [1-p, p],
             # with p the initial output probabilities
-            if y_preds.ndim == 1 or y_preds.shape[1] == 1:
-                y_preds = numpy.concatenate((1 - y_preds, y_preds), axis=1)
+            # This is similar to what is done in scikit-learn
+            if y_preds.ndim == 1:
+                return numpy.vstack([1 - y_preds, y_preds]).T
```
RomanBredehoft, line 812: like in sklearn

src/concrete/ml/sklearn/base.py
```diff
@@ -1387,8 +1393,13 @@ def quantize_input(self, X: numpy.ndarray) -> numpy.ndarray:
     def dequantize_output(self, q_y_preds: numpy.ndarray) -> numpy.ndarray:
         self.check_model_is_fitted()
 
-        q_y_preds = self.output_quantizers[0].dequant(q_y_preds)
-        return q_y_preds
+        y_preds = self.output_quantizers[0].dequant(q_y_preds)
+
+        # If the preds have shape (n, 1), squeeze it to shape (n,) like in scikit-learn
+        if y_preds.ndim == 2 and y_preds.shape[1] == 1:
```
RomanBredehoft, line 1399: like in sklearn (same for the followings one)

tests/sklearn/test_sklearn_models.py
```diff
@@ -204,13 +204,11 @@ def check_correctness_with_sklearn(
...
             y_scores_fhe = model.decision_function(x, fhe=fhe)
 
-            # Currently, for single target data sets, Concrete models' outputs have shape (n, 1)
-            # while scikit-learn models' outputs have shape (n, )
-            # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4029
-            # assert y_scores_sklearn.shape == y_scores_fhe.shape, (
-            #     "Method 'decision_function' outputs different shapes between scikit-learn and "
-            #     f"Concrete ML in FHE (fhe={fhe})"
-            # )
+            assert y_scores_sklearn.shape == y_scores_fhe.shape, (
```
RomanBredehoft, line 207: put back the assert here and below

jfrery (approved):
Thanks!

PR #587, opened 2024-04-04
fix: disable overflow protection in rounding

andrei-stoian-zama:


src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -1017,5 +1017,7 @@ def cnp_round(
                     x, lsbs_to_remove=lsbs_value, exactness=exactness, overflow_protection=False
                 )
             else:
-                x = fhe.round_bit_pattern(x, lsbs_to_remove=lsbs_value, exactness=exactness)
+                x = fhe.round_bit_pattern(
```
jfrery, line 1020: You can just remove the if else.

jfrery (approved):

PR #588, opened 2024-04-04
chore: avoid checking for coverage in weekly tests with pypi

RomanBredehoft:
second step towards a green weekly CI

closes https://github.com/zama-ai/concrete-ml-internal/issues/4251


jfrery (commented):
I am not sure I understand how the diff on parallel pytest is related to the coverage?

RomanBredehoft:
@jfrery if you look at the makefile, command `pytest` is basically `pytest_parallel_internal` + coverage plugin, but we don't want coverage (and keep it parallelized), so that's why !

PR #589, opened 2024-04-04
chore: add test_reduce_sum configuration as flaky

RomanBredehoft:
refs https://github.com/zama-ai/concrete-ml-internal/issues/4357


jfrery (approved):

RomanBredehoft:
it looks like the new flaky will happen a lot. I first want to make sure with CP if there's actually an issue on their side or if the test does not make much sense

RomanBredehoft:
actually, let's see what CP says about https://github.com/zama-ai/concrete-internal/issues/681 before putting this as a known flaky

PR #608, opened 2024-04-15
docs: fix broken links

yuxizama:


RomanBredehoft (changes requested):
there are other similar changes to be done, I'll push them here as well, but thanks for these yes !

bcm-at-zama:
@RomanBredehoft : how did you find those additional changes?

RomanBredehoft:
just with ctrl + f , I asically just removed all `https` links, but we might still have some broken links as I didn't / couldn't check all relative references , hence why I opened this issue : https://github.com/zama-ai/concrete-ml-internal/issues/4372

RomanBredehoft (approved):
now good !

PR #610, opened 2024-04-15
chore: remove TLU in reduce_sum test

RomanBredehoft:
New steps towards a green CI 

We are removing the tlu in this test as sum should sort of always be followed with rounding before a tlu


jfrery (approved):
Thanks!

PR #616, opened 2024-04-16
chore: integrate input compression

RomanBredehoft:
enable input ciphertext compression by default through an environmental variable and add a test to make sure the size is indeed smaller

based on experiments, this feature can reduce input ciphertexts of hundreds of MB to less than 1 MB. Also, no significant slowdown have been observed for the inference time executions, expect with fully levelled circuits (where the inference can take a few hundreds milliseconds more). 

For example : 
- no input compression (https://github.com/zama-ai/concrete-ml/actions/runs/8783678006/job/24100432554)
- with input compression (https://github.com/zama-ai/concrete-ml/actions/runs/8784270343/job/24102239332)

closes https://github.com/zama-ai/concrete-ml-internal/issues/4365


tests/deployment/test_client_server.py
```diff
@@ -110,8 +111,15 @@ def test_client_server_sklearn(
             x_test, model, key_dir, check_array_equal, check_float_array_equal
         )
 
+    compilation_kwargs = {
+        "X": x_train,
+        "configuration": default_configuration,
+    }
```
fd0r, line 117: Not sure how this is better but why not

fd0r (approved):
Looks good to me, thanks for integrating this!

RomanBredehoft:
> Looking at this I'm wondering if we shouldn't expose the compilation options to our users with the use of keyword arguments in the future.

we can talk about it at the weekly but I think we want to avoid having many non-ml parameters on CML side (cf https://github.com/zama-ai/concrete-ml/pull/519#discussion_r1521389819) 

> Also not sure why this is set through a keyword argument instead of an attribute in the Configuration object.

in CP you can either pass kwargs or use a config object, if both are set than the value from the kwargs is taken, so I think it's best to do it like this since we want to make it default !

fd0r:
> avoid having many non-ml parameters on CML side 

Makes sense

> if both are set than the value from the kwargs is taken

That's imo a confusing behavior for the users, we should probably talk about it with @umut-sahin. Would be great to have a `Configuration` object with CML defaults or something like that.

umut-sahin:
I think it makes sense, kwargs are like a the last place you can set configuration, right within the compile call. You can have a base configuration and overwrite it with kwargs for example instead of changing the base configuration.

andrei-stoian-zama (approved):
Good for me, thanks!

PR #619, opened 2024-04-17
chore: use cp vl when circuit is CRT based

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4393

### DIFF
--- a/src/concrete/ml/quantization/quantized_module.py
+++ b/src/concrete/ml/quantization/quantized_module.py
@@ -526,7 +526,7 @@ class QuantizedModule:
 
                 # Whether the old simulation method should be used based on USE_OLD_VL
                 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4391
-                # and the circuit encoding (CRT encoding should use CP VL)
+                # and the circuit encoding (CRT encoding should use Concrete VL)
                 if USE_OLD_VL or is_crt_encoding:
                     predict_method = partial(
                         self.fhe_circuit.graph, p_error=self.fhe_circuit.p_error
--- a/src/concrete/ml/quantization/quantized_module.py
+++ b/src/concrete/ml/quantization/quantized_module.py
@@ -522,14 +522,15 @@ class QuantizedModule:
 
             # If the inference should be executed using simulation
             if simulate:
+                is_crt_encoding = self.fhe_circuit.statistics["packing_key_switch_count"] != 0
 
-                # If the old simulation method should be used
-                if USE_OLD_VL:
+                # Whether the old simulation method should be used based on USE_OLD_VL
+                # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4391
+                # and the circuit encoding (CRT encoding should use CP VL)
+                if USE_OLD_VL or is_crt_encoding:
                     predict_method = partial(
                         self.fhe_circuit.graph, p_error=self.fhe_circuit.p_error
                     )  # pragma: no cover
-
-                # Else, use the official simulation method
                 else:
                     predict_method = self.fhe_circuit.simulate
 
--- a/src/concrete/ml/quantization/quantized_module.py
+++ b/src/concrete/ml/quantization/quantized_module.py
@@ -524,13 +524,16 @@ class QuantizedModule:
             if simulate:
                 is_crt_encoding = self.fhe_circuit.statistics["packing_key_switch_count"] != 0
 
-                # Whether the old simulation method should be used based on USE_OLD_VL
+                # If the virtual library method should be used
+                # For now, use the virtual library when simulating
+                # circuits that use CRT  encoding because the official simulation is too slow
                 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4391
-                # and the circuit encoding (CRT encoding should use Concrete VL)
                 if USE_OLD_VL or is_crt_encoding:
                     predict_method = partial(
                         self.fhe_circuit.graph, p_error=self.fhe_circuit.p_error
                     )  # pragma: no cover
+
+                # Else, use the official simulation method
                 else:
                     predict_method = self.fhe_circuit.simulate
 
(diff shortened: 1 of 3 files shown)


andrei-stoian-zama (approved):
Looks good, just a minor doubt on whether the compiled model will always be compilable and CRT with n_bits=4. But otherwise great!

tests/quantization/test_quantized_module.py
```diff
@@ -512,3 +512,49 @@ def test_quantized_module_initialization_error():
             ordered_module_output_names=None,  # This makes the combination invalid
             quant_layers_dict={"layer1": (["input1"], "QuantizedOp")},
         )
+
+
+@pytest.mark.parametrize("model_class, input_shape", [pytest.param(FC, (100, 32 * 32 * 3))])
+def test_crt_simulation_using_cp_vl(model_class, input_shape, default_configuration):
```
RomanBredehoft, line 518: the test name is more about checking that rounding makes the circuit non-crt, there's nothing about simulation in this test, so maybe you can rename `test_crt_simulation_using_cp_vl` to `test_crt_to_non_crt_using_rouding` or else

src/concrete/ml/sklearn/base.py
```diff
@@ -645,17 +645,19 @@ def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.
...
+                    # and the circuit encoding (CRT encoding should use Concrete VL)
+                    if USE_OLD_VL or is_crt_encoding:
                         predict_method = partial(
                             self.fhe_circuit.graph, p_error=self.fhe_circuit.p_error
                         )  # pragma: no cover
 
                     # Else, use the official simulation method
                     else:
                         predict_method = self.fhe_circuit.simulate
-
```
RomanBredehoft, line 658: better to keep a space between a code line and a following comment

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -522,14 +522,15 @@ def _fhe_forward(
...
-                if USE_OLD_VL:
+                # Whether the old simulation method should be used based on USE_OLD_VL
+                # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4391
+                # and the circuit encoding (CRT encoding should use Concrete VL)
+                if USE_OLD_VL or is_crt_encoding:
                     predict_method = partial(
                         self.fhe_circuit.graph, p_error=self.fhe_circuit.p_error
                     )  # pragma: no cover
-
-                # Else, use the official simulation method
```
RomanBredehoft, line 532: not sure why the comment was removed, and same, better to keep an empty line above !

src/concrete/ml/sklearn/base.py
```diff
@@ -645,17 +645,19 @@ def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.
 
                 # If the inference should be executed using simulation
                 if fhe == "simulate":
+                    is_crt_encoding = self.fhe_circuit.statistics["packing_key_switch_count"] != 0
 
-                    # If the old simulation method should be used
-                    if USE_OLD_VL:
+                    # Whether the old simulation method should be used based on USE_OLD_VL
```
RomanBredehoft, line 650: I would rephrase this with : 
```
# If the virtual library method should be used 
# For now, use the virtual library when simulating circuits that uses CRT 
# decomposition because the official simulation is too slow 
# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4391
```

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -522,14 +522,15 @@ def _fhe_forward(
 
             # If the inference should be executed using simulation
             if simulate:
+                is_crt_encoding = self.fhe_circuit.statistics["packing_key_switch_count"] != 0
 
-                # If the old simulation method should be used
-                if USE_OLD_VL:
+                # Whether the old simulation method should be used based on USE_OLD_VL
```
RomanBredehoft, line 527: I would rephrase this with : 
```
# If the virtual library method should be used 
# For now, use the virtual library when simulating circuits that uses CRT 
# decomposition because the official simulation is too slow 
# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4391
```

tests/quantization/test_quantized_module.py
```diff
@@ -512,3 +512,49 @@ def test_quantized_module_initialization_error():
...
+    "Test simulation based on whether the circuit uses CRT encoding."
```
RomanBredehoft, line 519: same here, maybe put "Tes that using rounding makes circuits non-CRT" or something like that

RomanBredehoft (changes requested):
thanks for this update, I would just make some things a bit more clear !

src/concrete/ml/sklearn/base.py
jfrery: in what?

src/concrete/ml/quantization/quantized_module.py
jfrery: I don't like this `# If the virtual library method should be used `. Feels like something else is coming to complete the sentence.

src/concrete/ml/sklearn/base.py
RomanBredehoft: in this : https://github.com/zama-ai/concrete-ml/pull/619/files#r1568710561

andrei-stoian-zama:
This also closes https://github.com/zama-ai/concrete-ml-internal/issues/4391 right ?

src/concrete/ml/quantization/quantized_module.py
jfrery: Added it back

src/concrete/ml/sklearn/base.py
jfrery: Fixed

src/concrete/ml/quantization/quantized_module.py
jfrery: Anyway back to the original comment with some of your suggestion.

RomanBredehoft (approved):
great thanks !

RomanBredehoft:
no @andrei-stoian-zama because the issue you ink is used as a fixme, until this is somehow fixed by concrete

PR #622, opened 2024-04-18
chore: clean encrypted training and SGDClassifier files

RomanBredehoft:
I was starting to work on https://github.com/zama-ai/concrete-ml-internal/issues/4374 but realized the SGDClassifier could benefit from a bit of cleaning

also wanted to fix https://github.com/zama-ai/concrete-ml-internal/issues/4184 but it's actually not that trivial so we'll see later if it's really needed

### DIFF
--- a/src/concrete/ml/sklearn/linear_model.py
+++ b/src/concrete/ml/sklearn/linear_model.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Optional, Union
 
 import numpy
 import sklearn.linear_model
-from sklearn.linear_model import SGDClassifier as SklearnSGDClassifier
 from sklearn.preprocessing import LabelEncoder
 
 from ..common.utils import FheMode
@@ -229,7 +228,9 @@ class SGDClassifier(SklearnSGDClassifierMixin):
         self.average = average
 
         # Checks the coherence of some attributes
-        assert isinstance(self.max_iter, int)
+        assert (
+            isinstance(self.max_iter, int) and self.max_iter >= 1
+        ), "Parameter 'max_iter' must be an integer in range [1, inf]"
         assert isinstance(self.tol, (float, type(None)))
 
         # Checks and warnings for FHE training
@@ -307,8 +308,8 @@ class SGDClassifier(SklearnSGDClassifierMixin):
                 [self.parameters_range[0], self.parameters_range[1]],  # Bias
             )
         )
-        compile_size = len(combinations)
 
+        compile_size = len(combinations)
         n_targets = 1
 
         # Generate the input values to consider for compilation
@@ -317,13 +318,7 @@ class SGDClassifier(SklearnSGDClassifierMixin):
         # Generate the target values to consider for compilation
         # Update this once we support multi-class
         # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4182
-        y_compile_set = numpy.empty(
-            (
-                compile_size,
-                self.batch_size,
-                n_targets,
-            )
-        )
+        y_compile_set = numpy.empty((compile_size, self.batch_size, n_targets))
 
         # Generate the weight values to consider for compilation
         weights_compile_set = numpy.empty((compile_size, x_min.shape[0], n_targets))
@@ -338,8 +333,10 @@ class SGDClassifier(SklearnSGDClassifierMixin):
             compile_set[0][index] = x_value
             compile_set[1][index] = label
             compile_set[2][index] = coef_value
+
             if not self.fit_intercept:
                 bias_value *= 0.0
+
             compile_set[3][index] = bias_value
 
         # Instantiate the LogisticRegressor model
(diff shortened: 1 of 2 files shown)


src/concrete/ml/sklearn/linear_model.py
```diff
@@ -398,6 +400,8 @@ def _fit_encrypted(
                 optimization. Default to None.
             intercept_init (Optional[numpy.ndarray]): The initial intercept to warm-start the
                 optimization. Default to None.
+            is_partial_fit (bool): Indicates if this fit represents a partial fit. A partial fit is
```
RomanBredehoft, line 403: basically I removed `_encrypted_fit_one_step` (made for `partial_fit`) and added this new param in `_fit_encrypted` to handle the few differences with between `fit` and `partial_fit`

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -418,90 +422,115 @@ def _fit_encrypted(
...
 
-        # Initialize the weight values
+        # Initialize the weight values with the given ones if some are provided
         if coef_init is not None:
             weights = coef_init
-        elif self.warm_start and self._weights_encrypted_fit is not None:
+
+        # Else, if warm start is activated or this is a partial fit, use some already computed
+        # weight values have if there are some
+        elif (self.warm_start or is_partial_fit) and self._weights_encrypted_fit is not None:
```
RomanBredehoft, line 484: this is more or less a fix : before, we were asking users to set `warm_start` to True in order to benefit from `partial_fit`, but that's not the case in sklearn (https://scikit-learn.org/stable/glossary.html#term-partial_fit)

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -663,132 +709,6 @@ def fit(  # type: ignore[override]
             sample_weight=sample_weight,
         )
 
-    # pylint: disable-next=too-many-branches,too-many-statements
-    def _fit_encrypted_one_step(self, X, y, fhe):
```
RomanBredehoft, line 667: as I mentioned above, this is no longer needed (and it removes a lot of duplicated code)

tests/sklearn/test_fhe_training.py
```diff
@@ -293,259 +302,236 @@ def test_clear_fit(
     model.predict_proba(x, fhe="simulate")
 
 
-# pylint: disable=too-many-statements,protected-access,too-many-locals
-@pytest.mark.parametrize("fit_intercept", [True, False])
-@pytest.mark.parametrize("label_offset", [0, 1])
-@pytest.mark.parametrize("n_bits, max_iter, parameter_min_max", [pytest.param(7, 30, 1.0)])
-def test_encrypted_fit_coherence(
-    fit_intercept, label_offset, n_bits, max_iter, parameter_min_max, check_accuracy
+def check_encrypted_fit(
```
RomanBredehoft, line 305: I refactored the test by creating this function

tests/sklearn/test_fhe_training.py
```diff
@@ -293,259 +302,236 @@ def test_clear_fit(
...
+    assert array_allclose_and_same_shape(y_pred_proba_disable, y_pred_proba_mixed)
+    assert array_allclose_and_same_shape(y_pred_class_disable, y_pred_class_mixed)
+
+    # Define warn start parameters
+    warm_start_fit_kwargs = {
+        "coef_init": weights_disable,
+        "intercept_init": bias_disable,
+    }
+
+    # Because the model's `random_number_generator` will be re-initialized with the same random
```
RomanBredehoft, line 523: I tried to add checking values coherence at the end but unfortunately the fact that we re-initialize the rng instance at each model init with the exact same random_state more or less prevents us to do so

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -361,19 +360,21 @@ def _get_training_quantized_module(
...
         )
         end = time.time()
+
         if self.verbose:
             print(f"Compilation took {end - start:.4f} seconds.")
 
         return training_quantized_module
 
-    # pylint: disable-next=too-many-branches, too-many-statements
+    # pylint: disable-next=too-many-branches, too-many-statements, too-many-locals
```
fd0r, line 369: 🙈

src/concrete/ml/sklearn/linear_model.py
fd0r: That's indeed probably a better solution than the 2 method one we had before!

src/concrete/ml/sklearn/linear_model.py
fd0r: Good catch!

fd0r (approved):

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: 🙉

tests/sklearn/test_fhe_training.py
RomanBredehoft: just changed and made passing the RNG object between calls possible indeed, the test is green , thanks !

fd0r (approved):

PR #624, opened 2024-04-18
chore: add check proper keyword for rounding_threshold_bits

jfrery:


RomanBredehoft (approved):
great !

fd0r (approved):

src/concrete/ml/common/utils.py
```diff
@@ -632,6 +633,11 @@ def process_rounding_threshold_bits(rounding_threshold_bits):
         if isinstance(rounding_threshold_bits, int):
             n_bits_rounding = rounding_threshold_bits
         elif isinstance(rounding_threshold_bits, dict):
+            valid_keys = {"n_bits", "method"}
+            if not valid_keys.issuperset(rounding_threshold_bits.keys()):
+                raise KeyError(
+                    f"Invalid keys in rounding_threshold_bits. Allowed keys are {valid_keys}."
+                )
```
fd0r, line 640: We should probably add a check after the "auto" detection that the n_bits_rounding is indeed an integer or at least can be casted to it.

fd0r (approved):

src/concrete/ml/common/utils.py
jfrery: Yes I did it in  https://github.com/zama-ai/concrete-ml/pull/625

PR #625, opened 2024-04-18
chore: rename n bits in rounding threshold bits plus checks

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4398


fd0r (approved):
I'm fine with it but this is a breaking change no?

It doesn't really appear like it but it is, even if this is a relatively new attribute.

jfrery:
> I'm fine with it but this is a breaking change no?
> 
> It doesn't really appear like it but it is, even if this is a relatively new attribute.

Hm yes you are right :/ annoying.

jfrery:
Added it in #624

PR #631, opened 2024-04-19
chore: remove sphinx doc build

andrei-stoian-zama:
Removes sphinx doc alltogether

Closes https://github.com/zama-ai/concrete-ml-internal/issues/4394


Makefile
```diff
@@ -401,46 +401,10 @@ docker_clean_volumes:
...
-	@# Docs
-	cd docs-copy && poetry run "$(MAKE)" html SPHINXOPTS='-W --keep-going'
-	@# Copy images from GitBook
-	cp docs/.gitbook/assets/*.png docs-copy/_build/html/_images
-	cp -r use_case_examples docs-copy/_build/
-	cp -r docs-copy/_build docs/
-	rm -rf docs-copy
-
 .PHONY: docs # Build docs and check links
-docs: docs_no_links
```
RomanBredehoft, line 443: could you keep `docs_no_links` target ? so that `make docs` is `docs_no_links` + `check_links` like before ? `docs_no_links` is used in the CI in order to avoid checking links when releasing, as it is sometimes flaky (and independent from us)

Makefile
```diff
@@ -775,19 +729,14 @@ check_links:
 	@#  --ignore-url=https://github.com/zama-ai/concrete-ml-internal/issues: because issues are
 	@#		private
 	@#	--ignore-url=.gitbook/assets : some gitbook functionalities use links to images to include
```
RomanBredehoft, line 731: you removed `--ignore-url=.gitbook/assets` but you kept a comment about it  

same for `--ignore-url=_static/webpack-macros.html` above

RomanBredehoft (changes requested):
thanks !  I have a few comments : 
- better to keep `docs no_no_links` for the release
- you can also remove the dependencies related to sphinx in the toml: sphinx, nbsphinx, sphinx-rtd-theme, sphinx_copy, myst_parser, sphinx_copybutton
- there are some remaining things to be removed I think : `docs/make.bat`

Makefile
andrei-stoian-zama: done

RomanBredehoft (changes requested):
since you've changed the toml, could you update the licenses ? you could simply run the update license action (https://github.com/zama-ai/concrete-ml/actions) and target your branch

RomanBredehoft (approved):

PR #639, opened 2024-04-21
Update README.md

Budalebah:
added https://www.python.org/downloads/ link


andrei-stoian-zama:
Thanks for this suggestion, but I don't think it brings any interesting new information to the README.

PR #640, opened 2024-04-22
fix: bad symlink and add checks

andrei-stoian-zama:
With many thanks to https://github.com/sukrucildirr for finding the bad symlink


Makefile
```diff
@@ -844,3 +844,12 @@ run_all_use_case_examples:
 .PHONY: check_utils_use_case # Check that no utils.py are found in use_case_examples
 check_utils_use_case:
 	./script/make_utils/check_utils_in_use_case.sh
+
+.PHONY: check_symlinks # Check that no utils.py are found in use_case_examples
+check_symlinks:
```
bcm-at-zama, line 849: needs to be added in the CI or as a sub target of something we already launch, no?

RomanBredehoft: could be done in the `- name: Check links` step

RomanBredehoft (dismissed):

RomanBredehoft (approved):

PR #641, opened 2024-04-23
fix: identify ConstantOfShape ONNX nodes as constant values

RomanBredehoft:
this PR is in draft until we find a proper solution, as the initial suggestion is not right (`ConstantOfShape` are not real constants, they have an input coming in)

a model that contains a "ConstantOfShape" node can trigger an obscure `AttributeError: 'RawOpOutput' object has no attribute 'quantizer'` in the `QuantizedAdd` operator

closes https://github.com/zama-ai/concrete-ml-internal/issues/4406

### DIFF
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1564,3 +1564,35 @@ class Conv1dModel(nn.Module):
         x = self.bn1(x)
         x = self.fc1(x)
         return x
+
+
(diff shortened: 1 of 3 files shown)


andrei-stoian-zama (approved):
Nicely done

RomanBredehoft:
unfortunately, the issue started to be more complex than I initially thought, and the solution provided does not entirely fix the issue : https://github.com/zama-ai/concrete-ml-internal/issues/4413

PR #651, opened 2024-04-25
fix: update qgpt2_class.py to fix typo

fd0r:
This commit fixes a bug in use_cases_examples/llm/qgpt2_class.py demo where `simulate` was not called properly.

It raised `TypeError: unsupported operand type(s) for /: 'method' and 'int'` because `<bound method Circuit.simulate>` was not called.

### DIFF
--- a/use_case_examples/llm/qgpt2_class.py
+++ b/use_case_examples/llm/qgpt2_class.py
@@ -129,7 +129,7 @@ class QuantizedModel:
                 q_x = np.expand_dims(q_x, axis=0)
 
                 if fhe == "simulate":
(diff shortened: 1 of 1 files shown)


RomanBredehoft (approved):

RomanBredehoft (changes requested):
we can merge it but CLA is not happy, so let's make sure everything is right

RomanBredehoft (approved):

fd0r:
Only waiting for @HugoLB0 to sign the CLA.

aquint-zama:
@cla-bot check

PR #659, opened 2024-04-29
chore: add check initialization variables cml models vs sklearn

jfrery:
Useful to keep consistency with sklearn / xgboost

### DIFF
--- a/src/concrete/ml/sklearn/rf.py
+++ b/src/concrete/ml/sklearn/rf.py
@@ -173,7 +173,7 @@ class RandomForestRegressor(BaseTreeRegressorMixin):
         min_samples_split=2,
         min_samples_leaf=1,
         min_weight_fraction_leaf=0.0,
-        max_features="sqrt",
+        max_features=1.0,
         max_leaf_nodes=None,
         min_impurity_decrease=0.0,
         bootstrap=True,
--- a/src/concrete/ml/sklearn/xgb.py
+++ b/src/concrete/ml/sklearn/xgb.py
@@ -33,7 +33,7 @@ class XGBClassifier(BaseTreeClassifierMixin):
         max_depth: Optional[int] = 3,
         learning_rate: Optional[float] = None,
         n_estimators: Optional[int] = 20,
-        objective: Optional[str] = "binary:logistic",
+        objective: Optional[str] = None,
         booster: Optional[str] = None,
         tree_method: Optional[str] = None,
         n_jobs: Optional[int] = None,
@@ -278,7 +278,7 @@ class XGBRegressor(BaseTreeRegressorMixin):
         max_depth: Optional[int] = 3,
         learning_rate: Optional[float] = None,
         n_estimators: Optional[int] = 20,
-        objective: Optional[str] = "reg:squarederror",
+        objective: Optional[str] = None,
         booster: Optional[str] = None,
         tree_method: Optional[str] = None,
         n_jobs: Optional[int] = None,
@@ -450,10 +450,14 @@ class XGBRegressor(BaseTreeRegressorMixin):
         metadata["max_cat_to_onehot"] = self.max_cat_to_onehot
         metadata["grow_policy"] = self.grow_policy
         metadata["sampling_method"] = self.sampling_method
-        metadata["callbacks"] = self.callbacks
         metadata["early_stopping_rounds"] = self.early_stopping_rounds
         metadata["eval_metric"] = self.eval_metric
-        metadata["kwargs"] = self.kwargs
+
+        # Callables are not serializable
+        assert not self.kwargs, "kwargs are not supported for serialization"
+        assert not self.callbacks, "callbacks are not supported for serialization"
+        metadata["kwargs"] = None
+        metadata["callbacks"] = None
 
         return metadata
 
--- a/src/concrete/ml/sklearn/xgb.py
+++ b/src/concrete/ml/sklearn/xgb.py
@@ -2,10 +2,11 @@
 
 import platform
 import warnings
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Union
 
 import numpy
 import xgboost.sklearn
+from numpy.random import RandomState
 from xgboost.callback import TrainingCallback
 
 from ..common.debugging.custom_assert import assert_true
@@ -33,7 +34,7 @@ class XGBClassifier(BaseTreeClassifierMixin):
         max_depth: Optional[int] = 3,
         learning_rate: Optional[float] = None,
         n_estimators: Optional[int] = 20,
-        objective: Optional[str] = None,
+        objective: Optional[str] = "binary:logistic",
         booster: Optional[str] = None,
         tree_method: Optional[str] = None,
         n_jobs: Optional[int] = None,
@@ -51,7 +52,7 @@ class XGBClassifier(BaseTreeClassifierMixin):
         missing: float = numpy.nan,
         num_parallel_tree: Optional[int] = None,
         monotone_constraints: Optional[Union[Dict[str, int], str]] = None,
-        interaction_constraints: Optional[Union[str, List[Tuple[str]]]] = None,
+        interaction_constraints: Optional[Union[str, Sequence[Sequence[str]]]] = None,
         importance_type: Optional[str] = None,
         gpu_id: Optional[int] = None,
         validate_parameters: Optional[bool] = None,
@@ -180,14 +181,24 @@ class XGBClassifier(BaseTreeClassifierMixin):
         metadata["random_state"] = self.random_state
         metadata["verbosity"] = self.verbosity
         metadata["max_bin"] = self.max_bin
-        metadata["callbacks"] = self.callbacks
         metadata["early_stopping_rounds"] = self.early_stopping_rounds
         metadata["max_leaves"] = self.max_leaves
-        metadata["eval_metric"] = self.eval_metric
         metadata["max_cat_to_onehot"] = self.max_cat_to_onehot
         metadata["grow_policy"] = self.grow_policy
         metadata["sampling_method"] = self.sampling_method
-        metadata["kwargs"] = self.kwargs
+
+        if callable(self.eval_metric):
+            raise NotImplementedError("Callable eval_metric is not supported for serialization")
+
+        if self.kwargs:
+            raise NotImplementedError("kwargs are not supported for serialization")
+
+        if self.callbacks:
+            raise NotImplementedError("callbacks are not supported for serialization")
+
+        metadata["eval_metric"] = self.eval_metric
+        metadata["kwargs"] = None
+        metadata["callbacks"] = None
 
         return metadata
 
@@ -277,8 +288,8 @@ class XGBRegressor(BaseTreeRegressorMixin):
         n_bits: Union[int, Dict[str, int]] = 6,
         max_depth: Optional[int] = 3,
         learning_rate: Optional[float] = None,
-        n_estimators: Optional[int] = 20,
-        objective: Optional[str] = None,
+        n_estimators: int = 20,
+        objective: Optional[str] = "reg:squarederror",
         booster: Optional[str] = None,
         tree_method: Optional[str] = None,
         n_jobs: Optional[int] = None,
@@ -296,23 +307,23 @@ class XGBRegressor(BaseTreeRegressorMixin):
         missing: float = numpy.nan,
         num_parallel_tree: Optional[int] = None,
         monotone_constraints: Optional[Union[Dict[str, int], str]] = None,
-        interaction_constraints: Optional[Union[str, List[Tuple[str]]]] = None,
+        interaction_constraints: Optional[Union[str, Sequence[Sequence[str]]]] = None,
         importance_type: Optional[str] = None,
         gpu_id: Optional[int] = None,
         validate_parameters: Optional[bool] = None,
         predictor: Optional[str] = None,
         enable_categorical: bool = False,
-        random_state: Optional[int] = None,
+        random_state: Optional[Union[RandomState, int]] = None,
         verbosity: Optional[int] = None,
-        eval_metric: Optional[str] = None,
+        eval_metric: Optional[Union[str, List[str], Callable]] = None,
         sampling_method: Optional[str] = None,
         max_leaves: Optional[int] = None,
         max_bin: Optional[int] = None,
         max_cat_to_onehot: Optional[int] = None,
         grow_policy: Optional[str] = None,
-        callbacks: Optional[List[Callable]] = None,
+        callbacks: Optional[List[TrainingCallback]] = None,
         early_stopping_rounds: Optional[int] = None,
-        **kwargs,
+        **kwargs: Any,
     ):
         # base_score != 0.5 or None does not seem to not pass our tests
         # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/474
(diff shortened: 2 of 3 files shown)


src/concrete/ml/sklearn/xgb.py
```diff
@@ -159,6 +179,15 @@ def dump_dict(self) -> Dict[str, Any]:
         metadata["use_label_encoder"] = self.use_label_encoder
         metadata["random_state"] = self.random_state
         metadata["verbosity"] = self.verbosity
+        metadata["max_bin"] = self.max_bin
+        metadata["callbacks"] = self.callbacks
```
RomanBredehoft, line 183: great, but I'm not sure we support some of these parameters when it comes to serialization. for example for QNNs we have this :

```
        # skorch attributes that cannot be serialized
        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3550
        # Disable mypy as running isinstance with a Callable type unexpectedly raises an issue:
        # https://github.com/python/mypy/issues/3060
        if isinstance(self.train_split, Callable) and not isinstance(  # type: ignore[arg-type]
            self.train_split, ValidSplit
        ):
            raise NotImplementedError(
                "Serializing a custom Callable object is not secure and is therefore disabled. "
                "Please set `train_split` to either None or a ValidSplit instance."
            )

        if self.callbacks != "disable":
            raise NotImplementedError(
                "Serializing a custom Callable object is not secure and is therefore disabled. "
                "Additionally, the serialization of skorch's different callback classes is not "
                f"supported. Please set `callbacks` to 'disable'. Got {self.callbacks}."
            )
```

src/concrete/ml/sklearn/xgb.py
```diff
@@ -318,9 +363,17 @@ def __init__(
         self.validate_parameters = validate_parameters
         self.predictor = predictor
         self.enable_categorical = enable_categorical
-        self.use_label_encoder = use_label_encoder
```
RomanBredehoft, line 321: is xgboost not supporting this ? does xgboost never fail when given a new but unsupported attribute at instantiation ? and why did we even have that ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -2032,3 +2033,63 @@ def test_error_raise_unsupported_pandas_values(model_class, bad_value, expected_
...
+
+    # Accumulate parameters from all base classes of the sklearn model
+    def get_params(cls):
+        params = set(inspect.signature(cls).parameters.keys())
+        for base in cls.__bases__:
+            params.update(get_params(base))
+        return params
+
+    # Conditionally gather parameters from base classes for XGBClassifier and XGBRegressor
+    if model.sklearn_model.__class__.__name__ in ["XGBClassifier", "XGBRegressor"]:
```
RomanBredehoft, line 2073: why are they  handled differently ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -2032,3 +2033,63 @@ def test_error_raise_unsupported_pandas_values(model_class, bad_value, expected_
...
+        expected_difference.add("fit_encrypted")
+        expected_difference.add("parameters_range")
+
+    # Calculate differences
+    missing_params = sklearn_params - cml_params
+    extra_params = (cml_params - sklearn_params) - expected_difference
+
+    assert (
+        not missing_params
+    ), f"Concrete ML {model.__class__.__name__} is missing these init parameters: {missing_params}"
```
RomanBredehoft, line 2092: btw instead of having to this all the time, we have utility functions : 
```
from ..common.utils import get_model_class, get_model_name
```

to use like `model_name = get_model_name`. the benefits are that it also handles partial models and uninstantiated models

tests/sklearn/test_sklearn_models.py
```diff
@@ -2032,3 +2033,63 @@ def test_error_raise_unsupported_pandas_values(model_class, bad_value, expected_
...
+
+@pytest.mark.parametrize(
+    "model_class, parameters",
+    get_sklearn_linear_models_and_datasets()
+    + get_sklearn_tree_models_and_datasets()
+    + get_sklearn_neighbors_models_and_datasets(),
+)
+def test_initialization_variables_match(model_class, parameters, load_data, is_weekly_option):
+    """Test that CML models can be initialized with the same parameters scikit-learn models."""
+    n_bits = N_BITS_THRESHOLD_FOR_SKLEARN_CORRECTNESS_TESTS
```
RomanBredehoft, line 2046: we usually use  `min(N_BITS_REGULAR_BUILDS)` although we actually have `get_n_bits_non_correctness(model_class)` , we don't really care about n_bits here anyway

tests/sklearn/test_sklearn_models.py
```diff
@@ -2032,3 +2033,63 @@ def test_error_raise_unsupported_pandas_values(model_class, bad_value, expected_
...
         model.fit(x_train, y_train)
+
+
+@pytest.mark.parametrize(
+    "model_class, parameters",
+    get_sklearn_linear_models_and_datasets()
+    + get_sklearn_tree_models_and_datasets()
+    + get_sklearn_neighbors_models_and_datasets(),
+)
+def test_initialization_variables_match(model_class, parameters, load_data, is_weekly_option):
```
RomanBredehoft, line 2044: should we also check the default values btw ? I know that we don't always have the same default values so this could be tricky to do, but then maybe add a fixme for that ?

RomanBredehoft (changes requested):
great ! I have a few questions though

src/concrete/ml/sklearn/xgb.py
```diff
@@ -159,6 +179,15 @@ def dump_dict(self) -> Dict[str, Any]:
...
         metadata["verbosity"] = self.verbosity
+        metadata["max_bin"] = self.max_bin
+        metadata["callbacks"] = self.callbacks
+        metadata["early_stopping_rounds"] = self.early_stopping_rounds
+        metadata["max_leaves"] = self.max_leaves
+        metadata["eval_metric"] = self.eval_metric
+        metadata["max_cat_to_onehot"] = self.max_cat_to_onehot
+        metadata["grow_policy"] = self.grow_policy
+        metadata["sampling_method"] = self.sampling_method
+        metadata["kwargs"] = self.kwargs
```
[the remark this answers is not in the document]
jfrery: lol I am curious to see if we have a test on this. I guess it won't fail as we never bench with kwargs. I will remove it from the serialization process.

src/concrete/ml/sklearn/xgb.py
jfrery: It was marked as deprecated and is now removed as far as I can see.

tests/sklearn/test_sklearn_models.py
jfrery: xgboost sklearn api doesn't really follow sklearn on parameter initialization. They have them in the parent class instead.

tests/sklearn/test_sklearn_models.py
jfrery: Ok I can use it.

tests/sklearn/test_sklearn_models.py
```diff
@@ -2032,3 +2033,63 @@ def test_error_raise_unsupported_pandas_values(model_class, bad_value, expected_
 
     with pytest.raises(ValueError, match=expected_error):
         model.fit(x_train, y_train)
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
jfrery: Good question. Not sure.

tests/sklearn/test_sklearn_models.py
jfrery: Good point! I think we should.

src/concrete/ml/sklearn/xgb.py
RomanBredehoft: but you did not update xgb in this PR 🤔  do you mean it's already deprecated in the current version we're using ?

src/concrete/ml/sklearn/xgb.py
jfrery: Hmm don't we have a test that checks we can serailize and desarialize all parameters? Tests seem to pass.

src/concrete/ml/sklearn/xgb.py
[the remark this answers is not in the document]
jfrery: Removed.

src/concrete/ml/sklearn/xgb.py
jfrery: AH yes I got confused. So: in this PR I remove use_label_encoder from XGBRegressor because it's just not part of the xgboost.sklearn model. Must have been a copy past from the classifier.

But in the other internal PR the new xgboost version has

```python
        if use_label_encoder is True:
            raise ValueError("Label encoder was removed in 1.6.")
```

so it will disappear from both eventually.

tests/sklearn/test_sklearn_models.py
jfrery: Ok I can use the min(N_BITS_REGULAR_BUILDS).

src/concrete/ml/sklearn/xgb.py
RomanBredehoft: yes but it uses the default values, and we are able to serialize a None 🙂  here I'm talking about serializing exotic stuff like `TrainingCallback` (for `self.callbacks`) or callable (ex: `self. eval_metric`) objects that a user could put when initializing

tests/sklearn/test_sklearn_models.py
RomanBredehoft: I think we should better use `get_n_bits_non_correctness(model_class)` as it does `min(N_BITS_REGULAR_BUILDS)` but handles KNN differently, but I won't insist too much here because it seems we usually do `min(N_BITS_REGULAR_BUILDS)` anyway

src/concrete/ml/sklearn/xgb.py
jfrery: I see :/ I will assert potential callable parameter are None. eval_metric isn't callable it's a string though, right?

src/concrete/ml/sklearn/xgb.py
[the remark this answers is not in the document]
jfrery: Actually I assert it's none.

tests/sklearn/test_sklearn_models.py
jfrery: You are right we should use the utils function and adapt them to our needs. We should probably make sure we use them everywhere.

src/concrete/ml/sklearn/xgb.py
RomanBredehoft: better to have a value error raised + tested like we do for QNNs ! putting asserts for user inputs is not good a good practice (refs https://github.com/zama-ai/concrete-ml-internal/issues/4232) 

besides about `eval_metric`, I can see that you've put `eval_metric: Optional[Union[str, List[str], Callable]] = None,` , so I imagine it can also be a `callable`

src/concrete/ml/sklearn/rf.py
```diff
@@ -173,7 +173,7 @@ def __init__(
         min_samples_split=2,
         min_samples_leaf=1,
         min_weight_fraction_leaf=0.0,
-        max_features="sqrt",
```
RomanBredehoft, line 176: 😮

src/concrete/ml/sklearn/xgb.py
```diff
@@ -32,7 +33,7 @@ def __init__(
         max_depth: Optional[int] = 3,
         learning_rate: Optional[float] = None,
         n_estimators: Optional[int] = 20,
-        objective: Optional[str] = "binary:logistic",
```
RomanBredehoft, line 35: we should check that this does not change the output and thus the way we do post-processing though 🤔

src/concrete/ml/sklearn/xgb.py
```diff
@@ -241,7 +278,7 @@ def __init__(
         max_depth: Optional[int] = 3,
         learning_rate: Optional[float] = None,
         n_estimators: Optional[int] = 20,
-        objective: Optional[str] = "reg:squarederror",
```
RomanBredehoft, line 244: same

src/concrete/ml/sklearn/xgb.py
```diff
@@ -391,6 +445,19 @@ def dump_dict(self) -> Dict[str, Any]:
...
+        metadata["max_bin"] = self.max_bin
+        metadata["max_leaves"] = self.max_leaves
+        metadata["max_cat_to_onehot"] = self.max_cat_to_onehot
+        metadata["grow_policy"] = self.grow_policy
+        metadata["sampling_method"] = self.sampling_method
+        metadata["early_stopping_rounds"] = self.early_stopping_rounds
+        metadata["eval_metric"] = self.eval_metric
+
+        # Callables are not serializable
+        assert not self.kwargs, "kwargs are not supported for serialization"
```
RomanBredehoft, line 457: cf comment above, we should start having valuerror + tests instead of assert with user inputs (refs https://github.com/zama-ai/concrete-ml-internal/issues/4232)

tests/sklearn/test_sklearn_models.py
jfrery: Done

src/concrete/ml/sklearn/xgb.py
jfrery: let's see what the tests say

src/concrete/ml/sklearn/rf.py
RomanBredehoft: more than that I think this was a mistake no ? why would the default value for `max_features` be `sqrt` ? 😅

jfrery: No it's actually a proper parameter ^^. It is the default for the classifier but not the regressor.

src/concrete/ml/sklearn/xgb.py
jfrery: > besides about eval_metric, I can see that you've put eval_metric: Optional[Union[str, List[str], Callable]] = None, , so I imagine it can also be a callable

Hmm I see. I just copy pasted the one from XGBoostClassifier. Looks like the XGBoostRegressor type is wrong then.

Maybe we should also check the types of the params. It's starting to be quite heavy lol

jfrery: I added the raising of errors.

src/concrete/ml/sklearn/xgb.py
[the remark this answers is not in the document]
jfrery: Raising error now.

src/concrete/ml/sklearn/xgb.py
jfrery: Actually their was a bug in how I grabbed the default param. I added back the objective. It was correct.

src/concrete/ml/sklearn/xgb.py
jfrery: fixed

src/concrete/ml/sklearn/xgb.py
jfrery: Done

src/concrete/ml/sklearn/rf.py
RomanBredehoft: weird naming then haha but ok good to know !

RomanBredehoft (approved):
thanks a lot for this !!

RomanBredehoft (approved):

PR #660, opened 2024-04-29
feat: use composition for non-interactive encrypted training

RomanBredehoft:
it's still a WIP as we can observe some mismatches between disable and simulate

closes https://github.com/zama-ai/concrete-ml-internal/issues/4374


src/concrete/ml/quantization/quantized_module.py
```diff
@@ -483,12 +487,22 @@ def _clear_forward(
...
 
-        results = tuple(
+        q_results = tuple(
             elt.qvalues for elt in output_quantized_arrays if isinstance(elt, QuantizedArray)
         )
-        if len(results) == 1:
-            return results[0]
-        return results
+
+        if self._composition_mapping is not None:
```
RomanBredehoft, line 494: this part adds the requant in FHE

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -347,6 +355,13 @@ def _get_training_quantized_module(
             fit_bias=self.fit_intercept,
         )
 
+        # Enable the underlying FHE circuit to be composed with itself
+        # This feature is used in order to be able to iterate in the clear n times without having
+        # to encrypt/decrypt the weight/bias values between each loop
+        configuration = Configuration(composable=True)
+
+        composition_mapping = {0: 2, 1: 3}
```
RomanBredehoft, line 363: don't love this part but I don't see how to properly indicate which input quantizers should be used for which input quantizers

src/concrete/ml/quantization/quantized_module.py
jfrery: Unless I miss something important on how the composition works, I don't think we should touch the quantized module.

We could probably add a quantized op in the graph that is created in the post_training.py at calibration. Maybe a QuantizedRequant that uses the model input quantizer as its output quantizer.

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -483,12 +487,22 @@ def _clear_forward(
...
+            q_results = tuple(
+                self.input_quantizers[input_i].quant(
+                    self.output_quantizers[output_i].dequant(q_results[output_i])
+                )
+                for output_i, input_i in self._composition_mapping.items()
+            )
+
+            if len(q_results) == 1:
```
andrei-stoian-zama, line 502: can you assert on the shape here ? the input/output shapes should match

andrei-stoian-zama (changes requested):

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: so that's an interesting point. I took some time to check if output order is deterministic or not in ONNX, but I couldn't find a definitive answer. It looks like it is (for example, `torch.onnx.export` has a parameter `output_names` that asks for an ordered list of output names), but still, there nothing clear about it in the documentation (ex: https://discuss.pytorch.org/t/are-there-any-guarantees-on-the-output-ordering-of-onnx-model-beyond-topological-sort/151920) 

however, I don't know how we could actually check that + I believe we've always assumed it is, so yes I'll add some checks for the shape but maybe not more than that

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: I've added the checks in the new `_add_requant_for_composition` method (name to be confirmed)

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: see my answer here : https://github.com/zama-ai/concrete-ml/pull/660#discussion_r1624431458

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -290,6 +293,61 @@ def _set_output_quantizers(self) -> List[UniformQuantizer]:
         )
         return output_quantizers
 
+    # Remove this once we handle the re-quantization step in post-training only
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4472
+    def _add_requant_for_composition(self, composition_mapping: Optional[Dict]):
```
RomanBredehoft, line 298: new (private) method for quantized module

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -290,6 +293,61 @@ def _set_output_quantizers(self) -> List[UniformQuantizer]:
...
+        if not isinstance(composition_mapping, Dict):
+            raise ValueError(
+                "Parameter 'composition_mapping' mus be a dictionary. Got "
+                f"{type(composition_mapping)}"
+            )
+
+        max_output_pos = len(self.output_quantizers) - 1
+        max_input_pos = len(self.input_quantizers) - 1
+
+        for output_position, input_position in composition_mapping.items():
```
RomanBredehoft, line 322: make sure the mapping is of the form `{0:1, 3:2}`

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -561,30 +631,52 @@ def _fhe_forward(
...
             "properly initialized or the input data has been changed since its initialization.",
             ValueError,
         )
 
-        q_x = tuple(self.input_quantizers[idx].quant(x[idx]) for idx in range(len(x)))
+        assert not all(x_i is None for x_i in x), "Please provide at least one input to quantize."
+
+        # Ignore [arg-type] check from mypy as it is not able to see that the input to `quant`
+        # cannot be None
+        q_x = tuple(
```
RomanBredehoft, line 660: these are needed to match how CP works with encrypt, ie `encrypt(None, x) = None, x_enc`, since we do not encrypt all inputs at the same time with composition

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -493,120 +516,156 @@ def _fit_encrypted(
...
+
+            else:
+                X_batch_enc, y_batch_enc = q_X_batch, q_y_batch
+
+            X_batches_enc.append(X_batch_enc)
+            y_batches_enc.append(y_batch_enc)
+
+        # Similarly, we only quantize the weight and bias values using the third and fourth
+        # position parameter
+        _, _, q_weights, q_bias = self.training_quantized_module.quantize_input(
```
RomanBredehoft, line 605: see https://github.com/zama-ai/concrete-ml/pull/660/files#r1624776488

src/concrete/ml/torch/compile.py
```diff
@@ -181,16 +184,32 @@ def _compile_torch_or_onnx_model(
             for each input. By default all arguments will be encrypted.
         reduce_sum_copy (bool): if the inputs of QuantizedReduceSum should be copied to avoid
             bit-width propagation
+        composition_mapping (Optional[Dict]): Dictionary that maps output positions with input
```
RomanBredehoft, line 187: adding this new parameter to the private funct `_compile_torch_or_onnx_model` instead of the other public ones to keep things internal

src/concrete/ml/torch/compile.py
```diff
@@ -219,6 +238,13 @@ def _compile_torch_or_onnx_model(
     # Find the right way to set parameters for compiler, depending on the way we want to default
     p_error, global_p_error = manage_parameters_for_pbs_errors(p_error, global_p_error)
 
+    # If a mapping between input and output quantizers is set, add a re-quantization step at the
+    # end of the forward call. This is only useful for composable circuits in order to make sure
+    # that input and output quantizers match
+    if composition_mapping is not None:
```
RomanBredehoft, line 244: this is where we decide to add the requant step or not

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -493,102 +556,157 @@ def _fit_encrypted(
...
 
-            # If early stopping is enabled, compute the loss and stop the training if it gets under
-            # the given tolerance
+            # If early stopping is enabled, decrypt (if needed) and de-quantize the weight and bias
+            # values. Then, compute the loss and stop the training if it gets under the given
+            # tolerance
             # Additionally, there is no point in computing the following in case of a partial fit,
             # as it only represents a single iteration
             if self.early_stopping and not is_partial_fit:
+                weights_float, bias_float = self._decrypt_dequantize_training_output(
```
RomanBredehoft, line 704: we keep early possible possible with composition by adding this decrypt/dequant step here (since this is only for development, we believe it's not really an issue to do that)

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -124,37 +124,68 @@ def load(self):
 
         self.server = fhe.Server.load(Path(self.path_dir).joinpath("server.zip"))
 
+    # We should make 'serialized_encrypted_quantized_data' handle unpacked inputs, as Concrete does,
+    # instead of tuples
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4477
+    # We should also rename the input arguments to remove the `serialized` part, as we now accept
+    # both serialized and deserialized input values
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4476
     def run(
```
RomanBredehoft, line 133: we now allow serialized and deserialized inputs (avoids having to deser + ser at each server call with composition)

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -357,97 +388,78 @@ def get_serialized_evaluation_keys(self) -> bytes:
...
 
-        # Return a single value if the original input was a single value
-        return serialized_enc_qx[0] if len(serialized_enc_qx) == 1 else serialized_enc_qx
+        return q_x_enc_serialized
 
+    # We should find a better name for `serialized_encrypted_quantized_result`
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4476
     def deserialize_decrypt(
-        self, serialized_encrypted_quantized_result: Union[bytes, Tuple[bytes, ...]]
+        self, *serialized_encrypted_quantized_result: Optional[bytes]
```
RomanBredehoft, line 416: same

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -357,97 +388,78 @@ def get_serialized_evaluation_keys(self) -> bytes:
...
 
-        return deserialized_decrypted_quantized_result
+        return q_result
 
+    # We should find a better name for `serialized_encrypted_quantized_result`
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4476
     def deserialize_decrypt_dequantize(
-        self, serialized_encrypted_quantized_result: Union[bytes, Tuple[bytes, ...]]
-    ) -> numpy.ndarray:
+        self, *serialized_encrypted_quantized_result: Optional[bytes]
```
RomanBredehoft, line 440: same

src/concrete/ml/common/utils.py
```diff
@@ -668,3 +682,50 @@ def process_rounding_threshold_bits(rounding_threshold_bits):
         rounding_threshold_bits = {"n_bits": n_bits_rounding, "method": method}
 
     return rounding_threshold_bits
+
+
+def serialize_encrypted_values(
```
jfrery, line 687: Do we need these outside of deployment? If not I would keep these in the deployment directory.

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -124,37 +124,68 @@ def load(self):
...
-            if isinstance(result, tuple)
-            else result.serialize()
-        )
+        # Make sure inputs are either only serialized values or encrypted values
+        assert (
+            inputs_are_serialized ^ inputs_are_encrypted_values
+        ), "Inputs must be all of the same types, either 'bytes' or 'concrete.fhe.Value'"
+
+        # Deserialize the values if they are all serialized
+        if inputs_are_serialized:
```
jfrery, line 172: nice thanks!

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -124,37 +124,68 @@ def load(self):
...
+
+        q_result_enc = self.server.run(*q_data_enc, evaluation_keys=evaluation_keys)
+
+        # If inputs were serialized, return serialized values as well
+        if inputs_are_serialized:
+            q_result_enc = serialize_encrypted_values(*to_tuple(q_result_enc))
+
+        # Mypy complains because the outputs of `serialize_encrypted_values` can be None, but here
+        # we already made sure this is not the case
+        return q_result_enc  # type: ignore[return-value]
```
jfrery, line 188: We usually do `assert q_result_enc is not None` for mypy instead of having type: ignore

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -357,97 +388,78 @@ def get_serialized_evaluation_keys(self) -> bytes:
...
         # Serialize the encrypted values to be sent to the server
-        serialized_enc_qx = tuple(e.serialize() for e in enc_qx)
+        q_x_enc_serialized = serialize_encrypted_values(*q_x_enc)
 
-        # Return a single value if the original input was a single value
-        return serialized_enc_qx[0] if len(serialized_enc_qx) == 1 else serialized_enc_qx
+        return q_x_enc_serialized
 
+    # We should find a better name for `serialized_encrypted_quantized_result`
+    # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4476
```
jfrery, line 414: Did you put the fixme in things we need to change for 2.0?

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -357,97 +388,78 @@ def get_serialized_evaluation_keys(self) -> bytes:
...
-            *deserialized_decrypted_dequantized_result
-        )
-
-        return deserialized_decrypted_dequantized_result
+        # Side note: `post_processing` method from built-in models (not Quantized Modules) only
+        # handles a single input. Calling the following is however not an issue for now as we expect
+        # 'f_result' to be a tuple of length 1 in this case anyway. Still, we need to make sure this
+        # does not break in the future if any built-in models starts to handle multiple outputs :
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4474
+        result = self.model.post_processing(*f_result)
```
jfrery, line 463: Can't we have an assert maybe? Either here or in built in post processing?

jfrery: I would still write explicitely in the variable name the difference between q_result f_result and result. I don't find these explicit enough.

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -730,8 +824,15 @@ def compile(
...
+        # Make sure all inputs are quantized to int64 and are not None
+        assert all_values_are_of_dtype(
+            *to_tuple(q_inputs), dtypes="int64", allow_none=False
+        ), "Inputs were not quantized to int64"
+
         # Generate the input-set with proper dimensions
-        inputset = _get_inputset_generator(q_inputs)
+        # Ignore [arg-type] check from mypy as it is not able to see that no values in `q_inputs`
+        # is None
+        inputset = _get_inputset_generator(q_inputs)  # type: ignore[arg-type]
```
jfrery, line 836: Here also we should probably just do

```python
assert all(item is not None for item in q_inputs)
```

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -493,102 +556,157 @@ def _fit_encrypted(
...
+        if self.verbose and (not is_partial_fit or self.training_quantized_module is None):
             mode_string = " (simulation)" if fhe == "simulate" else ""
             print(f"Training on encrypted data{mode_string}...")
 
         # A partial fit is similar to running a fit with a single iteration
         max_iter = 1 if is_partial_fit else self.max_iter
 
-        # Iterate on the training quantized module in the clear
-        for iteration_step in range(max_iter):
+        # Quantize and encrypt the batches
```
jfrery, line 606: Not sure about the comment change? Quantization happen further down. The original comment was more on point.

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -331,8 +331,9 @@ def remote_call(self, x: torch.Tensor) -> torch.Tensor:  # pragma:no cover
             encrypted_result = inference_query.content
             decrypted_prediction = client.deserialize_decrypt_dequantize(encrypted_result)[0]
             inferences.append(decrypted_prediction)
+
         # Concatenate results and move them back to proper device
-        return torch.Tensor(numpy.array(inferences)).to(device=base_device)
+        return torch.Tensor(inferences).to(device=base_device)
```
jfrery, line 336: What's happening here?

.gitignore
```diff
@@ -167,6 +168,8 @@ execution_time_of_individual_pytest_files.txt
 # Docs: Advance Examples MNIST data
 docs/advanced_examples/data/MNIST/
 
+# Docs: Advance Examples FHE training deployment files
+docs/advanced_examples/fhe_training
```
jfrery, line 172: Maybe we should target a tmp folder instead. This kind of practice leaves tons of circuit behind

jfrery (changes requested):
Well done! Wasn't easy to integrate that composition.

I have some comments + 1 question:
- is the composition set by default in sgdclassifier? If so are we sure the new PBS dequant requant isn't impacting the convergence?

jfrery (changes requested):
Thanks for this. Wasn't easy to integrate composition!

I have a few comment + a question:

- is the composition enabled by default in sgd fhe training? If so, are we sure the added PBS isn't impacting convergence?

.gitignore
RomanBredehoft: that's what I did, but if you don't run the full notebook until the cleanup part then some file stays

but I was planning to add a gitignore in the advances_examples directory instead, and put this there

src/concrete/ml/common/utils.py
RomanBredehoft: well, currently I'm using them in the notebook on the server side to serialize/deserialize values outside of the `FheModelServer.run`, but now that I'm thinking of it, it's not mandatory because they are not tuples. Instead, they can directly call concrete's method (`.serialize()` or `fhe.Value.deserialize(...)`) 

also I do something similar in pandas encrypted so I initially thought I could merge both, but not really actually

finally I did not want to make the `fhe_client_server.py` file heavier, but I could create a `utils.py` in `deployment` for that actually, let me know what you think !

src/concrete/ml/deployment/fhe_client_server.py
RomanBredehoft: here q_result_enc can be a tuple , and when you do something `not  all(x is None for x in q_data_enc)`, mypy still complains, but maybe there's a better way ?

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: mypy still complains after that unfortunately

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: yep that's a mistake, good catch !

src/concrete/ml/torch/hybrid_model.py
RomanBredehoft: so, for the above changes, mypy started to complain about this typing (idk why it started on my side but whatever). The issue was that : 
- `inferences` is typed `List[numpy.ndarray]`
- but `deserialize_decrypt_dequantize` always returns a `numpy.ndarray`
- so mypy was complaining that `inferences.append(decrypted_prediction)` has some type mismatches

the `numpy.array` removal from `numpy.array(inferences)` is a mistake indeed, although it should not break anything, I'll put it back for clarity

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -179,7 +181,7 @@ def __init__(
         # Concrete ML attributes for FHE training
         # These values are hardcoded for now
         # We don't expose them in the __init__ arguments but they are taken
-        # into account when training, so we can just modify them manually.
+        # into account when training, so wecan just modify them manually.
```
RomanBredehoft, line 184: I'll fix that as well

src/concrete/ml/deployment/fhe_client_server.py
RomanBredehoft: yes I did !

RomanBredehoft:
> is the composition set by default in sgdclassifier? 

yes, there is no interactive training anymore, only non-interactive with composition

> If so are we sure the new PBS dequant requant isn't impacting the convergence?
 
good question, i don't know. The requant part makes sure that values are the same with respect to what we were doing in interactive training before, so this part is ok. But I guess you are more worried of the PBS part and the fact that we are using rounding, right ? In that case yes not sure how we could easily assess that it does not impact the convergence. All I can say for now is that tests are fine and the notebook looks good 😅  @jfrery

src/concrete/ml/deployment/fhe_client_server.py
RomanBredehoft: > Can't we have an assert maybe? Either here or in built in post processing?

I thought of it but I didn't find any proper way of doing it. The simplest way would be to check if `self.model` is a QuantizedModule (ok) or not (not ok), but I'm afraid this is a bit light and not very robust. Still would you prefer that we do that ? 

also there won't be a way to add any check in post-processing since the error is raised before entering in the method (ie: give several inputs instead of one)

RomanBredehoft: > I would still write explicitely in the variable name the difference between q_result f_result and result. I don't find these explicit enough.

yes I get this, but then I would be more in favor of a compromise then (like `float_result` or something) 😅 . I really found `deserialized_decrypted_dequantized_result` and associated hard to read and follow, especially when there are many successive lines. Like I couldn't differentiate `deserialized_decrypted_dequantized_result` from `deserialized_decrypted_quantized_result ` with a single look honestly 😅

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: no actually after looking at this, this loop is for quantizing and encrypting the values, the training iteration happens in the following loop

src/concrete/ml/common/utils.py
RomanBredehoft: I moved them to a `_utils.py` file the deployment directory

src/concrete/ml/deployment/fhe_client_server.py
RomanBredehoft: I added an assert with QM + I renamed the variables (`q` -> `quant`, `enc` -> `encrypted`)

.gitignore
RomanBredehoft: I've created a `.gitignore` in `advanced_examples`

src/concrete/ml/quantization/quantized_module.py
jfrery: that is weird. Then yeah not choice but to ignore

src/concrete/ml/deployment/fhe_client_server.py
jfrery: >  I really found deserialized_decrypted_dequantized_result and associated ones hard to read and follow

Agreed. Can you also make the before and after post processing variable different in

```python
result = to_tuple(self.model.dequantize_output(*result_quant))
...
result = self.model.post_processing(*result)
```

and then I am good !

src/concrete/ml/deployment/fhe_client_server.py
jfrery: I see. I don't have any solution. mypy is boring..

jfrery:
> > If so are we sure the new PBS dequant requant isn't impacting the convergence?
> 
> good question, i don't know. The requant part makes sure that values are the same with respect to what we were doing in interactive training before, so this part is ok. But I guess you are more worried of the PBS part and the fact that we are using rounding, right ? In that case yes not sure how we could easily assess that it does not impact the convergence. All I can say for now is that tests are fine and the notebook looks good 😅 @jfrery

I did the analysis -> https://github.com/zama-ai/concrete-ml/pull/660#issuecomment-2150464304. 

All in all, looks like the convergence isn't impacted. Good to go!

andrei-stoian-zama (approved):
Very well done! thanks

jfrery (approved):
All good thanks!

PR #663, opened 2024-05-01
docs: update documentation on client / server API

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4430

### DIFF
--- a/docs/advanced_examples/ClientServer.ipynb
+++ b/docs/advanced_examples/ClientServer.ipynb
@@ -232,10 +232,7 @@
     "fhemodel_client = FHEModelClient(network.client_dir.name, key_dir=network.client_dir.name)\n",
     "\n",
     "# The client first need to create the private and evaluation keys.\n",
-    "fhemodel_client.generate_private_and_evaluation_keys()\n",
-    "\n",
-    "# Get the serialized evaluation keys\n",
-    "serialized_evaluation_keys = fhemodel_client.get_serialized_evaluation_keys()"
+    "serialized_evaluation_keys = fhemodel_client.generate_private_and_evaluation_keys()"
    ]
   },
   {
--- a/docs/advanced_examples/ClientServer.ipynb
+++ b/docs/advanced_examples/ClientServer.ipynb
@@ -232,7 +232,7 @@
     "fhemodel_client = FHEModelClient(network.client_dir.name, key_dir=network.client_dir.name)\n",
     "\n",
     "# The client first need to create the private and evaluation keys.\n",
-    "serialized_evaluation_keys = fhemodel_client.generate_private_and_evaluation_keys()"
+    "serialized_evaluation_keys = fhemodel_client.get_serialized_evaluation_keys()"
    ]
   },
   {
--- a/docs/guides/client_server.md
+++ b/docs/guides/client_server.md
@@ -27,35 +27,51 @@ The `FHEModelDev`, `FHEModelClient`, and `FHEModelServer` classes in the `concre
 
 ### Example Usage
 
-<!--pytest-codeblocks:skip-->
-
 ```python
-# Development machine
-dev = FHEModelDev(path_dir='path/to/save', model=my_trained_model)
+from concrete.ml.sklearn import DecisionTreeClassifier
+from concrete.ml.deployment import FHEModelDev, FHEModelClient, FHEModelServer
+import numpy as np
+
+# Define the directory for FHE client/server files
+fhe_directory = '/tmp/fhe_client_server_files/'
+
+# Initialize the Decision Tree model
+model = DecisionTreeClassifier()
+
+# Generate some random data for training
+X = np.random.rand(100, 20)
+y = np.random.randint(0, 2, size=100)
+
+# Train and compile the model
+model.fit(X, y)
+model.compile(X)
+
+# Setup the development environment
+dev = FHEModelDev(path_dir=fhe_directory, model=model)
 dev.save()
 
-# Client machine
-client = FHEModelClient(path_dir='path/to/client/zip', key_dir='path/to/keys')
-client.generate_private_and_evaluation_keys()
+# Setup the client
+client = FHEModelClient(path_dir=fhe_directory, key_dir="/tmp/keys_client")
 serialized_evaluation_keys = client.get_serialized_evaluation_keys()
-# The client encrypts the data to be sent to the server for processing.
-encrypted_data = client.quantize_encrypt_serialize(my_data)
 
-# Server machine
-server = FHEModelServer(path_dir='path/to/server/zip')
+# Client pre-processes new data
+X_new = np.random.rand(1, 20)
+encrypted_data = client.quantize_encrypt_serialize(X_new)
+
+# Setup the server
+server = FHEModelServer(path_dir=fhe_directory)
 server.load()
 
-# Run the FHE execution on the encrypted data.
+# Server processes the encrypted data
 encrypted_result = server.run(encrypted_data, serialized_evaluation_keys)
 
-# Back to Client machine
-# The client decrypts the result.
+# Client decrypts the result
 result = client.deserialize_decrypt_dequantize(encrypted_result)
 ```
 
 > **Data Transfer Overview:**
 >
-> - **From Client to Server:** `serialized_evaluation_keys`, `encrypted_data`.
+> - **From Client to Server:** `serialized_evaluation_keys` (once), `encrypted_data`.
 > - **From Server to Client:** `encrypted_result`.
 
 These objects are serialized into bytes to streamline the data transfer between the client and server.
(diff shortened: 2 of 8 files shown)


docs/guides/client_server.md
```diff
@@ -13,7 +13,38 @@ The diagram above shows the steps that a developer goes through to prepare a mod
...
+> **Important Note:** In a client-server production using FHE, the server's output format depends on the model type. For regressors, the output matches the `predict()` method from scikit-learn, providing direct predictions. For classifiers, the output uses the `predict_proba()` method format, offering probability scores for each class, which allows clients to determine class membership by applying a threshold (commonly 0.5).
+
+### Using the API Classes
+
+The `FHEModelDev`, `FHEModelClient`, and `FHEModelServer` classes in the `concrete.ml.deployment` module make it easy to deploy and interact between the client and server:
+
+- **`FHEModelDev`**: This class is used during the development phase to prepare and save the model artifacts (`client.zip` and `server.zip`). It handles the serialization of model parameters and adds versioning information to ensure compatibility between the client and server components.
```
RomanBredehoft, line 22: `It handles the serialization of model parameters` I think this could be a bit confusing as one could think we are actually serializing the weight and bias values / hyper-params. I think it would be more accuracy to say `It handles the serialization of the underlying FHE circuit as well as the crypto-parameters used for generating the keys` (you could replace `circuit` by `model if you really prefer)

docs/guides/client_server.md
```diff
@@ -13,7 +13,38 @@ The diagram above shows the steps that a developer goes through to prepare a mod
...
+
+- **`FHEModelClient`**: This class is used on the client side to manage cryptographic keys, encrypt data before sending it to the server, and decrypt results received from the server. It also handles the loading of quantization parameters and pre/post-processing from `serialized_processing.json`.
```
RomanBredehoft, line 24: I would say : 
- `manage cryptographic keys` -> `generate and serialize the cryptographic keys`
- `encrypt data` -> `encrypt the data` or `encrypt the user's data` 
- `decrypt results` -> `decrypt the results`

docs/guides/client_server.md
```diff
@@ -13,7 +13,38 @@ The diagram above shows the steps that a developer goes through to prepare a mod
...
+
+- **`FHEModelServer`**: This class is used on the server side to load the FHE circuit from `server.zip` and execute the model on encrypted data received from the client.
+
+### Example Usage
+
+<!--pytest-codeblocks:skip-->
+
+```python
+# Development machine
+dev = FHEModelDev(path_dir='path/to/save', model=my_trained_model)
```
RomanBredehoft, line 34: I think you could add 3/4 lines where you init and train a model to make this example more complete !

docs/guides/client_server.md
```diff
@@ -13,7 +13,38 @@ The diagram above shows the steps that a developer goes through to prepare a mod
...
+dev.save()
+
+# Client machine
+client = FHEModelClient(path_dir='path/to/client/zip', key_dir='path/to/keys')
+client.load()
+encrypted_data = client.quantize_encrypt_serialize(my_data)
+result = client.deserialize_decrypt_dequantize(encrypted_result)
```
RomanBredehoft, line 41: putting this line before the server part makes things confusing imo, I would add another `Client machine`  after the `Server machine`

docs/guides/client_server.md
```diff
@@ -13,7 +13,38 @@ The diagram above shows the steps that a developer goes through to prepare a mod
...
+
+```python
+# Development machine
+dev = FHEModelDev(path_dir='path/to/save', model=my_trained_model)
+dev.save()
+
+# Client machine
+client = FHEModelClient(path_dir='path/to/client/zip', key_dir='path/to/keys')
+client.load()
+encrypted_data = client.quantize_encrypt_serialize(my_data)
```
RomanBredehoft, line 40: could you also add how we get the `evaluation_keys` , needed for the server ?

docs/guides/client_server.md
```diff
@@ -13,7 +13,38 @@ The diagram above shows the steps that a developer goes through to prepare a mod
...
+result = client.deserialize_decrypt_dequantize(encrypted_result)
+
+# Server machine
+server = FHEModelServer(path_dir='path/to/server/zip')
+server.load()
+encrypted_result = server.run(encrypted_data, evaluation_keys)
```
RomanBredehoft, line 46: if we want to be exhaustive I would also add the se/deserialize calls before and after the run

RomanBredehoft (changes requested):
great update, I suggested a few improvements !

docs/guides/client_server.md
jfrery: desarialization and serialization are done within the run.

docs/guides/client_server.md
jfrery: I will put the server in the middle. I don't want this to become like the notebook.

docs/guides/client_server.md
jfrery: I am afraid we make this too complicated if we run the whole thing. Let me see.

RomanBredehoft: well, you skip the codeblock so no worries on running it no ?

docs/guides/client_server.md
RomanBredehoft: ah yes indeed my bad !

docs/guides/client_server.md
RomanBredehoft: but ok for me if you leave it out to make this example simpler !

jfrery: Ah I thought your idea was to actually run the code. I will check is it doesn't make it too hard it makes sense.

docs/guides/client_server.md
jfrery: Alright I follow you but I have no idea what's the proper way between data / the data here.

docs/guides/client_server.md
jfrery: Good point.

docs/guides/client_server.md
```diff
@@ -13,7 +13,52 @@ The diagram above shows the steps that a developer goes through to prepare a mod
...
+
+```python
+# Development machine
+dev = FHEModelDev(path_dir='path/to/save', model=my_trained_model)
+dev.save()
+
+# Client machine
+client = FHEModelClient(path_dir='path/to/client/zip', key_dir='path/to/keys')
+client.generate_private_and_evaluation_keys()
+serialized_evaluation_keys = client.get_serialized_evaluation_keys()
```
RomanBredehoft, line 40: small note: could you add a few newlines to make this part more readable ?

also, makes me think that it could maybe be a good idea to insert `self.generate_private_and_evaluation_keys()` in `client.get_serialized_evaluation_keys()` if keys are not generated, api-wise I mean !

jfrery: Good idea. Done

docs/guides/client_server.md
jfrery: Should be good.

docs/guides/client_server.md
jfrery: Done

docs/guides/client_server.md
jfrery: I have the example running should be good.

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -190,9 +190,8 @@ def init_fhe_client(
                 path_dir=str(path_to_client.resolve()), key_dir=str(self.path_to_keys.resolve())
             )
             # The client first need to create the private and evaluation keys.
-            client.generate_private_and_evaluation_keys()
-            # Get the serialized evaluation keys
-            serialized_evaluation_keys = client.get_serialized_evaluation_keys()
+            serialized_evaluation_keys = client.generate_private_and_evaluation_keys()
```
RomanBredehoft, line 193: I was thinking more the other way 😅  like keep ` serialized_evaluation_keys = client.get_serialized_evaluation_keys()` but include `self.generate_private_and_evaluation_keys()` in `get_serialized_evaluation_keys` 

or then have a new method called "generate_keys_and_get_serialized_evaluation_keys` or similar

but that was more of an idea, we could make it an issue and consider it later if you prefer

jfrery: Yeah actually it makes more sense the way you describe. Will update.

docs/advanced_examples/ClientServer.ipynb
```diff
@@ -232,10 +232,7 @@
     "fhemodel_client = FHEModelClient(network.client_dir.name, key_dir=network.client_dir.name)\n",
     "\n",
     "# The client first need to create the private and evaluation keys.\n",
-    "fhemodel_client.generate_private_and_evaluation_keys()\n",
-    "\n",
-    "# Get the serialized evaluation keys\n",
-    "serialized_evaluation_keys = fhemodel_client.get_serialized_evaluation_keys()"
+    "serialized_evaluation_keys = fhemodel_client.generate_private_and_evaluation_keys()"
```
RomanBredehoft, line 235: this line should still use `get_serialized_evaluation_keys`

RomanBredehoft (changes requested):
there's a last change to fix

docs/advanced_examples/ClientServer.ipynb
jfrery: Good catch. updated.

RomanBredehoft (approved):
thanks a lot !

PR #664, opened 2024-05-01
chore: remove versions.json at root of deployment folder

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4434


RomanBredehoft (approved):
good catch !

PR #665, opened 2024-05-01
feat: add FHE training deployment

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4373
closes https://github.com/zama-ai/concrete-ml-internal/issues/4454

Proposition:

```python
from concrete.ml.deployment import FHEModelDev

# Save the training FHE circuit for production 
fhe_dev = FHEModelDev("fhe_training_sgd", sgd_clf_binary_fhe)
fhe_dev.save(mode="training|inference")

# No change on the client and server side
```

Check the notebook to see how to use it. Basically, it's the same as before for the user. We only add a single parameter in the fhe_dev.save


src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -25,6 +26,27 @@
...
 
 
+class Mode(Enum):
+    """Mode for the FHE API."""
+
+    INFERENCE = "inference"
+    TRAINING = "training"
+
+
+def ensure_tuple(x: Union[Any, Tuple[Any, ...]]) -> Tuple[Any, ...]:
```
RomanBredehoft, line 36: we already have a `to_tuple` function in `concrete.ml.common.utils` that does exactly this !

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -25,6 +26,27 @@
     from importlib_metadata import version
 
 
+class Mode(Enum):
```
RomanBredehoft, line 29: imo `Mode` is not explicit enough, even if it's defined here. Since we already have `FheMode`, maybe `ServerMode`, `RunMode`or `EncryptedMode`  ?

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -300,72 +354,109 @@ def get_serialized_evaluation_keys(self) -> bytes:
...
+            x (Union[numpy.ndarray, Tuple[numpy.ndarray, ...]]): the values to quantize,
+                encrypt and serialize
 
         Returns:
-            bytes: the quantized, encrypted and serialized values
+            Union[bytes, Tuple[bytes, ...]]: the quantized, encrypted and serialized values
         """
+
+        # Ensure x is a tuple
+        if not isinstance(x, tuple):
```
RomanBredehoft, line 371: you can use `to_tuple` here then

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -300,72 +354,109 @@ def get_serialized_evaluation_keys(self) -> bytes:
...
+            x = (x,)
+
         # Quantize the values
-        quantized_x = self.model.quantize_input(x)
+        quantized_x = self.model.quantize_input(*x)
+
+        # Ensure quantized_x is a tuple
+        if not isinstance(quantized_x, tuple):
```
RomanBredehoft, line 378: same

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -300,72 +354,109 @@ def get_serialized_evaluation_keys(self) -> bytes:
...
+            quantized_x = (quantized_x,)
 
         # Encrypt the values
-        enc_qx = self.client.encrypt(quantized_x)
+        enc_qx = self.client.encrypt(*quantized_x)
+
+        # Ensure enc_qx is a tuple
+        if not isinstance(enc_qx, tuple):
```
RomanBredehoft, line 385: same

src/concrete/ml/pytest/utils.py
```diff
@@ -184,6 +186,14 @@ def get_sklearn_linear_models_and_datasets(
             partial(TweedieRegressor, link="identity", power=0.0),
         ]
 
+    # If the linear model is SGDClassifier,
+    # we need to handle the training parameters
+    if is_model_class_in_a_list(SGDClassifier, linear_classes):
```
RomanBredehoft, line 191: yes great

tests/deployment/test_client_server.py
```diff
@@ -89,18 +96,31 @@ def test_client_server_sklearn(
...
         pytest.skip("Skipping KNN, because FHE predictions and clear ones are differents.")
 
+    if (
+        get_model_name(model_class) == "SGDClassifier"
+        and parameters.get("n_classes") != 2
+        and getattr(model_class, "keywords", {}).get("fit_encrypted") is True
+    ):
+        # Skipping SGDClassifier for this test
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4182
+        pytest.skip("Skipping SGDClassifier, because it only supports 2 classes.")
```
RomanBredehoft, line 106: why do you need to add this here ?

tests/deployment/test_client_server.py
```diff
@@ -89,18 +96,31 @@ def test_client_server_sklearn(
...
     x, y = load_data(model_class, **parameters)
 
     x_train = x[:-1]
     y_train = y[:-1]
     x_test = x[-1:]
 
-    model = instantiate_model_generic(model_class, n_bits=n_bits)
+    # Instantiate the model
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
```
RomanBredehoft, line 117: what is this userwarning ?

andrei-stoian-zama (changes requested):
The notebook seems a bit loaded.. 

I think we could:
- move the plot_decision boundary functions to some file in `utils`. 
- why do we need two `plot_decision_boundary` functions?
- only show the decision boundary graph at iteration 0 and the last iteration

Next, we should showcase Deployment first:
1. "Training on encrypted data in production": export the trainer with the FheModelDev/etc, load client/server, etc.. 
2. "Develop a Logistic regression trainer for encrypted data" -> here we show the simulation part

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -149,17 +175,22 @@ def __init__(self, path_dir: str, model: Any = None):
...
             Path: the path to the json file
         """
+        module_to_export = self.model.training_quantized_module if is_training else self.model
         serialized_processing = {
-            "model_type": self.model.__class__,
-            "model_post_processing_params": self.model.post_processing_params,
-            "input_quantizers": self.model.input_quantizers,
-            "output_quantizers": self.model.output_quantizers,
+            "model_type": module_to_export.__class__,
+            "model_post_processing_params": module_to_export.post_processing_params,  # TODO
```
RomanBredehoft, line 190: you still have a TODO here

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -149,17 +175,22 @@ def __init__(self, path_dir: str, model: Any = None):
 
         Path(self.path_dir).mkdir(parents=True, exist_ok=True)
 
-    def _export_model_to_json(self) -> Path:
+    def _export_model_to_json(self, is_training: bool = False) -> Path:
         """Export the quantizers to a json file.
 
+        Args:
+            is_training (bool): If True, we export the training mode.
```
RomanBredehoft, line 182: `If True, we export the training mode.` -> `module` or `circuit`, but not `mode` right ?

RomanBredehoft (changes requested):
thanks ! I have several comments and observations

src/concrete/ml/deployment/fhe_client_server.py
fd0r: `ModelMode` also could work

jfrery: DeploymentMode ?

src/concrete/ml/deployment/fhe_client_server.py
jfrery: changed to circuit

src/concrete/ml/deployment/fhe_client_server.py
jfrery: yep removed thanks!

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -173,35 +204,56 @@ def _export_model_to_json(self) -> Path:
...
-                For more details on the topic please refer to Concrete-Python's documentation.
 
         Raises:
-            Exception: path_dir is not empty
+            Exception: path_dir is not empty or training module does not exist
         """
+
+        if isinstance(mode, str):
+            mode_upper = mode.upper()
+            if mode_upper not in Mode.__members__:
```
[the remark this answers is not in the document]
jfrery: Agreed. Updated.

src/concrete/ml/deployment/fhe_client_server.py
jfrery: done

src/concrete/ml/deployment/fhe_client_server.py
jfrery: done

src/concrete/ml/deployment/fhe_client_server.py
jfrery: done

tests/deployment/test_client_server.py
jfrery: Because we test sklearn models with more than 2 classes but SGDClassifier isn't compatible yet. Or am I wrong?

tests/deployment/test_client_server.py
jfrery: It's this that appear when we use `fit_encrypted = True`

https://github.com/zama-ai/concrete-ml/blob/da83329717df905da6e69e946619339cbd51a5cc/src/concrete/ml/sklearn/linear_model.py#L239-L243

tests/deployment/test_client_server.py
```diff
@@ -337,3 +357,66 @@ def check_input_compression(model, fhe_circuit_compressed, is_torch, **compilati
...
+        ("inference", False, None),
+        ("inference", True, None),
+        ("training", False, ERROR_MSG_NO_FHE_CIRCUIT),
+        ("training", True, None),
+        (Mode.INFERENCE, False, None),
+        (Mode.TRAINING, False, ERROR_MSG_NO_FHE_CIRCUIT),
+        (Mode.TRAINING, True, None),
+    ],
+)
+def test_save_mode(n_bits, fit_encrypted, mode, error_message):
```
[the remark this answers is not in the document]
jfrery: This test checks the different mode  / compatibility of the FHEModelDev save method. I will rename the test which, I agree, is too generic.

`test_save_mode_handling` maybe

src/concrete/ml/deployment/fhe_client_server.py
andrei-stoian-zama: let's go with `FunctionToDeploy` = Training/Inference

tests/deployment/test_client_server.py
RomanBredehoft: yes but why now ? I mean why wasn't it already here, did you change somthing with SGDClassifier itself ?

jfrery: We were not testing it with `fit_encrypted = True` this skip is only when n_classes != 2 and `fit_encrypted = True`

RomanBredehoft: gotcha thanks !

tests/deployment/test_client_server.py
RomanBredehoft: got it thanks !

tests/deployment/test_client_server.py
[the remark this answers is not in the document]
RomanBredehoft: thanks !

RomanBredehoft (dismissed):
thanks a lot for this !

RomanBredehoft (dismissed):

fd0r:
Determinism failed

fd0r (dismissed):

jfrery:
Interesting. I don't see any failure

RomanBredehoft:
@jfrery @fd0r it looks like the diff from the determinism file did not pass : 
```
diff "${OUTPUT_DIRECTORY}/one.txt" "${OUTPUT_DIRECTORY}/two.txt" -I "passed in"
```
because looking at the CI's log, it seems the difference happens with an MLIR : 
```
func.func @main(%arg0: tensor<1x10x!FHE.esint<17>>) -> tensor<1x1x!FHE.esint<17>> {
```
vs 
```
func.func @main(%arg0: tensor<1x10x!FHE.esint<16>>) -> tensor<1x1x!FHE.esint<16>> {
```

the only place where we print the mlir is with the python script's last line : 

```
    try:
        model.compile(x, configuration=default_configuration, show_mlir=True)
    except RuntimeError as err:
        print(err)
    except AssertionError as err:
        print(err)
```
if the compilation fails, there is not error, we only put the traceback

so my guess is : the compilation failed in both runs, but for some weird reasons the bitwidths were not the same 

but if it's not that then i don't know it's weird indeed

jfrery:
Thanks for looking at it @RomanBredehoft. I will improve the determinism logs as I have no idea how to fix this diff.

RomanBredehoft:
you could try to frist try to run `test_seeding.py` (with seed `8533` as written in the log) and see if all models properly compile

else yes idk much more

jfrery:
> you could try to frist try to run `test_seeding.py` (with seed `8533` as written in the log) and see if all models properly compile
> 
> else yes idk much more

Tests are fine. It's the diff comparison that is different and should not.

RomanBredehoft:
> Tests are fine. It's the diff comparison that is different and should not.

I know but as I explained above, the diff compares rwo files that are basically storing the traceback of `test_seeding.py` in case the circuits' compilation fail, that's how part of the determinism step work, so I would still try what I suggested above !

andrei-stoian-zama (approved):

docs/advanced_examples/LogisticRegressionTraining.ipynb
```diff
@@ -462,6 +395,339 @@
...
+    "            # Back on the client, we deserialize the result\n",
+    "            weights, bias = fhe_client.deserialize_decrypt_dequantize(results)\n",
+    "\n",
+    "            # Compute and store accuracy\n",
+    "            acc = compute_model_accuracy(weights, bias, X_binary, y_binary)\n",
+    "            acc_history.append(acc)\n",
+    "\n",
+    "        print(f\"Epoch {epoch + 1}/{n_epochs} completed. Accuracy: {acc_history[-1]}\")\n",
+    "\n",
+    "    return weights, bias, acc_history\n",
```
RomanBredehoft, line 626: haven't noticed that before, but I find it pretty weird that we are only handling weights and bias as results. The FheModelClient has the model's instance, wouldn't it be possible to return a trained SGDClassifier instead of the dequantized values ? 

because looking at the notebook I see that since we only have the weights, we need to re-implement the inference manually. Log reg is simple enough so that works but honestly not sure this is a good long-term solution

docs/advanced_examples/LogisticRegressionTraining.ipynb
```diff
@@ -462,6 +395,339 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Define utils function to evaluate the model, print it's accuracy\n",
+    "\n",
+    "\n",
+    "def model_inference(weights, bias, X):\n",
```
RomanBredehoft, line 506: see comment https://github.com/zama-ai/concrete-ml/pull/665/files#r1618751835, I don't think we should need to re-implement the inference like this

docs/advanced_examples/LogisticRegressionTraining.ipynb
```diff
@@ -462,6 +395,339 @@
...
...bzZuXEpm5glq1mxEeHg9p6N5pczMk2RmniAgIATT1ElHRemaa4bRqdMQ1q9fwpEj+6lRoxF167bAMAyno4mXUZElIiIiBeLnF0CLFl2djuG1EhMTeP/9x/nxx/nYtkW5clXp1u02evceR0BA/uvxSMHt3r2BDz54nB9++BjLyqJs2fJ06XIrN944nuDgSk7HEy+ij0dEREREHLZ69QLGjWvHmjXbse1ngQ85duxffPTRczzwQCzHj6c4HbHE27RpOWPHtuaHH1ZjWY8DH3LixK189tnrjB3bliNHDjgdUbyIiiwRERERB2VkpDN9+s3Ydncsay1wN3AjMAPL+oE9e7Yzd+7DDqcs2dxuN08/PYjs7FZY1s/kLEp8IzANy1rD77+n8Oab9zucUryJiiwRERERBy1f/gHp6Uex7Wf456LPTbCskXz11ZtkZp50Ip5X+OmnRRw69Cu2PR0IOKM1Esv6N8uWvU9q6iEn4okXUpElIjkSEiAxkfmJ9YmLjyApyelAIiKlw+7dG3C56gGR+fTowokTRzl8WC/M5+vXXzdgmuWBFvn0uAa3O5P9+7d7MpZ4MU18IVLa/VlcJTUngZHEH2pCeOOKxGohYhERj/DzCwCOAG4gr0WdfwfA17esB1N5Fz+/AGw7HUjnnyNZAH8Aeoyl6GgkS6Q0i4tjfnwQE5JGEscg4ulIbJ+KDBqkAktExFNat74Bt/t34NM8Wm0M41Xq1ImmQoVqHk7mPVq16oFtZwLv5NNjFlWq1KFmzUaejCVeTCNZIqVVQgJERZFALxKJAmDcIIcziYiUQnXrtqBp0y5s2DAcy/IHupHzOfgRYAK2/Q39+n3qaMaSLiysNldeOYDly+/DskLJmfTCBRwDngQ+oG/f17QumRQZFVkiQlSU0wlEREq3Bx6Yy5Qpvfnll+twuWph29Ww7XUYRjYjRrxI69Y9nY5Y4t111yxOnEhj9er+uFwR2HYNYCO2fZwBAx7lmmuGOR1RvIiKLBERERGHBQWF8vjjS9i0aTnff/8hJ04cIyLiejp1Gkr58mFOx/MKfn4BTJjwKdu2rWbp0vc4duwwYWFd6dRpKJUr13A6nngZFVkiIiIO+OKLV5g3bxpHj/6Br68vMTHXc8stTxMUFOp0NHFIZuYJkpK2sW/fdtLTj2EYLn7//TcVWUWsXr1W1KvXyukY4uVUZImI
```
RomanBredehoft, line 679: I'm not sure I get the point of this 90 degrees part

script/make_utils/check_pytest_determinism.sh
```diff
@@ -33,12 +33,35 @@ then
...
-
 # Exceptions:
-#   passed in: since it is related to timings
-diff "${OUTPUT_DIRECTORY}/one.txt" "${OUTPUT_DIRECTORY}/two.txt" -I "passed in"
-echo "Successful determinism check"
+#   in X.Xs: since it is related to timings
+diff_output=$(diff "${OUTPUT_DIRECTORY}/one.txt" "${OUTPUT_DIRECTORY}/two.txt" -I "in [0-9]*\.[0-9]*s")
+
+# If a diff is present, we need to print the tests that failed to be reproduced for debugging.
+if [ -n "$diff_output" ]; then
```
RomanBredehoft, line 41: not sure I get everything here but I trust you !

src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -105,30 +126,36 @@ def load(self):
...
 
         Returns:
-            bytes: the result of the model
+            Union[bytes, Tuple[bytes, ...]]: the result of the model
         """
         assert_true(self.server is not None, "Model has not been loaded.")
 
-        deserialized_encrypted_quantized_data = fhe.Value.deserialize(
-            serialized_encrypted_quantized_data
+        if isinstance(serialized_encrypted_quantized_data, bytes):
```
RomanBredehoft, line 144: you can use `to_tuple` here as well I think

src/concrete/ml/pytest/utils.py
```diff
@@ -127,7 +129,13 @@ def _get_sklearn_models_and_datasets(model_classes: List, unique_models: bool =
             # the test execution timings
             n_classes_to_test = (
                 [2]
-                if unique_models or get_model_class(model_class) == KNeighborsClassifier
+                if unique_models
+                or get_model_class(model_class) == KNeighborsClassifier
+                or (
+                    isinstance(model_class, partial)
+                    and model_class.func == SGDClassifier
+                    and model_class.keywords.get("fit_encrypted", False)
```
RomanBredehoft, line 137: I don't get why it's working : shouldn't it be the opposite ? iirc, FHE training does not handle other than binary classification, but here we are puting 2 classes if *not* using FHE training (`and model_class.keywords.get("fit_encrypted", False)`), while 2 and 4 with FHE training 🤔 It should fail with 4 classes no ?

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -252,7 +252,9 @@ def post_processing_params(self, post_processing_params: Dict[str, Any]):
         self._post_processing_params = post_processing_params
 
     # pylint: disable-next=no-self-use
-    def post_processing(self, values: numpy.ndarray) -> numpy.ndarray:
+    def post_processing(
```
RomanBredehoft, line 255: great, this is going to combine well with what I've done for quantized modules in https://github.com/zama-ai/concrete-ml/pull/660/files#diff-84c5dca0cde5235ebd72ca3a3d3ee9e9ca1c4d17e9e28ed7b92885705e5ae346

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -687,13 +680,7 @@ def fit(  # type: ignore[override]
 
         # If the model should be trained using FHE training
         if self.fit_encrypted:
-            if fhe is None:
-                fhe = "disable"
-                warnings.warn(
```
RomanBredehoft, line 692: no this warning was made on purpose, I think we should keep it no ? the idea if that users should set `fhe` to something if they activate `fit_encrypted` to be sure they know what they are doing

the reason why `fhe` is not defaulted to `disable` is to avoid the ambiguous situation of having `fit_encrypted` to false and `fhe` to `disable`, which wouldn't make much sense since training would be done in floating points with sklearn

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -763,13 +750,7 @@ def partial_fit(
         # A partial fit is similar to a fit with a single iteration. The slight differences between
         # both are handled in the encrypted method when setting `is_partial_fit` to True.
         if self.fit_encrypted:
-            if fhe is None:
```
RomanBredehoft, line 766: same as above

src/concrete/ml/pytest/utils.py
```diff
@@ -184,6 +192,15 @@ def get_sklearn_linear_models_and_datasets(
             partial(TweedieRegressor, link="identity", power=0.0),
         ]
 
+    # If the linear model is SGDClassifier,
+    # we need to handle the training parameters
+    if is_model_class_in_a_list(SGDClassifier, linear_classes):
+        linear_classes += [
+            partial(SGDClassifier, fit_encrypted=False),
+            # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4460
+            # partial(SGDClassifier, fit_encrypted=True, parameters_range=(-1, 1)),
```
RomanBredehoft, line 201: why not activating it here ? does it cause issues ?

tests/deployment/test_client_server.py
```diff
@@ -3,25 +3,41 @@
...
 from concrete.ml.quantization.quantized_module import QuantizedModule
+from concrete.ml.sklearn.linear_model import SGDClassifier
 from concrete.ml.torch.compile import compile_torch_model
 
 # pylint: disable=too-many-statements,too-many-locals
 
 
+# Add encrypted training with SGDClassifier manually
+# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4460
+MODELS_AND_DATASETS = MODELS_AND_DATASETS + [
```
RomanBredehoft, line 32: see comment https://github.com/zama-ai/concrete-ml/pull/665/files#r1618781294

tests/deployment/test_client_server.py
```diff
@@ -337,3 +352,61 @@ def check_input_compression(model, fhe_circuit_compressed, is_torch, **compilati
...
+
+    # Create FHEModelDev instance
+    with tempfile.TemporaryDirectory() as temp_dir:
+        model_dev = FHEModelDev(path_dir=temp_dir, model=model)
+
+        if error_message:
+            with pytest.raises((AssertionError, ValueError), match=error_message):
+                model_dev.save(mode=mode)
+        else:
+            try:
```
RomanBredehoft, line 409: better to remove this try/except so that we properly catch the error is there is one no ? worst case, could you at least print the traceback using something like `except ValueError as e:` (and include e in the `pytest.fail`)

RomanBredehoft (changes requested):
thanks for the feature !! lots of cleaning, which is great. I have several questions and observations

src/concrete/ml/pytest/utils.py
jfrery: > if not using FHE training (and model_class.keywords.get("fit_encrypted", False))

`.get(key, default)` returns the value for the key. If key doesn't exist it returns the default so False. So should be good, right? But yeah it's confusing.

We could write

```python
    and "fit_encrypted" in model_class.keywords
    and model_class.keywords["fit_encrypted"]
```

instead.


Note that this is useless because of this:

https://github.com/zama-ai/concrete-ml/blob/4f759790ab77d25f296bfa41626b3fa0a80021bc/src/concrete/ml/pytest/utils.py#L200-L201

script/make_utils/check_pytest_determinism.sh
jfrery: Yeah it wasn't as straightforward as I thought it would be. Basically, I print lines that are above the diffs for better debugging.

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: It's a fine-tuning example. Also showing that what we really have is a training circuit that can adapt to different dataset.

src/concrete/ml/pytest/utils.py
RomanBredehoft: ah yes my bad of course, read too fast. But yes you could write `model_class.keywords["fit_encrypted"]` instead since we are checking that we have a `SGDClassifier` model first anyway !

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: So you are asking for a feature to be able to go back to SGDClassifier from FHEModelClient. This makes sense to me! But this requires more work. E.g. we will need to implement a method that load the learned parameter back in the model and every model will have it's own function so not obvious. I would rather implement that in another PR.

src/concrete/ml/sklearn/linear_model.py
jfrery: Maybe I miss something but this didn't really make sense to me. If we have a warning here then why not for all our models in the predict?

> avoid the ambiguous situation of having fit_encrypted to false and fhe to disable, which wouldn't make much sense since training would be done in floating points with sklearn

If fit_encrypted = False then [we have this](https://github.com/zama-ai/concrete-ml/blob/4f759790ab77d25f296bfa41626b3fa0a80021bc/src/concrete/ml/sklearn/linear_model.py#L705-L710)

Or what do I miss?

src/concrete/ml/pytest/utils.py
jfrery: I would like to activate it here but we need more work to make our test our custom SGDClassifier. It causes lots of issues yes. E.g. correctness.

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: https://github.com/zama-ai/concrete-ml-internal/issues/4464

tests/deployment/test_client_server.py
jfrery: Ah yes this was for a previous version of the test useless now. Thanks!

jfrery: done

docs/advanced_examples/LogisticRegressionTraining.ipynb
RomanBredehoft: ah I see thanks ! worth emphasizing a bit on this in the text then !

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: ah yes you right about the second case, good thing then !

but the reason we don't have this in the `predict` methods is simply because there is not ambiguity there, we only have one parameter related to fhe execution : `fhe`. Whereas with encrypted training, we have this additional `fit_encrypted` in the init.

so more or less I believe the idea here was to better determine the `fhe` parameter's role, by having something like : 
- `None` : float clear
- `disable | simulate | fhe ` : the usual

so when a user sets `fit_encrypted`, it's better to make sure he's aware of what mode he's using

src/concrete/ml/pytest/utils.py
RomanBredehoft: I see, thanks !

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: Yes agreed.

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: Yes I opened https://github.com/zama-ai/concrete-ml-internal/issues/4464.

src/concrete/ml/sklearn/linear_model.py
jfrery: There is no notion of fhe= if it's a training in the clear and we have a assert on this. To me, fhe="disable" should be the default if users use fit_encrypted = True.

I don't think that the warning ` "Parameter 'fhe' isn't set while FHE training is enabled.\n Defaulting to '{fhe=}'"` is of any help to the user.

If the API is really confusing, having warning to try to explain isn't the right direction. Instead we should rethink the API.

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: I added a cell right before the fine-tuning

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: well iirc there was some discussion on this in particular when the API was made and this is how we decided to go for, maybe @fd0r can tell us more about it !

fd0r: Yes @RomanBredehoft got it right.

We don't want `disable` by default since it wouldn't make sense if `fit_encrypted=False`.
But not providing a value for `fhe` if `fit_encrypted=True` is also "dangerous". The idea here was to make sure that the user  has to provide some parameter for `fhe` in this case.

Now that I'm re-reading that I would have put an Exception instead of a warning here.

I don't think we should remove the warning.

docs/advanced_examples/LogisticRegressionTraining.ipynb
fd0r, line 1: Can the deployment be done without any call to the `fit` method?

We should be able to deploy with only the batch size, the number of features and the batch size as inputs.

fd0r: Also we should probably take composition into account such that n-iter is included in the query.

fd0r (changes requested):
Sorry for reviewing this so late.

I have a few comments on the API design.

IMHO we should consider composition for this, or merge as is and improve upon that, but it would be easier if it was already taken into account.

Also we shouldn't have to call the `fit` function before deploying imo. It's true that we would have to take the inputs ranges into account, the batch-size and number of features and targets.

src/concrete/ml/sklearn/linear_model.py
jfrery: > We don't want disable by default since it wouldn't make sense if fit_encrypted=False.

But if `fit_encrypted=False.` and fhe != None, we raise an error so this is not possible.


> But not providing a value for fhe if fit_encrypted=True is also "dangerous"

I really don't understand why this is "dangerous". For the inference, disable is the default. Why would it be dangerous to do it for training?


Also, what should the user learn from  `"Parameter 'fhe' isn't set while FHE training is enabled.\n Defaulting to '{fhe=}'"`?

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: > Can the deployment be done without any call to the fit method?
> 
> We should be able to deploy with only the batch size, the number of features and the batch size as inputs.

I just use the existing way of creating a training FHE circuit. Do we have something else than the `_fit_encrypted` method to create such circuit? Or are you proposing to wrap this into another method within the FHEModel API?

That being said, I definitely agree that having to instantiate fhe circuit for deployment with a call to fit is super weird...



> Also we should probably take composition into account such that n-iter is included in the query.

I am not sure where the n-iter would belong but I believe it's a parameter that needs to be sent to the server so I would say it would need to be sent at the same time as the serialized encrypted values. I don't think it has anything to do with the FHEModel API unless we start implementing the communication protocol.

src/concrete/ml/sklearn/linear_model.py
fd0r: > I really don't understand why this is "dangerous". For the inference, disable is the default. Why would it be dangerous to do it for training?

Because we would be changing the default from the signature mostly

fd0r: > Also, what should the user learn from "Parameter 'fhe' isn't set while FHE training is enabled.\n Defaulting to '{fhe=}'"?

Mainly that the default value is changing is shouldn't really  happen. And that the prefered way is to specify this argument.

jfrery: Alright 🏳️ well I will add back the warnings and tests for it.

docs/advanced_examples/LogisticRegressionTraining.ipynb
jfrery: I created this for the first point https://github.com/zama-ai/concrete-ml-internal/issues/4466

fd0r: Thanks a lot for these changes, let's address the rest in a follow-up PR.

fd0r (approved):
Let's iterate on that! Thanks a lot for taking care of this!

src/concrete/ml/sklearn/linear_model.py
```diff
@@ -687,13 +680,7 @@ def fit(  # type: ignore[override]
...
         # If the model should be trained using FHE training
         if self.fit_encrypted:
-            if fhe is None:
-                fhe = "disable"
-                warnings.warn(
-                    "Parameter 'fhe' isn't set while FHE training is enabled.\n"
-                    f"Defaulting to '{fhe=}'",
-                    stacklevel=2,
-                )
+            fhe = "disable" if fhe is None else fhe
```
fd0r, line 683: This wasn't changed back

fd0r (changes requested):
The warning is still removed.

src/concrete/ml/sklearn/linear_model.py
jfrery: Should be good

fd0r (approved):

RomanBredehoft (approved):
looks fine, but how is it going to work once we integrate composition with training (https://github.com/zama-ai/concrete-ml/pull/660) ?

PR #668, opened 2024-05-05
Update UPGRADING.md

sukrucildirr:
Added Discord link.


andrei-stoian-zama:
Thank you but we don't think that link should be there. If you are unsure about what can be a good contribution, it's best to first make an issue about it and the team will discuss your proposal.

PR #673, opened 2024-05-06
chore(ci): update usage of slab-github-runner to last version

soonum:
With the version of Slab (and the runner associated), there is a breaking change on how a backend is handled.
This will be merged once the latest release of Slab is shipped to production.


RomanBredehoft (approved):
thanks a lot @soonum !

PR #676, opened 2024-05-07
chore: allow users to pass schema in encrypted data-frames

RomanBredehoft:
refs https://github.com/zama-ai/concrete-ml-internal/issues/4376


src/concrete/ml/pandas/_processing.py
```diff
@@ -86,9 +92,37 @@ def dequant(
...
+            f"data-frame's column names to their value mappings. Got {type(schema)=}"
+        )
+
+    column_names = list(pandas_dataframe.columns)
+
+    for column_name, column_mapping in schema.items():
+        if column_name not in column_names:
+            # TODO: Is this check actually relevant ? Can't the schema provide more columns than the
+            # one found in the data-frame ?
+            raise ValueError(
```
RomanBredehoft, line 111: should we allow schema with column names that do not match the ones found in the given data-frame ?

src/concrete/ml/pandas/_processing.py
```diff
@@ -189,10 +297,15 @@ def pre_process_dtypes(pandas_dataframe: pandas.DataFrame) -> Tuple[pandas.DataF
                 "supported."
             )
 
+    # TODO: Should all non-integers columns be considered by the schema if not None ? Currently,
```
RomanBredehoft, line 300: should we raise an error/warning if all non-integer columns from the data-frame were not covered by the given schema ?

src/concrete/ml/pandas/client_engine.py
```diff
@@ -37,7 +41,9 @@ def keygen(self, keys_path: Optional[Union[Path, str]] = None):
         else:
             self.client.keygen(True)
 
-    def encrypt_from_pandas(self, pandas_dataframe: pandas.DataFrame) -> EncryptedDataFrame:
+    def encrypt_from_pandas(
+        self, pandas_dataframe: pandas.DataFrame, schema: Optional[Dict] = None
```
RomanBredehoft, line 45: a schema is optional. If set, it should follow a specific format. 

if needed, we could also handle the output of `get_schema` (pandas data-frames) as an input here

tests/pandas/test_pandas.py
```diff
@@ -571,3 +574,211 @@ def test_print_and_repr():
...
+        if numpy.issubdtype(column.dtype, numpy.floating):
+            schema[column_name] = {
+                "min": column.min(),
+                "max": column.max(),
+            }
+
+        elif column.dtype == "object":
+            unique_values = column.unique()
+
+            # Only take strings into account and thus avoid NaN values
```
fd0r, line 593: You could also use the old `x != x` to detect NaNs

src/concrete/ml/pandas/_processing.py
fd0r: What would happen if they are missing from the given schema?

src/concrete/ml/pandas/_processing.py
fd0r: Imo raising the error as you are doing here is the correct behavior.

fd0r (approved):
lgtm, responded to the todos

src/concrete/ml/pandas/_processing.py
RomanBredehoft: they are automatically computed

RomanBredehoft: I'm just wondering because if could be an easy mistake to forget to put some columns, but no error will be raised

PR #684, opened 2024-05-17
chore: update cp version to 2.6.2dev20250529

fd0r:
For now we target Zama's index but once 2.6.2 is on pypi we'll need to change it (it will just be about removing the source argument from pyproject.toml).


pyproject.toml
```diff
@@ -90,7 +97,6 @@ jinja2 = "^3.1.2"
 LinkChecker = "^10.1.0"
 kaggle = "^1.5.12"
 tf2onnx = "^1.10.1"
-types-requests = "^2.28.0"
```
RomanBredehoft, line 93: why removing this ?

fd0r: Because we removed requests from our requirements so it seemed odd to have it here

RomanBredehoft: gotcha !

jfrery (dismissed):
Looks good to me. Thanks!

RomanBredehoft (changes requested):
do we want to do this upgrade knowing this issue : https://github.com/zama-ai/concrete-internal/issues/717 ?

fd0r:
> do we want to do this upgrade knowing this issue : [zama-ai/concrete-internal#717](https://github.com/zama-ai/concrete-internal/issues/717) ?

Very good question @RomanBredehoft since this isn't a release version it's fine with me with the verbose stuff but maybe @andrei-stoian-zama can have the last word on that?

fd0r:
Writing here what was decided on slack.

We wait to have a flag to disable these warnings and let the Concrete team investigate where these overflows come from and how badly they impact us.

andrei-stoian-zama:
The nightly with the disable flag should be available

fd0r:
Latest nightly still outputting warnings, let's wait for the next one!

Makefile
```diff
@@ -19,7 +19,7 @@ OPEN_PR="true"
 # Force the installation of a Concrete Python version, which is very useful with nightly versions
 # /!\ WARNING /!\: This version should NEVER be a wildcard as it might create some
 # issues when trying to run it in the future.
-CONCRETE_PYTHON_VERSION="concrete-python==2024.4.19"
+CONCRETE_PYTHON_VERSION="concrete-python==2.6.2.dev20240529"
```
RomanBredehoft, line 22: btw, now that nightlies are completely public and handled directly in the lock file, I think we can safely remove this part

pyproject.toml
```diff
@@ -34,7 +34,8 @@ readme = "README.md"
 # Investigate if it is better to fix specific versions or use lower and upper bounds  
 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/2665
 python = ">=3.8.1,<3.11"
-concrete-python = "2.6.0-rc1"
+concrete-python = {version="==2.6.2.dev20240529", source = "zama-pypi"}
+# concrete-python = ">=2.6.2.dev20240523,<3.0"
```
fd0r, line 38: Couldn't use that requirement because 2.6.2 > 2.6.2dev, so poetry lock would install 2.6.2

Makefile
fd0r: That's correct, let's do that in a follow up PR

pyproject.toml
```diff
@@ -78,6 +85,8 @@ mdformat = "^0.7.14"
 mdformat_myst = "^0.1.4"
 mdformat-toc = "^0.3.0"
 pip-audit = "^2.1.0"
+types-requests = "^2.32.0"
+requests="^2.32.1"
```
fd0r, line 89: `requests` was installed by `type-requests`

pyproject.toml
RomanBredehoft: but we want to fix CP's version no ?

Makefile
RomanBredehoft: sure yes : https://github.com/zama-ai/concrete-ml-internal/issues/4463

RomanBredehoft (approved):
a few questions that could be handled in a following PR, thanks !

pyproject.toml
fd0r: Well that's a good question. I'm not so sure about that.

In theory we should only require the major version of a package unless we know that there is a change that breaks our code base.

Introducing more flexibility here would let us avoid the need of doing a patch release whenever Concrete Python does ones. That wouldn't be reflected in the lock file though

pyproject.toml
fd0r: We need requests for dev scripts. @jfrery had removed it some time ago but by keeping `type-requests` we were in fact still installing it. It was weird to me that we had the type stubs of a packages that we weren't requiring. So I tried removing it but everything broke. So I added back requests here since I feel like it's more explicit than relying on the fact that `type-requests` requires `requests`.

pyproject.toml
RomanBredehoft: yeah for our other libraries I pretty much agree, but CP is a bit different, especially since we consider nightly versions and I don't think we would want poetry to consider the most recent nightly each time we update the lock (?)

pyproject.toml
RomanBredehoft: hmm interesting, gotcha

PR #689, opened 2024-05-21
feat: support `from_sklearn` for trees

fd0r:
Support `from_sklearn` for tree based models.

Two options:
- Quantization from thresholds: the main idea is to consider the thresholds of the nodes of the trees for quantization to not have to use data.
- Quantization from data: build a quantizer from the data provided by the user and quantize the thresholds based on that.

This also raises the question of non-uniform input quantization. We could quantize the data based on the thresholds thus reducing the number of bits required to log2(max_{feature}(node_{feature})).

That would leak the thresholds used in the model per feature but not the structure of the tree itself while increasing significantly the number of bits required.

We could try to automatically determine the n-bits to use to properly represent all thresholds but this might result in a very high bit-with.


src/concrete/ml/sklearn/base.py
```diff
@@ -1328,6 +1339,279 @@ def __init__(self, n_bits: Union[int, Dict[str, int]]):
 
         BaseEstimator.__init__(self)
 
+    # TODO: FIX EXACT PREDICTION WITH HIGH BIT WIDTH
+    # pylint: disable=too-many-locals,too-many-statements,too-many-branches
+    @classmethod
+    def from_sklearn_model(
```
jfrery, line 1345: We should put that somewhere else. It manipulates onnx which we shouldn't do here

src/concrete/ml/sklearn/base.py
```diff
@@ -65,7 +68,15 @@
...
-from .tree_to_numpy import tree_to_numpy
+from .tree_to_numpy import (
+    _compute_lsb_to_remove_for_trees,
+    get_equivalent_numpy_forward_from_onnx_tree,
+    get_onnx_model,
+    is_regressor_or_partial_regressor,
+    preprocess_tree_predictions,
+    tree_onnx_graph_preprocessing,
+    tree_to_numpy,
+)
```
jfrery, line 79: Can we keep using all these within tree_to_numpy?

src/concrete/ml/sklearn/base.py
```diff
@@ -1560,6 +1844,8 @@ def from_sklearn_model(
         Returns:
             The FHE-compliant fitted model.
         """
+        # For now we don't use X for quantization, only the thresholds
+        # We could  support import with data as quantizer too
```
jfrery, line 1848: I am not sure what this comment refers to? Are we changing the way we fit trees?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,212 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+# Some snippet to do this:
+# if n_bits > 17:
+#     diff = init_tensor - init_tensor_as_int
+#     max_diff = numpy.abs(diff).max()
+#     if max_diff > 1e-3:
+#         raise ValueError(f"{max_diff=} > 1e-4")
+
+
+# pylint: disable-next=too-many-locals,too-many-statements
+# TODO: make this pass with rounding (high-bit-width create overflow)
```
jfrery, line 1151: Some todo to remove / add FIXME

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,212 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+        # Sometimes, we miss convergence, which is not a problem for our test
+        warnings.simplefilter("ignore", category=ConvergenceWarning)
+        concrete_model, sklearn_model = concrete_model.fit_benchmark(x, y)
+
+    # This step is needed in order to handle partial classes
+    model_class = get_model_class(model_class)
+    max_n_bits = 18
+    reasonable_n_bits = 8
+
+    # TODO: add normal bit-width comparison
```
jfrery, line 1188: FIXME or remove

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,212 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+    if isinstance(concrete_model, BaseTreeClassifierMixin):
+        for n_bits, cml_tolerance, sklearn_tolerance in [
+            (max_n_bits, 1e-1, 1e-7),
+            (reasonable_n_bits, 6e-2, 6e-2),
```
jfrery, line 1192: Can we have these values at the top of the file?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,212 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+                value = numpy.abs(loaded_accuracy_from_data_accuracy - cml_accuracy)
+                assert (
+                    value < cml_tolerance
+                ), f"{loaded_accuracy_from_data_accuracy=} != {cml_accuracy} ({value})"
+
+    # Regressor
+    else:
+        for n_bits, cml_tolerance, sklearn_tolerance in [
+            (max_n_bits, 0.8, 1e-5),
+            (reasonable_n_bits, 1.4, 1.4),
```
jfrery, line 1276: Same, could we have that at the top of the file?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,212 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+                )
+            with subtests.test(
+                msg="Regression Sklearn vs Data", n_bits=n_bits, tolerance=sklearn_tolerance
+            ):
+                value = numpy.abs(loaded_mse_from_data_mse - sklearn_mse) / numpy.abs(y).max()
+                assert (
+                    value < sklearn_tolerance
+                ), f"{loaded_mse_from_data_mse=} != {sklearn_mse} ({value=}>={sklearn_tolerance=})"
+
+            # # Compare with Concrete ML
```
jfrery, line 1329:
```suggestion
            # Compare with Concrete ML
```

docs/advanced_examples/DecisionTreeClassifier.ipynb
```diff
@@ -169,16 +182,15 @@
      "text": [
       "Number of test samples: 691\n",
       "Number of spams in test samples: 304\n",
-      "True Negative (legit mail well classified) rate: 0.9612403100775194\n",
-      "False Positive (legit mail classified as spam) rate: 0.03875968992248062\n",
+      "True Negative (legit mail well classified) rate: 0.9302325581395349\n",
+      "False Positive (legit mail classified as spam) rate: 0.06976744186046512\n",
```
jfrery, line 186: We lose a bit of accuracy here?

docs/advanced_examples/DecisionTreeClassifier.ipynb
```diff
@@ -279,15 +291,15 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Execution time: 0.53 seconds per sample\n"
+      "Execution time: 3.75 seconds per sample\n"
```
jfrery, line 294: We probably need to refresh notebook on the aws machine.

docs/advanced_examples/DecisionTreeClassifier.ipynb
```diff
@@ -300,8 +312,8 @@
      "output_type": "stream",
      "text": [
       "Ground truth:       [0 0 0 1 0 1 0 0 0 0]\n",
-      "Prediction sklearn: [0 0 0 1 0 1 0 0 0 0]\n",
-      "Prediction FHE:     [0 0 0 1 0 1 0 0 0 0]\n"
+      "Prediction sklearn: [0 0 0 0 0 0 0 1 0 0]\n",
+      "Prediction FHE:     [0 0 0 0 0 0 0 1 0 0]\n"
```
jfrery, line 316: Hmm that's bad.

docs/advanced_examples/DecisionTreeClassifier.ipynb
```diff
@@ -332,6 +344,234 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "cv = False\n",
+    "\n",
+    "for model_name in [\"tree\", \"rf\", \"xgb\"]:\n",
```
jfrery, line 456: This is the decisiontreeclassifier notebook so not sure this should be here. However I like the visualization. We should probably have this in the documentation or in a new notebook.

jfrery (changes requested):
Thanks! I have a few comments / questions.

pyproject.toml
```diff
@@ -64,18 +64,19 @@ pylint = "^2.13.0"
 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/2541
 pytest = "7.4.1"
 pytest-cov = "^4.1.0"
+pytest-xdist = "^3.3.1"
+pytest-randomly = "^3.11.0"
+pytest-repeat = "^0.9.1"
+pytest-subtests = "^0.11.0"
```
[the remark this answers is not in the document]
fd0r: I just regrouped `pytest-*` dependencies

src/concrete/ml/quantization/quantizers.py
```diff
@@ -797,7 +797,7 @@ def dequant(self, qvalues: numpy.ndarray) -> Union[numpy.ndarray, Tracer]:
 
         values = self.scale * (qvalues - numpy.asarray(self.zero_point, dtype=numpy.float64))
 
-        assert isinstance(values, (numpy.ndarray, Tracer))
+        assert isinstance(values, (float, int, numpy.ndarray, Tracer))
```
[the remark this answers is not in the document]
RomanBredehoft: yeah don't think this is a good idea, if it can be avoided

pyproject.toml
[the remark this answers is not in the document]
jfrery: Isn't subtest new? Never used it.

docs/advanced_examples/DecisionTreeClassifier.ipynb
RomanBredehoft: does not seem very good indeed

pyproject.toml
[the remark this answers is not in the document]
RomanBredehoft: I think it's used to run only part of a test , or something like that, right @fd0r ?

src/concrete/ml/onnx/onnx_impl_utils.py
```diff
@@ -5,7 +5,7 @@
 import numpy
 from concrete.fhe import conv as fhe_conv
 from concrete.fhe import ones as fhe_ones
-from concrete.fhe import round_bit_pattern
+from concrete.fhe import truncate_bit_pattern
```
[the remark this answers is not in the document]
RomanBredehoft: yes this is weird right @fd0r ?

src/concrete/ml/sklearn/base.py
```diff
@@ -1328,6 +1339,279 @@ def __init__(self, n_bits: Union[int, Dict[str, int]]):
 
         BaseEstimator.__init__(self)
 
+    # TODO: FIX EXACT PREDICTION WITH HIGH BIT WIDTH
```
RomanBredehoft, line 1342: todo to remove

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,212 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+        warnings.simplefilter("ignore", category=ConvergenceWarning)
+        concrete_model, sklearn_model = concrete_model.fit_benchmark(x, y)
+
+    # This step is needed in order to handle partial classes
+    model_class = get_model_class(model_class)
+    max_n_bits = 18
+    reasonable_n_bits = 8
+
+    # TODO: add normal bit-width comparison
+    if isinstance(concrete_model, BaseTreeClassifierMixin):
```
RomanBredehoft, line 1189: if you are checking if this model is. a classifier, better use our `is_classifier_or_partial_classifier` utility tool !

pyproject.toml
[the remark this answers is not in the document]
fd0r: yes subtest is new. it allows us to check multiple things before finishing a test.

src/concrete/ml/onnx/onnx_impl_utils.py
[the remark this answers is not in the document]
fd0r: It was related to https://github.com/zama-ai/concrete-ml-internal/issues/4143

tests/sklearn/test_sklearn_models.py
fd0r: Doesn't look clearer to me tbh

But I can change it if you want.

RomanBredehoft: not clearer indeed, but safer / more robust imo

pyproject.toml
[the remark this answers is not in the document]
jfrery: got it thanks!

src/concrete/ml/onnx/onnx_impl_utils.py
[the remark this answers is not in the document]
jfrery: But are we sure it works fine? I understood truncate was not doing exactly what we wanted for the comparison.

src/concrete/ml/sklearn/base.py
fd0r: removed

docs/advanced_examples/DecisionTreeClassifier.ipynb
fd0r: We should probably seed notebooks?

docs/advanced_examples/DecisionTreeClassifier.ipynb
fd0r: No idea why, I didn't change anything that would affect that.

We could probably seed notebooks

tests/sklearn/test_sklearn_models.py
fd0r: changed

tests/sklearn/test_sklearn_models.py
fd0r: We can but it doesn't really make sense imo

src/concrete/ml/sklearn/base.py
fd0r: It's legacy comment I believe, I can remove it

src/concrete/ml/onnx/onnx_impl_utils.py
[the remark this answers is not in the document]
fd0r: To be confirmed by @kcelia , but previously we needed to had remove some offset to simulate truncation.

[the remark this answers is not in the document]
jfrery: Hmm alright. I though moving our trees to truncate was a big move but if all tests pass with truncate now then fine. Have you removed all the work around that rounding made us do? (e.g. the half trick and so on?)

src/concrete/ml/quantization/quantizers.py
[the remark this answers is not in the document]
jfrery: I would remove it if not needed yes.

src/concrete/ml/sklearn/base.py
jfrery: tree_to_numpy would be good. Maybe you can even rename this file to something more generic.

docs/advanced_examples/DecisionTreeClassifier.ipynb
jfrery: I am pretty sure this is seeded. At least, these results never changed since Jan 2022 so something changed here.

src/concrete/ml/onnx/onnx_impl_utils.py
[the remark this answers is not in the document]
fd0r: I removed it here for the comparison thing but not anywhere else.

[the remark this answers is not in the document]
jfrery: Alright well. If it work then that's good. I would still prefer to have this in a specific PR to make it clearer but as you want.

src/concrete/ml/sklearn/base.py
jfrery: Thanks

RomanBredehoft:
that's a test I've added recently, I might have an idea of why this happens actually, I'll let you open an issue @fd0r  and I'll see if I can reproduce it

src/concrete/ml/quantization/quantizers.py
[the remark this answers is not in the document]
fd0r: Update, this is actually needed.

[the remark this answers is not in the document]
jfrery: Why would we dequantize a floating point?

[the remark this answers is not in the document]
RomanBredehoft: yes something seems fishy here, maybe there's a unexpected `squeeze` somewhere / forgotten `unsqueeze` or similar ?

fd0r:
@RomanBredehoft , I already have a fix, just adding some tests as we speak

fd0r:
https://github.com/zama-ai/concrete-ml/pull/692

src/concrete/ml/onnx/onnx_impl_utils.py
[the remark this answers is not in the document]
fd0r: Maybe @andrei-stoian-zama and @kcelia can confirm that everything is good from this point.

src/concrete/ml/quantization/quantizers.py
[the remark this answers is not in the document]
fd0r: I'll try keeping only `int` indeed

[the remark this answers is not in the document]
fd0r: Okay you made me doubt but this is the dequantized value that we are checking here, so it's normal that we have floats here.

fd0r:
Quick summary of things to do here:
- Verify the source of the degradation of the model of the notebook.
- Move most of the code to `tree_to_numpy`
- Validate the move to `truncate_bit_pattern`

anything else @RomanBredehoft @jfrery ?

jfrery:
> Verify the source of the degradation of the model of the notebook.

+ why have the target value changed? They never changed since we started this notebook
+ remove rf xgb from decision tree classifier notebook I suppose.

src/concrete/ml/onnx/onnx_impl_utils.py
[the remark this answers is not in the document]
RomanBredehoft: would be great indeed !

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,200 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+    os.environ["TREES_USE_ROUNDING"] = str(int(use_rounding))
+
+    x, y = get_dataset(
+        model_class, parameters, min(N_BITS_REGULAR_BUILDS), load_data, is_weekly_option
+    )
+
+    if verbose:
+        print("Run check_load_pre_trained_sklearn_models")
+
+    assert issubclass(model_class, BaseTreeEstimatorMixin)
```
RomanBredehoft, line 1164: same here, safer to use : 
```
 if is_model_class_in_a_list(
            model_class,
            _get_sklearn_tree_models(),
        )
```

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,200 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+    concrete_model = instantiate_model_generic(model_class, n_bits=min(N_BITS_REGULAR_BUILDS))
+    # Fit the model and retrieve both the Concrete ML and the scikit-learn models
+    with warnings.catch_warnings():
+        # Sometimes, we miss convergence, which is not a problem for our test
+        warnings.simplefilter("ignore", category=ConvergenceWarning)
+        concrete_model, sklearn_model = concrete_model.fit_benchmark(x, y)
+
+    # This step is needed in order to handle partial classes
+    model_class = get_model_class(model_class)
```
RomanBredehoft, line 1173: if you use `is_model_class_in_a_list` in the following, you don't need to worry about partial classes (that's one of the benefit of using the util func)

src/concrete/ml/quantization/quantizers.py
```diff
@@ -773,7 +773,7 @@ def quant(self, values: numpy.ndarray) -> numpy.ndarray:
 
         return qvalues.astype(numpy.int64)
 
-    def dequant(self, qvalues: numpy.ndarray) -> Union[numpy.ndarray, Tracer]:
+    def dequant(self, qvalues: numpy.ndarray) -> Union[float, int, numpy.ndarray, Tracer]:
```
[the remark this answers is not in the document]
fd0r: The thing is that I'm cheating a bit while using this function and providing a python int instead of a numpy array

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,200 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+            loaded_from_data = model_class.from_sklearn_model(
+                sklearn_model,
+                X=x,
+                n_bits=n_bits,
+            )
+
+            # Compile both the initial Concrete ML model and the loaded one
+            concrete_model.compile(x)
+            mode = "disable"
+            if n_bits <= 8:
```
RomanBredehoft, line 1281: we have constants like `N_BITS_THRESHOLD_FOR_CRT_FHE_CIRCUITS` in our tests if that's what you had in mind here

src/concrete/ml/sklearn/base.py
```diff
@@ -1328,6 +1339,266 @@ def __init__(self, n_bits: Union[int, Dict[str, int]]):
...
+        # scikit-learn models
+        # This should be fixed once Concrete ML models initialize their underlying scikit-learn
+        # models during initialization
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3373
+        # Needed for XGB
+
+        cls_signature = inspect.signature(cls)
+        init_params_keys = list(init_params.keys())
+        for key in init_params_keys:  # pragma: no cover
+            if key not in cls_signature.parameters:
```
RomanBredehoft, line 1387: why is it mandatory to do this ? couldn't we just do `cls(n_bits=n_bits, **init_params)` like for linear models ? especially since we now check that all sklearn attributes are supported in CML (thanks to @jfrery )

src/concrete/ml/sklearn/base.py
```diff
@@ -1328,6 +1339,266 @@ def __init__(self, n_bits: Union[int, Dict[str, int]]):
...
+
+        # Compute input/threshold quantizers
+        input_quantizers: List[UniformQuantizer] = []
+
+        # Quantization of each feature in X
+        for feature_index in range(n_features):
+
+            # Get all thresholds for a given feature
+            threshold_for_feature: numpy.ndarray = bias_1[weight_1[:, feature_index] == 1][:, 0]
+            # Sorting threshold values makes things easier afterwards
```
RomanBredehoft, line 1435: better to have a newline above

src/concrete/ml/sklearn/base.py
```diff
@@ -1328,6 +1339,266 @@ def __init__(self, n_bits: Union[int, Dict[str, int]]):
...
+            )
+        )
+
+        # Tree values pre-processing
+        # i.e., mainly predictions quantization
+        # but also rounding the threshold such that they are now integers
+        model._set_post_processing_params()
+
+        # Get the expected number of ONNX outputs in the sklearn model.
+        expected_number_of_outputs = 1 if is_regressor_or_partial_regressor(model) else 2
```
RomanBredehoft, line 1524: is this always true for classifiers ?

src/concrete/ml/sklearn/base.py
RomanBredehoft: renaming the file would technically be a breaking change, it has already been included in the list btw : https://github.com/zama-ai/concrete-ml-internal/issues/4159

src/concrete/ml/sklearn/base.py
```diff
@@ -1328,6 +1339,266 @@ def __init__(self, n_bits: Union[int, Dict[str, int]]):
...
 
+    # pylint: disable=too-many-locals,too-many-statements,too-many-branches
+    @classmethod
+    def from_sklearn_model(
+        cls,
+        sklearn_model: sklearn.base.BaseEstimator,
+        X: Optional[numpy.ndarray] = None,
+        n_bits: int = 8,
+    ):
+        """Build a FHE-compliant model using a fitted scikit-learn model.
```
RomanBredehoft, line 1350: I would try to divide the following function into smaller functions if possible, as it's not always easy to follow what's going on !

RomanBredehoft (changes requested):
added some new comments, but overall I agree with @jfrery's ones

src/concrete/ml/sklearn/base.py
fd0r: Most of this stuff is copy-pasted fromt tree to numpy so I guess it's ok. Also all tests pass so 🤷🏼

tests/sklearn/test_sklearn_models.py
fd0r: Actually that won't really work with trees since we have rounding.

So I don't think it's possible from parameters only to known if the circuit will use the CRT representation.

src/concrete/ml/sklearn/base.py
RomanBredehoft: got it thanks !

src/concrete/ml/quantization/quantizers.py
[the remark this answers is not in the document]
RomanBredehoft: not sure we would want to keepsuch change if if it's possible tho, I guess this is related to the other similar change @jfrery saw

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,200 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+                value = numpy.abs(loaded_accuracy_from_data_accuracy - cml_accuracy)
+                assert (
+                    value < cml_tolerance
+                ), f"{loaded_accuracy_from_data_accuracy=} != {cml_accuracy} ({value})"
+
+    # Regressor
+    else:
+        for n_bits, cml_tolerance, sklearn_tolerance in [
+            (max_n_bits, 0.8, 1e-5),
+            (reasonable_n_bits, 1.8, 1.8),
```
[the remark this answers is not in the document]
fd0r: Mainly to check that increasing the number of bits we get better results.

They thresholds were chosen by trial and error. (easiest way I find)

tests/sklearn/test_sklearn_models.py
fd0r: alright makes sense thanks

src/concrete/ml/sklearn/base.py
jfrery: This file is internal. Why would it break anything from the API POV?

tests/sklearn/test_sklearn_models.py
fd0r: On second thought I need the actual model class to be able to call `from_sklearn_model` 🤷🏼

tests/sklearn/test_sklearn_models.py
fd0r: I disagree but I'll change it

tests/sklearn/test_sklearn_models.py
fd0r: Ho yeah I should use `reasonable_n_bits` here

src/concrete/ml/sklearn/base.py
RomanBredehoft: because it's not explicitly internal (we haven't put any underscored in front of it)

so while I agree that it won't be a great risk to rename it as I doubt some users are really using it, it's still a breaking change (hence the "technically" from my answers above !). And I don't see any good reasons to rename this in a urgent manner, so imo waiting for the next major release is fine

tests/sklearn/test_sklearn_models.py
RomanBredehoft: ah I see, indeed you have no choice here then !

tests/sklearn/test_sklearn_models.py
fd0r: changed

docs/advanced_examples/DecisionTreeClassifier.ipynb
fd0r: I reverted the notebook

docs/advanced_examples/DecisionTreeClassifier.ipynb
fd0r: reverted

src/concrete/ml/sklearn/base.py
fd0r: Moved everything to tree_to_numpy.

src/concrete/ml/sklearn/base.py
fd0r: Fine with me too to wait

src/concrete/ml/sklearn/base.py
fd0r: Actually this wasn't needed

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,206 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+
+            loaded_from_data = model_class.from_sklearn_model(
+                sklearn_model,
+                X=x,
+                n_bits=n_bits,
+            )
+
+            # Compile both the initial Concrete ML model and the loaded one
+            concrete_model.compile(x)
+            mode = "disable"
```
andrei-stoian-zama, line 1200: I think you should use simulate here

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,206 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
```
andrei-stoian-zama, line 1286: again, simulate would be better

docs/advanced_examples/ImportingFromScikitLearn.ipynb
andrei-stoian-zama, line 1: Here we should show nice graphs with decision boundaries like in ClassifierComparison. We shouldn't show n_bits variation of accuracy, just show good configs. Ideally the default configs (dont set n_bits in import_sklearn) should be used and they should work well (except PTQ, where you should set n_bits, we don't have a good default for that)

docs/advanced_examples/XGBClassifier.ipynb
andrei-stoian-zama, line 1: Can you add a section where you import the xgboost sklearn classifier to compare to the CML xgb one?

src/concrete/ml/sklearn/base.py
```diff
@@ -1328,6 +1335,80 @@ def __init__(self, n_bits: Union[int, Dict[str, int]]):
 
         BaseEstimator.__init__(self)
 
+    @classmethod
+    def from_sklearn_model(
+        cls,
+        sklearn_model: sklearn.base.BaseEstimator,
+        X: Optional[numpy.ndarray] = None,
+        n_bits: int = 8,
```
andrei-stoian-zama, line 1343: so 8 bits is ok? I remember it wasn't always the best choice in the graphs and we had discussed 9 bits. What is the accuracy on the regressors ?

andrei-stoian-zama (changes requested):
In addition to the comments above, can you please:
- improve the importingfromscikitlearn notebook to make it work exactly like classifier comparison
- add comparison of imported xgbregressor in the https://github.com/zama-ai/concrete-ml/blob/main/docs/advanced_examples/XGBRegressor.ipynb notebook

src/concrete/ml/onnx/onnx_impl_utils.py
```diff
@@ -265,9 +265,16 @@ def rounded_comparison(
...
     # To determine if 'x' 'operation' 'y' (operation being <, >, >=, <=), we evaluate 'x - y'
-    rounded_subtraction = round_bit_pattern((x - y) - half, lsbs_to_remove=lsbs_to_remove)
+    # We cast to int because if half is too high the result might be float
+    # intermediate = ((x - y) - half)
+    # intermediate_as_int = intermediate.astype(numpy.int64)
+    #
+    # if not isinstance(intermediate, Tracer):
+    #     assert (intermediate == intermediate_as_int).all()
+
+    rounded_subtraction = truncate_bit_pattern(x - y, lsbs_to_remove=lsbs_to_remove)
```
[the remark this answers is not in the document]
RomanBredehoft: I see,  I guess I would be more convinced if we test this more than once (like running the associated tests 10 or 100 times) to be sure it does not make them flaky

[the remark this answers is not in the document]
jfrery: My only concern here is that we merge this truncate feature in the middle of a load from sklearn feature. If we could split this into 2 PR it would be cleaner I think.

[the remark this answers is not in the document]
jfrery: Discussed offline. Looks like truncate is required for the tests. Good for me to merge as long as we have an appropriate commit message.

src/concrete/ml/onnx/onnx_impl_utils.py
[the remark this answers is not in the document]
jfrery: Discussed offline. Looks like truncate is required for the tests. Good for me to merge as long as we have an appropriate commit message.

jfrery (dismissed):
Looks good to me. Thanks!

RomanBredehoft (dismissed):
thanks a lot !!

src/concrete/ml/quantization/quantizers.py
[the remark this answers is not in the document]
fd0r: I can remove int from the result of dequant

tests/sklearn/test_sklearn_models.py
fd0r: Ho yeah forgot to activate simulation mode with reasonable number of bits

tests/sklearn/test_sklearn_models.py
fd0r: same

src/concrete/ml/quantization/quantizers.py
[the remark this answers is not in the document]
RomanBredehoft: I would prefer both personally, I don't see why we should allow dequant to return a float (worst, an int) and not a numpy.array !

fd0r:
The notebooks can still be improved but I think I addressed most comments.

RomanBredehoft (approved):

src/concrete/ml/sklearn/tree_to_numpy.py
```diff
@@ -512,3 +524,227 @@ def get_lsbs_to_remove_for_trees(array: numpy.ndarray) -> int:
...
+
+    assert q_y is not None
+    output_quantizers = [q_y.quantizer]
+
+    if enable_rounding:
+        # Quantize data to compute lsbs to remove
+        if model_inputs is None:
+            # If we have no data we can just randomly generate a dataset
+            assert isinstance(n_features, int)
+            calibration_set_size = 100_000
```
andrei-stoian-zama, line 718: do you really need 100 000 sampels here ?

tests/sklearn/test_sklearn_models.py
```diff
@@ -1120,6 +1137,208 @@ def check_exposition_structural_methods_decision_trees(model, x, y):
...
+    parameters,
+    use_rounding,
+    load_data,
+    is_weekly_option,
+    verbose=True,
+):
+    """Test `from_sklearn_model` functionnality of tree-based models."""
+
+    numpy.random.seed(0)
+    os.environ["TREES_USE_ROUNDING"] = str(int(use_rounding))
```
andrei-stoian-zama, line 1155: are you sure this works?

andrei-stoian-zama (approved):
Looks good, just a few questions!

src/concrete/ml/sklearn/tree_to_numpy.py
fd0r: Probably not, it was to be safe when developing, I can reduce this

tests/sklearn/test_sklearn_models.py
fd0r: yes it does

src/concrete/ml/sklearn/tree_to_numpy.py
fd0r: Changed it to 1000

fd0r:
I got to rebase on `main`

jfrery (approved):

PR #690, opened 2024-05-23
feat: support encrypted mul div

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4418
closes https://github.com/zama-ai/concrete-ml-internal/issues/4163

ref https://github.com/zama-ai/concrete-internal/issues/716

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -533,7 +533,7 @@ pytest_codeblocks_one:
 .PHONY: shell_lint # Lint all bash scripts
 shell_lint:
 	@# grep -v "^\./\." is to avoid files in .hidden_directories
-	find . -type f -name "*.sh" | grep -v "^\./\." | \
+	find . -type f -name "*.sh" | grep -v "^\./\." | grep -v "^\./playground/" | \
 	xargs shellcheck
 
 .PHONY: set_version # Generate a new version number and update all files with it accordingly
--- a/Makefile
+++ b/Makefile
@@ -533,7 +533,7 @@ pytest_codeblocks_one:
 .PHONY: shell_lint # Lint all bash scripts
 shell_lint:
 	@# grep -v "^\./\." is to avoid files in .hidden_directories
-	find . -type f -name "*.sh" | grep -v "^\./\." | grep -v "^\./playground/" | \
+	find . -type f -name "*.sh" | grep -v "^\./\." | \
 	xargs shellcheck
 
 .PHONY: set_version # Generate a new version number and update all files with it accordingly
--- a/src/concrete/ml/quantization/base_quantized_op.py
+++ b/src/concrete/ml/quantization/base_quantized_op.py
@@ -257,6 +257,9 @@ class QuantizedOp:
             if attribute_name not in metadata:
                 metadata[attribute_name] = attribute_value
 
+        # Sort the metadata to be sure it is always saved in the same order
+        metadata = dict(sorted(metadata.items()))
+
         return metadata
 
     @staticmethod
--- a/src/concrete/ml/quantization/quantized_ops.py
+++ b/src/concrete/ml/quantization/quantized_ops.py
@@ -1592,8 +1592,8 @@ class QuantizedDiv(QuantizedMixingOp):
         **kwargs,
     ) -> None:
         super().__init__(*args, rounding_threshold_bits=rounding_threshold_bits, **kwargs)
-        self.divider_quantizer: UniformQuantizer
-        self.min_non_zero_value: numpy.float64
+        self.divider_quantizer: Optional[UniformQuantizer] = None
+        self.min_non_zero_value: Optional[numpy.float64] = None
 
     def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
         """Create corresponding QuantizedArray for the output of the activation function.
@@ -1605,9 +1605,11 @@ class QuantizedDiv(QuantizedMixingOp):
             numpy.ndarray: the output values for the provided calibration samples.
         """
 
-        # If two inputs, compute the divider
+        # If the op can not be fused and that the two inputs are not constants
+        # we need to compute the quantizer of the divider since we are doing
+        # an encrypted division where both numerator and denominator are encrypted
         if not self.can_fuse() and len(inputs) == 2:
-            min_non_zero_value = numpy.min(numpy.abs(inputs[1]))
+            min_non_zero_value = inputs[1][numpy.abs(inputs[1]).argmin()]
 
             # mypy
             assert min_non_zero_value is not None and min_non_zero_value > 0
@@ -1627,13 +1629,8 @@ class QuantizedDiv(QuantizedMixingOp):
         **attrs,
     ) -> ONNXOpInputOutputType:
 
-        # If the op can be fused or if any input
-        # is a RawOpOutput, perform the op in the clear
-        if (
-            len(q_inputs) == 1
-            or any(isinstance(input, RawOpOutput) for input in q_inputs)
-            or self.can_fuse()
-        ):
+        # If the op can be fused we perform the op in the clear
+        if self.can_fuse():
             return super().q_impl(*q_inputs, **attrs)
 
         # For mypy
@@ -1657,13 +1654,17 @@ class QuantizedDiv(QuantizedMixingOp):
         # Dequantize
         input_1 = q_input_1.dequant()
 
-        # Replace input_1 with min_non_zero_qvalue if input_1 is 0
+        # Replace input_1 with min_non_zero_value if input_1 is 0
+        # mypy
+        assert self.min_non_zero_value is not None
         input_1 = numpy.where(input_1 == 0, self.min_non_zero_value, input_1)
 
         # Compute the inverse of input_1
         input_1_inv = 1.0 / input_1
 
         # Re-quantize the inverse using the same quantization parameters as q_input_1
+        # mypy
+        assert self.divider_quantizer is not None
         q_input_1_inv_rescaled = self.divider_quantizer.quant(input_1_inv)
 
         # The product of quantized encrypted integer values
@@ -1700,6 +1701,7 @@ class QuantizedDiv(QuantizedMixingOp):
             # Apply Concrete rounding (if relevant)
             product_q_values = self.cnp_round(product_q_values, calibrate_rounding)
 
+        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4546
         # De-quantize the product
         dequant_product = (product_q_values - new_zero_point) * new_scale
 
@@ -1714,27 +1716,10 @@ class QuantizedDiv(QuantizedMixingOp):
             params=self.output_quant_params,
         )
 
-    def dump_dict(self) -> Dict:
-        metadata = super().dump_dict()
-        if hasattr(self, "min_non_zero_value"):
-            metadata["min_non_zero_value"] = self.min_non_zero_value
-        if hasattr(self, "divider_quantizer"):
-            metadata["divider_quantizer"] = self.divider_quantizer
-        return metadata
-
-    @staticmethod
-    def load_dict(metadata: Dict):
-        obj = super(QuantizedDiv, QuantizedDiv).load_dict(metadata)
-        if "min_non_zero_value" in metadata:
-            obj.min_non_zero_value = metadata.pop("min_non_zero_value")
-        if "divider_quantizer" in metadata:
-            obj.divider_quantizer = metadata.pop("divider_quantizer")
-        return obj
-
     def can_fuse(self) -> bool:
         """Determine if this op can be fused.
 
-        Add operation can be computed in float and fused if it operates over inputs produced
+        Div operation can be computed in float and fused if it operates over inputs produced
         by a single integer tensor.
 
         Returns:
@@ -1850,7 +1835,7 @@ class QuantizedMul(QuantizedMixingOp):
     def can_fuse(self) -> bool:
         """Determine if this op can be fused.
 
-        Add operation can be computed in float and fused if it operates over inputs produced
+        Mul operation can be computed in float and fused if it operates over inputs produced
         by a single integer tensor.
 
         Returns:
--- a/src/concrete/ml/quantization/quantized_ops.py
+++ b/src/concrete/ml/quantization/quantized_ops.py
@@ -1577,14 +1577,139 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
     _impl_for_op_named: str = "Or"
 
 
-class QuantizedDiv(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
-    """Div operator /.
+class QuantizedDiv(QuantizedMixingOp):
+    """Quantized Division operator.
 
-    This operation is not really working as a quantized operation. It just works when things got
-    fused, as in e.g., Act(x) = 1000 / (x + 42))
+    Can divide either two variables (both encrypted) or a variable and a constant
     """
 
     _impl_for_op_named: str = "Div"
+
+    def __init__(
+        self, *args, rounding_threshold_bits: int | Dict[str, str | int] | None = None, **kwargs
+    ) -> None:
+        super().__init__(*args, rounding_threshold_bits=rounding_threshold_bits, **kwargs)
+        self.divider_quantizer: UniformQuantizer
+        self.max_qvalue: int
+
+    def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
+        """Create corresponding QuantizedArray for the output of the activation function.
+
+        Args:
+            *inputs (numpy.ndarray): Calibration sample inputs.
+
+        Returns:
+            numpy.ndarray: the output values for the provided calibration samples.
+        """
+        q_array_divider = QuantizedArray(self.n_bits, 1 / inputs[1], is_symmetric=False)
+        self.divider_quantizer = q_array_divider.quantizer
+
+        # Get the max qvalue that will be used to handle division by zero in quantized
+        self.max_qvalue = numpy.max(q_array_divider.qvalues)
+        return super().calibrate(*inputs)
+
+    def q_impl(
+        self,
+        *q_inputs: ONNXOpInputOutputType,
+        calibrate_rounding: bool = False,
+        **attrs,
+    ) -> ONNXOpInputOutputType:
+
+        # Ensure q_inputs has exactly 2 elements
+        assert len(q_inputs) == 2, "QuantizedDiv should always have 2 elements in q_inputs"
+
+        # If the second input is a RawOpOutput, perform the op in the clear
+        if isinstance(q_inputs[1], RawOpOutput):
+            prepared_inputs = self._prepare_inputs_with_constants(
+                *q_inputs, calibrate=False, quantize_actual_values=False
+            )
+            return self.call_impl(*prepared_inputs, **attrs).view(RawOpOutput)
+
+        # For mypy
+        assert self.output_quant_params is not None
+        assert self.output_quant_params.scale is not None
+        assert self.output_quant_params.zero_point is not None
+
+        prepared_inputs = self._prepare_inputs_with_constants(
+            *q_inputs, calibrate=False, quantize_actual_values=True
+        )
+
+        q_input_0: QuantizedArray = prepared_inputs[0]
+        q_input_1: QuantizedArray = prepared_inputs[1]
+
+        assert q_input_0.quantizer.scale is not None
+        assert q_input_0.quantizer.zero_point is not None
+
+        assert q_input_1.quantizer.scale is not None
+        assert q_input_1.quantizer.zero_point is not None
+
+        # Dequantize
+        input_1 = q_input_1.dequant()
+
+        # Replace input_1 with max_qvalue if input_1 is 0
+        input_1 = numpy.where(input_1 == 0, self.max_qvalue, input_1)
+
+        # Compute the inverse of input_1
+        input_1_inv = 1.0 / input_1
+
+        # Re-quantize the inverse using the same quantization parameters as q_input_1
+        q_input_1_inv_rescaled = self.divider_quantizer.quant(input_1_inv)
+
+        # The product of quantized encrypted integer values
+        product_q_values = q_input_0.qvalues * q_input_1_inv_rescaled
+
+        # Integer quantized multiplication need adjustment based on the zero points.
+        if q_input_0.quantizer.zero_point:
+            product_q_values -= q_input_0.quantizer.zero_point * (
+                q_input_1.qvalues - q_input_1.quantizer.zero_point
+            )
+        if q_input_1.quantizer.zero_point:
+            product_q_values -= q_input_1.quantizer.zero_point * (
+                q_input_0.qvalues - q_input_0.quantizer.zero_point
+            )
+
+        # mypy
+        assert self.divider_quantizer.scale is not None
+        assert self.divider_quantizer.zero_point is not None
+
+        # Compute the scale and zero point based on the scale and zero point
+        # of the two quantized values multiplied together
+        new_scale = q_input_0.quantizer.scale * self.divider_quantizer.scale
+        new_zero_point = q_input_0.quantizer.zero_point * self.divider_quantizer.zero_point
+
+        if self.produces_graph_output:
+            return self.make_output_quant_parameters(product_q_values, new_scale, new_zero_point)
+
+        q_product = QuantizedArray(
+            n_bits=self.n_bits,
+            values=product_q_values,
+            value_is_float=False,
+            options=self._get_output_quant_opts(),
+            stats=self.output_quant_stats,
+            params=UniformQuantizationParameters(
+                scale=new_scale,
+                zero_point=new_zero_point,
+                offset=0,
+            ),
+        )
+
+        with tag(self.op_instance_name + ".rounding"):
+            # Apply Concrete rounding (if relevant)
+            product_q_values = self.cnp_round(product_q_values, calibrate_rounding)
+
+        # De-quantize the product
+        dequant_product = q_product.dequant()
+
+        # Return the raw float values without re-quantizing them to the new scale, as any
+        # following Gemm/Add/Conv will quantize them with _prepare_inputs_with_constants(...)
+        return QuantizedArray(
+            self.n_bits,
+            dequant_product,
+            value_is_float=True,
+            options=self._get_output_quant_opts(),
+            stats=self.output_quant_stats,
+            params=self.output_quant_params,
+        )
 
 
 class QuantizedMul(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
--- a/src/concrete/ml/quantization/quantized_ops.py
+++ b/src/concrete/ml/quantization/quantized_ops.py
@@ -1609,10 +1609,11 @@ class QuantizedDiv(QuantizedMixingOp):
         # we need to compute the quantizer of the divider since we are doing
         # an encrypted division where both numerator and denominator are encrypted
         if not self.can_fuse() and len(inputs) == 2:
-            min_non_zero_value = inputs[1][numpy.abs(inputs[1]).argmin()]
+            min_non_zero_index = numpy.abs(inputs[1]).argmin(axis=None)
+            min_non_zero_value = inputs[1].flat[min_non_zero_index]
 
             # mypy
-            assert min_non_zero_value is not None and min_non_zero_value > 0
+            assert min_non_zero_value is not None and min_non_zero_value != 0
             self.min_non_zero_value = min_non_zero_value
 
             q_array_divider = QuantizedArray(self.n_bits, 1 / inputs[1])
--- a/src/concrete/ml/quantization/quantized_ops.py
+++ b/src/concrete/ml/quantization/quantized_ops.py
@@ -1609,6 +1609,8 @@ class QuantizedDiv(QuantizedMixingOp):
         # we need to compute the quantizer of the divider since we are doing
         # an encrypted division where both numerator and denominator are encrypted
         if not self.can_fuse() and len(inputs) == 2:
+
+            # FIXME https://github.com/zama-ai/concrete-ml-internal/issues/4556
             min_non_zero_index = numpy.abs(inputs[1]).argmin(axis=None)
             min_non_zero_value = inputs[1].flat[min_non_zero_index]
 
@@ -1666,6 +1668,7 @@ class QuantizedDiv(QuantizedMixingOp):
         # Re-quantize the inverse using the same quantization parameters as q_input_1
         # mypy
         assert self.divider_quantizer is not None
+        # FIXME https://github.com/zama-ai/concrete-ml-internal/issues/4556
         q_input_1_inv_rescaled = self.divider_quantizer.quant(input_1_inv)
 
         # The product of quantized encrypted integer values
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -532,7 +532,7 @@ pytest_codeblocks_one:
 .PHONY: shell_lint # Lint all bash scripts
 shell_lint:
 	@# grep -v "^\./\." is to avoid files in .hidden_directories
-	find . -type f -name "*.sh" | grep -v "^\./\." | \
+	find . -type f -name "*.sh" | grep -v "^\./\." | grep -v "^\./playground/" | \
```
RomanBredehoft, line 535: I had the same issue, and changed my `playground` directory with `.playground` (which I then added in the gitignore), I think it's a better approach

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -257,6 +257,9 @@ def dump_dict(self) -> Dict:
             if attribute_name not in metadata:
                 metadata[attribute_name] = attribute_value
 
+        # Sort the metadata to be sure it is always saved in the same order
+        metadata = dict(sorted(metadata.items()))
```
RomanBredehoft, line 261: worth doing the same for built-in models if you think there's a real point of having that .. but it's going to be many copy-paste

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        """Create corresponding QuantizedArray for the output of the activation function.
+
+        Args:
+            *inputs (numpy.ndarray): Calibration sample inputs.
+
+        Returns:
+            numpy.ndarray: the output values for the provided calibration samples.
+        """
+
+        # If two inputs, compute the divider
```
RomanBredehoft, line 1608: could you clarify why this is necessary and in which case we are using this ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        if not self.can_fuse() and len(inputs) == 2:
+            min_non_zero_value = numpy.min(numpy.abs(inputs[1]))
+
+            # mypy
+            assert min_non_zero_value is not None and min_non_zero_value > 0
+            self.min_non_zero_value = min_non_zero_value
+
+            q_array_divider = QuantizedArray(self.n_bits, 1 / inputs[1])
```
RomanBredehoft, line 1616: is that handling both `const/enc` and `enc/const` cases ? if so, could you explain why in the comment ? if not, worth handling this and add a test for it then

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        return super().calibrate(*inputs)
+
+    def q_impl(
+        self,
+        *q_inputs: ONNXOpInputOutputType,
+        calibrate_rounding: bool = False,
+        **attrs,
+    ) -> ONNXOpInputOutputType:
+
+        # If the op can be fused or if any input
```
RomanBredehoft, line 1630: could you explain why ? like what does `len(q_inputs) == 1` mean, same for `any(isinstance(input, RawOpOutput) for input in q_inputs)`ect please ? it'll be easier to debug in the future, cause here I'm not entirely sure why we are checking these

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        assert q_input_0.quantizer.scale is not None
+        assert q_input_0.quantizer.zero_point is not None
+
+        assert q_input_1.quantizer.scale is not None
+        assert q_input_1.quantizer.zero_point is not None
+
+        # Dequantize
+        input_1 = q_input_1.dequant()
+
+        # Replace input_1 with min_non_zero_qvalue if input_1 is 0
```
RomanBredehoft, line 1660: `min_non_zero_qvalue`should be `min_non_zero_value`

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft, line 1616: also, how can we be sure that `inputs[1]`is not 0 ? I feel like `1 / inputs[1]` can fail here no ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        input_1 = numpy.where(input_1 == 0, self.min_non_zero_value, input_1)
```
RomanBredehoft, line 1661: how is this working ? just replacing 0 with the (float) min ? I imagine that means that values are expected to be positive ! And then still, should we instead add `self.min_non_zero_value` to ìnput_1`, instead of replacing 0, right ? 

in any case we should explain in comments how all this work

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+
+        with tag(self.op_instance_name + ".rounding"):
+            # Apply Concrete rounding (if relevant)
+            product_q_values = self.cnp_round(product_q_values, calibrate_rounding)
+
+        # De-quantize the product
+        dequant_product = (product_q_values - new_zero_point) * new_scale
+
+        # Return the raw float values without re-quantizing them to the new scale, as any
+        # following Gemm/Add/Conv will quantize them with _prepare_inputs_with_constants(...)
```
RomanBredehoft, line 1707: but no issue if the div is at a graph's end / if the following ops are like 'reshape' / 'squeeze' ect ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        input_1 = numpy.where(input_1 == 0, self.min_non_zero_value, input_1)
+
+        # Compute the inverse of input_1
+        input_1_inv = 1.0 / input_1
+
+        # Re-quantize the inverse using the same quantization parameters as q_input_1
+        q_input_1_inv_rescaled = self.divider_quantizer.quant(input_1_inv)
+
+        # The product of quantized encrypted integer values
+        product_q_values = q_input_0.qvalues * q_input_1_inv_rescaled
```
RomanBredehoft, line 1670: all of the following steps are doing an multiplication between 2 quantizers, don't we already have that somewhere in the codebase that we could re-use ? instead of having to implement all ZP and scale adjustments manually again like here I mean ! it could make things less error prone

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        return QuantizedArray(
+            self.n_bits,
+            dequant_product,
+            value_is_float=True,
+            options=self._get_output_quant_opts(),
+            stats=self.output_quant_stats,
+            params=self.output_quant_params,
+        )
+
+    def dump_dict(self) -> Dict:
```
RomanBredehoft, line 1717: imo you don't need this, it should be handled in the `QuantizedOps`'s `dump_dict`  method automatically

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        metadata = super().dump_dict()
+        if hasattr(self, "min_non_zero_value"):
+            metadata["min_non_zero_value"] = self.min_non_zero_value
+        if hasattr(self, "divider_quantizer"):
+            metadata["divider_quantizer"] = self.divider_quantizer
+        return metadata
+
+    @staticmethod
+    def load_dict(metadata: Dict):
```
RomanBredehoft, line 1726: same I think it's not necessary

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
 
+    def q_impl(
+        self,
+        *q_inputs: ONNXOpInputOutputType,
+        calibrate_rounding: bool = False,
+        **attrs,
+    ) -> ONNXOpInputOutputType:
+
+        # If either input is a RawOpOutput or if the op can be fused,
+        # perform the op in the TLU using FP32
```
RomanBredehoft, line 1763: same as above

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        q_input_0: QuantizedArray = prepared_inputs[0]
+        q_input_1: QuantizedArray = prepared_inputs[1]
+
+        assert q_input_0.quantizer.scale is not None
+        assert q_input_0.quantizer.zero_point is not None
+
+        assert q_input_1.quantizer.scale is not None
+        assert q_input_1.quantizer.zero_point is not None
+
+        # Remove the manual encrypted multiplication when we
```
RomanBredehoft, line 1790: so how's the following different than the `enc_mul` we already have ? same as above, couldn't the following be re-used from somewhere else (like `QuantizedGemm` or `enc_mul`) ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+            # Apply Concrete rounding (if relevant)
+            product_q_values = self.cnp_round(
+                product_q_values, calibrate_rounding, rounding_operation_id="product_q_values"
+            )
+
+        # De-quantize the product
+        dequant_product = (product_q_values + new_zero_point) * new_scale
+
+        # Return the raw float values without re-quantizing them to the new scale, as any
+        # following Gemm/Add/Conv will quantize them with _prepare_inputs_with_constants(...)
```
RomanBredehoft, line 1840: same comment as above, no issue if at the end of the graph or before any reshape/squeeze ect operators ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+            value_is_float=True,
+            options=self._get_output_quant_opts(),
+            stats=self.output_quant_stats,
+            params=self.output_quant_params,
+        )
+
+    def can_fuse(self) -> bool:
+        """Determine if this op can be fused.
+
+        Add operation can be computed in float and fused if it operates over inputs produced
```
RomanBredehoft, line 1853: same as above, not the add operator + some explanations would be great

tests/quantization/test_quantized_ops.py
```diff
@@ -363,8 +363,9 @@ def test_all_arith_ops(
         # Compute the quantized operator result
         quantized_output_vv = q_op(q_inputs_0, q_inputs_1).dequant()
 
-        # Check the R2 of raw output and quantized output
-        check_r2_score(raw_output_vv, quantized_output_vv)
+        if n_bits > 16:
```
RomanBredehoft, line 366: are we sure about this change ?

tests/quantization/test_quantized_ops.py
```diff
@@ -409,16 +410,19 @@ def test_all_arith_ops(
     check_r2_score(raw_output_cv, quantized_output_cv)
 
     # Check that we get the same fp32 results in V+V (if supported), V+C and C+V modes
-    if supports_enc_with_enc:
+    if supports_enc_with_enc and n_bits > 16:
```
RomanBredehoft, line 413: same

tests/quantization/test_quantized_ops.py
```diff
@@ -409,16 +410,19 @@ def test_all_arith_ops(
...
+    if operator is not QuantizedDiv:
+        check_float_array_equal(raw_output_cv, raw_output_vc)
+
+        # Check that V+C and C+V is symmetric (int+float mode)
+        check_float_array_equal(quantized_output_cv, quantized_output_vc)
 
     # As V+C and C+V work on float values they will not be exactly equal to
     # the V+V case which works in quantized, we only check R2 for a high bit-width in this case
-    if supports_enc_with_enc:
+    if supports_enc_with_enc and n_bits > 16:
```
RomanBredehoft, line 425: same

RomanBredehoft (changes requested):
thanks for this feature ! I have several questions and observations, mostly about adding more comments to make the code more clear

src/concrete/ml/quantization/quantized_ops.py
jfrery: I do this right before:

```python

File: src/concrete/ml/quantization/quantized_ops.py (hunk 8 above)
Comment (jfrery) [reply src/concrete/ml/quantization/quantized_ops.py:1707]: We have that:

```python
        if self.produces_graph_output:
            return self.make_output_quant_parameters(product_q_values, new_scale, new_zero_point)
```

src/concrete/ml/quantization/quantized_ops.py
jfrery: We could refactor QuantizedDiv and QuantizedMul to put the code they share in some place

src/concrete/ml/quantization/quantized_ops.py
jfrery: It doesn't. The check serialization test fails because the constants are not exported / re-imported

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -1577,25 +1577,288 @@ class QuantizedOr(QuantizedOpUnivariateOfEncrypted, QuantizedOp):
...
+        if "min_non_zero_value" in metadata:
+            obj.min_non_zero_value = metadata.pop("min_non_zero_value")
+        if "divider_quantizer" in metadata:
+            obj.divider_quantizer = metadata.pop("divider_quantizer")
+        return obj
+
+    def can_fuse(self) -> bool:
+        """Determine if this op can be fused.
+
+        Add operation can be computed in float and fused if it operates over inputs produced
```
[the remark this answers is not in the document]
jfrery: Mistake I will correct

src/concrete/ml/quantization/quantized_ops.py
jfrery: We could.

src/concrete/ml/quantization/quantized_ops.py
jfrery: same answer we have a if condition to return the integer if needed

src/concrete/ml/quantization/quantized_ops.py
andrei-stoian-zama: I wonder if some of these conditions are redundant.. shouldn't can_fuse handle all cases ?

src/concrete/ml/quantization/quantized_ops.py
jfrery: yes thanks

src/concrete/ml/quantization/quantized_ops.py
andrei-stoian-zama: no, this is standard

andrei-stoian-zama (approved):
If the tests pass, this looks good

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: ok good to me then !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: ah yes right, I missed the assert. But does that mean that we always expect values to be strictly positive ? how can we ensure that ?

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: that's weird

tests/quantization/test_quantized_ops.py
RomanBredehoft: not sure I see where we were doing the second chck 🤔

jfrery: `        check_r2_score(raw_output_vv, quantized_output_vv)` is right below the if

tests/quantization/test_quantized_ops.py
jfrery: We were checking correctness for low bit-width. Worked fine for univariate but now QuantizedMul|Div have a quant requant step which impact the correctness.

src/concrete/ml/quantization/quantized_ops.py
jfrery: I will double check. But I feel like you are right.

tests/quantization/test_quantized_ops.py
jfrery: Ah yeah what am saying -_-'. Sorry. So yes the problem is as I said, correctness were fine before with univariate LUT but in QuantizedDiv|Mul we have quant requant which impact the final accuracy.

src/concrete/ml/quantization/quantized_ops.py
andrei-stoian-zama: @RomanBredehoft very well spotted!

Makefile
jfrery: got it thanks!

RomanBredehoft: I would remove this added part to avoid future issues then !

src/concrete/ml/quantization/quantized_ops.py
jfrery: What I could check is that no value are quant dequant to a 0 maybe

src/concrete/ml/quantization/quantized_ops.py
jfrery: Had a closer look. Variables were not initialized so hasattr that we do in the loat_dict didn't catch it. It's fine now.

src/concrete/ml/quantization/base_quantized_op.py
fd0r: Why not done with an `OrderedDict` if the order is so important?

Makefile
jfrery: Yes done thanks

RomanBredehoft: I still see it 🤔

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: I don't think it's important, but I guess it was for debugging purposes

Soptq:
Can we merge this PR? I am working on compiling a model that involves encrypted mul, currently it prompts "AssertionError: Do not support this type of operation between encrypted tensors". I think this PR could fix this error.

jfrery:
> Can we merge this PR? I am working on compiling a model that involves encrypted mul, currently it prompts "AssertionError: Do not support this type of operation between encrypted tensors". I think this PR could fix this error.

Yes sure. This PR has been open for too long already. We will merge it soon.

src/concrete/ml/quantization/quantized_ops.py
jfrery: I updated

RomanBredehoft: not sure I see the update 😅

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: yes but my question is more about "how can you be sure of that ?" !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: so what should we do here then ?

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: worth an issue + fixme then !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: same as above, why couldn't we do : 
```
if any(isinstance(input, RawOpOutput) for input in q_inputs) or self.can_fuse():
```

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: same as above, worth an issue + fixme then

tests/quantization/test_quantized_ops.py
RomanBredehoft: do we know by how much it can impact ? how "strong" do we expect this impact to be ? I'm just being a bit suspicious here because of things like https://github.com/zama-ai/concrete-ml/pull/690/files#r1634728461 or https://github.com/zama-ai/concrete-ml/pull/690/files#r1634723604 😅

tests/quantization/test_quantized_ops.py
RomanBredehoft: I answered again above (https://github.com/zama-ai/concrete-ml/pull/690/files#r1677535203)

tests/quantization/test_quantized_ops.py
RomanBredehoft: same : https://github.com/zama-ai/concrete-ml/pull/690/files#r1677535203

RomanBredehoft (changes requested):
sorry, I still have some remaining questions !

src/concrete/ml/quantization/quantized_ops.py
jfrery: Sorry I use to write as I do it locally and then click the re-request review. So I didn't push yet^^

src/concrete/ml/quantization/quantized_ops.py
jfrery: I am sorry I have lost the context here.

> But does that mean that we always expect values to be strictly positive ?

We check the min abs (`numpy.min(numpy.abs(inputs[1]))`). Not sure why the strictly positive.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Actually, here we are within a PBS and we assign values that are 0 to self.min_non_zero_value. That should work fine within a PBS.

src/concrete/ml/quantization/quantized_ops.py
jfrery: I will clarify what's going on here. But I think it's due to the fact that we check QuantizedOps without calls to the `post_training.py` which is where we define the `can_fuse`. So I probably need to modify some test if I want to use a simple condition for all qops.

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: you are right we do the abs, so my questions is then : how can we be sure that we never get 0 in `inputs[1]`  ? I understand that we assert or could add more assert, but I don't get how we are sure that this/these assert(s) will never ever fail in any circumstances (and if we can't, that's an issue imo) !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: yes but my observation was that, since we replace all zeros with `self.min_non_zero_value` : 
- since this value is computed through `numpy.min(numpy.abs(inputs[1]))` , does that mean we expect `input_1` to always be positive ? 
    - if not, for example, `[-2, 4, 0]` would be replaced by `[-2, 4, 2]` and I don't think it makes much sense
    - if so, how can we be sure that `input_1` is `>= 0` ? more that having asserts I mean (like my comment in https://github.com/zama-ai/concrete-ml/pull/690/files#r1677747936)
- also, aren't we breaking the values' distribution by doing something like this ? initially I suggested to instead do something like `input_1 += min_non_zero_value` (or `input_1 += ` if positive) but don't think that's desirable as well. What about requantizing on [1, max_val] ? Anyway, I feel just replacing 0 by this new val is not exactly right

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: not sure what you're saying, my observation was that here you basically do 
```
if len(q_inputs) == 1 or any(isinstance(input, RawOpOutput) for input in q_inputs) or self.can_fuse():
``` 
instead of 
```
if any(isinstance(input, RawOpOutput) for input in q_inputs) or self.can_fuse():
```
like above

so more or less : is this `len(q_inputs) == 1` necessary ?

src/concrete/ml/quantization/quantized_ops.py
jfrery: We can't. If the user provides a 0 it will fail. As any division fails when there are 0 in the denominator.

RomanBredehoft: ah I see, in that case if it's a user issue, then we should have a `raise ValueError` with an explicit message rather than an empty assert. Unless an error is triggered before reaching this point (and in that case, is it explicit enough ?)

jfrery: The assert here is for mypy. Not sure we should rewrite basic error message such as division by 0.

src/concrete/ml/quantization/base_quantized_op.py
jfrery: It's important for our tests where we check that multiple serialization are the same.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Ah yes you are right good catch! I should keep the sign of the value. I fixed it.

src/concrete/ml/quantization/quantized_ops.py
jfrery: https://github.com/zama-ai/concrete-ml-internal/issues/4546

src/concrete/ml/quantization/quantized_ops.py
jfrery: Let's see what the test say.

jfrery:
@RomanBredehoft could you re-review this please.

src/concrete/ml/quantization/quantized_ops.py
jfrery: I changed to only use can fuse. Seems to work fine.

src/concrete/ml/quantization/quantized_ops.py
jfrery: Done

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: well I won't push too much on this one, but I just fear that the error will not be a clear "DivisionByZeroError" and might confuse the user, would be worth a check or even a test

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: hmm well although that fixes the issue, I don't think it's right though now ? eg if we have a `[-20, 3, 0, 4]`, your code changes it to `[-20, 3, -20, 4]`, which drastically changes the values' distribution ⚠️  This is basically what I was trying to say in my second point : 
> also, aren't we breaking the values' distribution by doing something like this ? initially I suggested to instead do something like input_1 += min_non_zero_value (or input_1 +=  if positive) but don't think that's desirable as well. What about requantizing on [1, max_val] ? Anyway, I feel just replacing 0 by this new val is not exactly right

imo what we are doing here is still wrong !

src/concrete/ml/quantization/quantized_ops.py
RomanBredehoft: thanks for the issue, would be great to add the fixme as well in our code so we don't forget !

RomanBredehoft (changes requested):
still think something is wrong sorry !

src/concrete/ml/quantization/quantized_ops.py
jfrery: > hmm well although that fixes the issue, I don't think it's right though ? eg if we have a [-20, 3, 0, 4], your code changes it to [-20, 3, -20, 4], which drastically changes the values' distribution ⚠️ This is basically what I was trying to say in my second point 

This input_1 are floating points. Here it would change `[-20, 3, 0, 4]` to `[-20, 3, 3, 4]`. Why ``[-20, 3, 3, 4]`. We don't take the min we take min of abs.

Another way we could do this is to take the smallest floating point represented in the quantized version of inputs[1]. But we don't have the quant information for this so I am not sure this is robust.

We can discuss this offline if needed.

jfrery: I totally understand your worry on this however! Let's find the best solution.

RomanBredehoft: oh ok sorry I don't know why I kept seeing `argmax` and not `argmin`, so yeah ok we just get the closest value to 0, makes much more sense indeed

RomanBredehoft: would it make sense to instead add an epsilon to this 0 ? or simply to all values ? not sure it'll change much since we re-quantize after the 1/x

apart from that, I'm not sure I have better ideas here

jfrery: > would it make sense to instead add an epsilon to this 0 ?

Yes that's what I proposed. An epsilon would make sense if it's represented by the quantized values. The scale is basically the smallest represented floating point. So we could use this instead of the min(abs(x)). But I am not entirely sure we can get the actual quantizer of the input. Also I am afraid that for large bit-width this epsilon is going to be super small and will lead to numerical errors.

RomanBredehoft: yeah numerical errors could be an issue indeed. But actually this epsilon should be added to all values not just the 0 maybe ? in order to keep the same distribution ? not sure how the quant params will handle this though

tests/quantization/test_quantized_ops.py
RomanBredehoft, line 366: also, would be better if we change this condition to `if n_bits > 16 and isinstance(operator, (QuantizedDiv, QuantizedMul)):` since it seems to only impact these ops ? or maybe only div ?

jfrery: I updated to have r2 n_bits threshold for each qops

tests/quantization/test_quantized_ops.py
jfrery: updated to have tight n_bits threshold for each qops

RomanBredehoft (approved):

RomanBredehoft (approved):

PR #692, opened 2024-05-23
fix: fix fhe-training classes behavior

fd0r:
We had an issue with the way we supported classes, and more specifically in the `partial_fit` method.

We now have the same behavior as scikit-learn.

### DIFF
--- a/src/concrete/ml/sklearn/linear_model.py
+++ b/src/concrete/ml/sklearn/linear_model.py
@@ -235,6 +235,7 @@ class SGDClassifier(SklearnSGDClassifierMixin):
 
         # Checks and warnings for FHE training
         if self.fit_encrypted:
+            self.classes_: Optional[numpy.ndarray] = None
 
             warnings.warn(
                 "FHE training is an experimental feature. Please be aware that the API might "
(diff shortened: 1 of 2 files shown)


src/concrete/ml/sklearn/linear_model.py
```diff
@@ -421,26 +425,42 @@ def _fit_encrypted(
...
+                self.label_encoder.fit(classes)
+                self.classes_ = numpy.array(classes)
+            # If partial fit we should probably support adding classes along the way
 
             assert isinstance(self.classes_, numpy.ndarray)
 
             # Allow the training set to only provide a single class. This can happen, for example,
             # when running 'partial_fit' on a small batch of values. Even with a single class, the
             # model remains binary
-            if len(self.classes_) not in [1, 2]:
```
RomanBredehoft, line 443: not but this was a real feature, check the above comment : 
>             # Allow the training set to only provide a single class. This can happen, for example,
            # when running 'partial_fit' on a small batch of values. Even with a single class, the
            # model remains binary

basically, if the first `partial_fit` only has a single calss, we should support that since the model remains binary

fd0r: Actually the previous behavior was wrong. We still support partial fit with only one class in the training batch. In the case of calls to `partial_fit` the argument `classes` should be provided, as is the case in scikit-learn

RomanBredehoft (approved):
great thanks !!

PR #694, opened 2024-05-23
chore: fix weekly workflow

RomanBredehoft:
without the secrets, the usual CI workflow could not run


.github/workflows/weekly_tests.yaml
```diff
@@ -18,5 +18,6 @@ jobs:
   weekly-tests:
     name: Run weekly tests
     uses: ./.github/workflows/continuous-integration.yaml
+    secrets: inherit
```
RomanBredehoft, line 21: the actual fix

jfrery (approved):

PR #662, opened 2024-04-30
chore: add flaky retry in weekly CI

RomanBredehoft:
Clean the weekly CI : 
- new workflow : scheduled + on demand , with dedicated status history
- re-run flaky issues like in regular PRs (+ report them on slack notifications for weeklys)
- remove slack reports for regular CIs (keep for weekly + push on main/release branches)
- better slack report message (ex: https://zama-ai.slack.com/archives/C02RX6CF3FV/p1715090858656719)

refs https://github.com/zama-ai/concrete-ml-internal/issues/4030
refs https://github.com/zama-ai/concrete-ml-internal/issues/4428
closes https://github.com/zama-ai/concrete-ml-internal/issues/4390

PR #698, opened 2024-05-27
chore: fix flaky re-run step in CI

RomanBredehoft:
this is following work done in https://github.com/zama-ai/concrete-ml/pull/662, it should fix the flaky re-run step (https://github.com/zama-ai/concrete-ml/actions/runs/9226803784/job/25387474616)

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -735,7 +735,7 @@ jobs:
           	poetry run python ./script/actions_utils/pytest_failed_test_report.py \
             --pytest-input-report "pytest_report.json" \
             --failed-tests-report "failed_tests_report.json" \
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -735,7 +735,7 @@ jobs:
           	poetry run python ./script/actions_utils/pytest_failed_test_report.py \
             --pytest-input-report "pytest_report.json" \
             --failed-tests-report "failed_tests_report.json" \
-            --failed-tests-comment "failed_tests_comment_${{ matrix.python_version }}.txt"
+            --failed-tests-comment "failed_tests_comment_${{ matrix.python_version }}.txt" \
```
fd0r, line 738: famous invisible bug

nice catch

fd0r (approved):
lgtm thanks for the fix

PR #711, opened 2024-06-03
docs: add a license FAQ in README

yuxizama:


README.md
```diff
@@ -249,7 +249,23 @@ To contribute to Concrete ML, please refer to [this section of the documentation
 
 ### License
 
-This software is distributed under the **BSD-3-Clause-Clear** license. If you have any questions, please contact us at hello@zama.ai.
+This software is distributed under the **BSD-3-Clause-Clear** license. Read [this](LICENSE) for more details.
```
bcm-at-zama, line 252: shouldn't this FAQ be for every products, and so in a common place?

andrei-stoian-zama (commented):
Thanks! I think this helps to clear up many things. 

I have a few improvement suggestions.

README.md
yuxizama: It is for every product. We have a long blog post explaining it (as a common place) that people don't find it easily and read it. 
So we put a simple version on the README of each repo. (it's also Rand's idea to have in on GitHub repo)

README.md
```diff
@@ -249,7 +249,23 @@ To contribute to Concrete ML, please refer to [this section of the documentation
...
+
+#### FAQ
+
+**Is Zama’s technology free to use?**
+>Zama’s libraries are free to use under the BSD 3-Clause Clear license only for development, research, prototyping, and experimentation purposes. However, for any commercial use of Zama's open source code, companies must purchase Zama’s commercial patent license.
+>
+>Everything we do is open source and we are very transparent on what it means for our users, you can read more about how we monetize our open source products at Zama in [this blog post](https://www.zama.ai/post/open-source).
+
+**What do I need to do if I want to use Zama’s technology for commercial purposes?**
+>To commercially use Zama’s technology you need to be granted Zama’s patent license. Please contact us at hello@zama.ai for more information.
```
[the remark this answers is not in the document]
yuxizama: This part is validated by Valentina. As it concerns legal matters, it's better for us to not modify it.

yuxizama:
@andrei-stoian-zama Thanks for the review! I've addressed your comments.

fd0r (approved):

README.md
bcm-at-zama: We have said that eg, we wanted to have a common doc for the noise / crypto, so why not starting this common doc for the license? Else, we'll have copies of this FAQ everywhere, and they'll start to differ at some point

yuxizama: I understand your point about having copies of FAQs everywhere. However, since we're talking about open source license issues, I think it's more conventional practice to include license info in GitHub's repo, instead of in documentation. We use READMEs as product pages, I think these are something people should know when they first discover the product, not when they're already using it. 

Regarding the common page for noise/crypto, I talked to @andrei-stoian-zama, we think maybe for Concrete ML (and blockchain) team they won't really need this level of detail, as the libraries are high-level abstractions. => It's something we need to discuss more to decide wether implement or not in Q3.

bcm-at-zama: >  I think it's more conventional practice to include license info in GitHub's repo, instead of in documentation

We can have links directly from the repo to the (common) file, we're not forced to link to doc. Actually, already in our GH .md's, we link to some other .md.

yuxizama: But we don't have a common repo for all the libraries, do we? And we also have the "license.txt" file under each repo.

bcm-at-zama: Yes we currently don't but we plan to have one for the common things, like the documentation about TFHE or noise or things like that. You said you would start such a thing to avoid copies between the different repositories and docs.

andrei-stoian-zama (approved):
Thanks!

README.md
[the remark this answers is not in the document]
andrei-stoian-zama: thanks for explaining it

README.md
andrei-stoian-zama: I'm guessing if there is a common one it should be copied over to all repos automatically with some script . This could be automated in the future.

We should merge this now and once there is a common file we can think about how to use a script to check that the version in each repo is the same as the common one. Alternatively, @yuxizama could simply check manually if the versions are in sync once per quarter when we push the doc.

Can we merge this @bcm-at-zama ?

bcm-at-zama: Well, you merge if you want, but I thought one of the goals was to avoid copies and here, we create one while we could directly do the right way

RomanBredehoft (approved):

PR #713, opened 2024-06-03
Ci/add slsa framework

aquint-zama:


.github/workflows/continuous-integration.yaml
fd0r, line 1: hard to read the diff due to the formatting but thanks for taking care of that

fd0r: alright just discovered the hide whitespaces of github diff

aquint-zama: you should review it by commit (I isolated the trailing ws one)

fd0r (approved):

RomanBredehoft (dismissed):
thanks for this !!

aquint-zama:
How could we test this workflow, is it easy to have a rc? cc @RomanBredehoft

fd0r:
we should probably have an option to do builds that target zama's pip index since we can't do a rc to pypi with CP's nightly being a requirements.

aquint-zama:
I prefer to have a planed test before merging otherwise we will have trouble at EOQ

fd0r (approved):
Let's merge and try to do a rc

RomanBredehoft (approved):

PR #719, opened 2024-06-06
chore: update gpt2 use case llm multi head attention run with rounding

jfrery:


use_case_examples/llm/QGPT2Evaluate.ipynb
```diff
@@ -270,24 +271,54 @@
     "output_logits_simulated = proj_single_head_qgpt2(input_ids).logits"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now we are ready to execute the single head attention in FHE"
```
fd0r, line 278: We shouldn't use "we"

use_case_examples/llm/QGPT2Evaluate.ipynb
```diff
@@ -270,24 +271,54 @@
...
-   "execution_count": 11,
+   "execution_count": 14,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Simulated logits are equal to the quantized clear ones: False\n"
+      "Time taken to execute the single head in FHE: 166.38 seconds\n"
```
fd0r, line 290: Yeah with InsertRounding we aren't equal anymore, best to show FHE runtime

fd0r (changes requested):
We should remove the comparison with clear values since they don't apply anymore with rounding.

use_case_examples/llm/QGPT2Evaluate.ipynb
```diff
@@ -270,24 +271,54 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Simulated logits are equal to FHE logits: False\n"
```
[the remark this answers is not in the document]
jfrery: This was correct for exact rounding but now it seems to be wrong with approx :/.

use_case_examples/llm/QGPT2Evaluate.ipynb
jfrery: Removed all the first person pronouns.

use_case_examples/llm/QGPT2Evaluate.ipynb
jfrery: We aren't equal to disable (expected) but there seem to be a different between simulate and fhe. Would that be expected?

jfrery:
> We should remove the comparison with clear values since they don't apply anymore with rounding.

I don't compare the prediction with clear values. I compare simulate vs fhe here.

fd0r:
> I compare simulate vs fhe here.

Still doesn't make sense in the approx mode imo

fd0r (approved):

PR #720, opened 2024-06-07
chore: fix pandas flaky test

RomanBredehoft:
this small flaky test made the weekly and other PRs) sometime fail : https://github.com/zama-ai/concrete-ml/actions/runs/9414309379/job/25933015157

### DIFF
--- a/tests/pandas/test_pandas.py
+++ b/tests/pandas/test_pandas.py
@@ -714,7 +714,9 @@ def check_invalid_schema_values():
     ):
         client.encrypt_from_pandas(pandas_df, schema=schema_float_column)
 
-    schema_float_oob = {f"{feat_name}_float_1": {"min": float_min // 2, "max": float_max // 2}}
+    # Here, we want to make sure that either min or max is out of the distribution. We make sure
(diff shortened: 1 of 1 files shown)


tests/pandas/test_pandas.py
```diff
@@ -714,7 +714,7 @@ def check_invalid_schema_values():
     ):
         client.encrypt_from_pandas(pandas_df, schema=schema_float_column)
 
-    schema_float_oob = {f"{feat_name}_float_1": {"min": float_min // 2, "max": float_max // 2}}
```
fd0r, line 717: I'm fine with that but could you maybe explain a little more what's happening here? Having min=max seems kinda weird.

fd0r (dismissed):
Some comment but fine with me

tests/pandas/test_pandas.py
RomanBredehoft: we want to always trigger then catch e error, so obviously we need weird configs like this 😅

RomanBredehoft: added a comment

fd0r (approved):
Perfect thanks!

PR #724, opened 2024-06-11
chore: clean and speed up fhe training tests

RomanBredehoft:
we are compiling the fhe training circuit for fhe execution but only do simulation, which should speed up the test by a x2 factor. Usually compilation is quick, but in the composition case it can take around tens of seconds so this can save several minutes of running

for that, I need to remove the fixed values we've put in our compile method, which are not breaking changes since these are Concrete's default values + should allow anyone to be able to only compile for simulation or fhe execution (as explained abvoe)

### DIFF
--- a/src/concrete/ml/deployment/fhe_client_server.py
+++ b/src/concrete/ml/deployment/fhe_client_server.py
@@ -470,6 +470,6 @@ class FHEModelClient:
 
         # In training mode, note that this step does not make much sense for now. Still, nothing
         # breaks since QuantizedModule don't do anything in post-processing
-        result = self.model.post_processing(*result)
+        result_post_processed = self.model.post_processing(*result)
 
-        return result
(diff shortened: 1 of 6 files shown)


src/concrete/ml/deployment/fhe_client_server.py
```diff
@@ -470,6 +470,6 @@ def deserialize_decrypt_dequantize(
 
         # In training mode, note that this step does not make much sense for now. Still, nothing
         # breaks since QuantizedModule don't do anything in post-processing
-        result = self.model.post_processing(*result)
+        result_post_processed = self.model.post_processing(*result)
```
RomanBredehoft, line 473: @jfrery

jfrery (approved):
Nice! Thanks for the cleaning

jfrery (approved):

PR #725, opened 2024-06-12
chore: add slack alert when stopping EC2 instances fails in CI

RomanBredehoft:
The last weekly CI failed for an unknown reason (internal error from GitHub), letting some EC2 instances running (which required to be manually terminated): https://github.com/zama-ai/concrete-ml/actions/runs/9468836256

this PR adds a slack alerts that will better warn us about such an issue  cc @fd0r

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -718,7 +718,7 @@ jobs:
           )
           && steps.conformance.outcome == 'success' 
           && !cancelled()
-          && false
+
         shell: bash +e {0}
         run: |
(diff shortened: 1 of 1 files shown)


fd0r (approved):
Great! Should we maybe add a ping to machinelearning to make sure that we don't miss it? Or even send it to the team slack channel and not the repo channel?

RomanBredehoft:
@fd0r I tried to tag machinlearing but was not able to : https://zama-ai.slack.com/archives/C01022E8SRJ/p1718195645050339

currently, it tags the slack whole channel where these notifications are sent, like here : https://zama-ai.slack.com/archives/C02RX6CF3FV/p1718269669499519

do you have something else in mind ?

PR #726, opened 2024-06-12
fix: key compression always

andrei-stoian-zama:
Enable key compression for all models

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -155,7 +155,6 @@ def default_configuration():
         fhe_simulation=False,
         fhe_execution=True,
         compress_input_ciphertexts=os.environ.get("USE_INPUT_COMPRESSION", "1") == "1",
-        compress_evaluation_keys=True,
     )
 
 
@@ -174,7 +173,6 @@ def simulation_configuration():
         fhe_simulation=True,
         fhe_execution=False,
         compress_input_ciphertexts=os.environ.get("USE_INPUT_COMPRESSION", "1") == "1",
-        compress_eval_keys=True,
     )
 
 
--- a/conftest.py
+++ b/conftest.py
@@ -155,7 +155,7 @@ def default_configuration():
         fhe_simulation=False,
         fhe_execution=True,
         compress_input_ciphertexts=os.environ.get("USE_INPUT_COMPRESSION", "1") == "1",
-        compress_eval_keys=True,
+        compress_evaluation_keys=True,
     )
 
 
--- a/conftest.py
+++ b/conftest.py
@@ -155,6 +155,7 @@ def default_configuration():
         fhe_simulation=False,
         fhe_execution=True,
         compress_input_ciphertexts=os.environ.get("USE_INPUT_COMPRESSION", "1") == "1",
+        compress_eval_keys=True,
     )
 
 
@@ -173,6 +174,7 @@ def simulation_configuration():
         fhe_simulation=True,
         fhe_execution=False,
         compress_input_ciphertexts=os.environ.get("USE_INPUT_COMPRESSION", "1") == "1",
+        compress_eval_keys=True,
     )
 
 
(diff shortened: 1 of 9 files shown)


src/concrete/ml/sklearn/linear_model.py
```diff
@@ -347,7 +347,7 @@ def _get_training_quantized_module(
         # Enable the underlying FHE circuit to be composed with itself
         # This feature is used in order to be able to iterate in the clear n times without having
         # to encrypt/decrypt the weight/bias values between each loop
-        configuration = Configuration(composable=True)
+        configuration = Configuration(composable=True, compress_evaluation_keys=True)
```
andrei-stoian-zama, line 350: compress pbs keys for training

tests/torch/test_hybrid_converter.py
```diff
@@ -40,6 +40,8 @@ def run_hybrid_llm_test(
     # Multi-parameter strategy is used in order to speed-up the FHE executions
     configuration = Configuration(
         single_precision=False,
+        compress_input_ciphertexts=True,
```
andrei-stoian-zama, line 43: these circuits can be fully leveled or not, always good to test with default compilation parameters

docker/release_resources/sanity_check.py
```diff
@@ -120,6 +122,8 @@ def function_to_compile(x):
         enable_unsafe_features=is_fast,
         use_insecure_key_cache=is_fast,
         insecure_key_cache_location=keyring_dir_as_str,
+        compress_input_ciphertexts=True,
```
andrei-stoian-zama, line 125: sanity checks should run with default compile parameters

conftest.py
```diff
@@ -173,6 +174,7 @@ def simulation_configuration():
         fhe_simulation=True,
         fhe_execution=False,
         compress_input_ciphertexts=os.environ.get("USE_INPUT_COMPRESSION", "1") == "1",
+        compress_eval_keys=True,
```
andrei-stoian-zama, line 177: tests should use the default for compressed keys, which is True

RomanBredehoft (changes requested):
imo we shouldn't force it like this but do like for input compression, meaning we use env variables. The main reason is that forcing it like this makes it impossible to disable it for tests or other things through input `configuration` objects. Basically this is what I reverted for fhe/simu exec in https://github.com/zama-ai/concrete-ml/pull/724 (which is going to conflict a bit with your PR btw) 

additionally we probably want to add tests to make sure that the compression properly works, like we did for input compression

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: which makes me thinkg, it should be using `enable_input_compression` as well here !

tests/torch/test_hybrid_converter.py
RomanBredehoft: so why not also enabling key compression ?

RomanBredehoft (approved):
thanks !! I think we could add a few small things in a follow up PR

tests/torch/test_hybrid_converter.py
andrei-stoian-zama: it is enabled by the QM compilation (the env variable defaults to 1)

src/concrete/ml/sklearn/linear_model.py
andrei-stoian-zama: yes, you're right, same as for DF which was not enabled before

tests/torch/test_hybrid_converter.py
RomanBredehoft: so is `compress_input_ciphertexts` then right ? imo either we put both here or we revert this small change 🤔

src/concrete/ml/sklearn/linear_model.py
RomanBredehoft: actually same as below, we should probably revert this change : it's emabled by QM's compilation anyway

PR #737, opened 2024-06-19
chore: change target of release candidates to private index

fd0r:
Pypi doesn't accept external dependencies so we need to change the target of our release candidates this they will use Concrete Python's versions that are not hosted on pypi anymore.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -308,6 +308,7 @@ jobs:
       PIP_EXTRA_INDEX_URL: ${{ secrets.PIP_EXTRA_INDEX_URL }}
       GIT_TAG: ${{ needs.release-checks.outputs.git_tag }}
       PROJECT_VERSION: ${{ needs.release-checks.outputs.project_version }}
+      IS_RC: ${{needs.release-checks.outputs.is_rc}}
 
     # Jobs are separated runners, we therefore need to install dependencies again in order to
     # pursue (without using cache or upload/download them as artifacts)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -594,7 +594,7 @@ jobs:
       # since release candidates and actual releases are public, we don't need to publish anything
```
RomanBredehoft, line 594: worth updating this comment if this change is meant to be permanent !

.github/workflows/release.yaml
```diff
@@ -584,7 +584,7 @@ jobs:
           docker image push --all-tags "${PUBLIC_RELEASE_IMAGE_BASE}"
 
       - name: Push package to PyPi
-        if: ${{ success() && !cancelled() }}
+        if: ${{ (release_checks.outputs.is_rc == 'false') && success() && !cancelled() }}
```
RomanBredehoft, line 587: small detail but maybe better to add 
```
env:
    IS_RC: ${{ needs.release-checks.outputs.is_rc }}
```
in this job and then call 
```
env.IS_RC == 'false' &&
```
like we do elsewhere, I find it a bit more clear and less error prone

.github/workflows/release.yaml
```diff
@@ -594,7 +594,7 @@ jobs:
       # since release candidates and actual releases are public, we don't need to publish anything
       # to the private internal repo
       - name: Push package to Internal PyPi
-        if: false
+        if: ${{ release_checks.outputs.is_rc == 'true' }}
```
RomanBredehoft, line 597: same

.github/workflows/release.yaml
fd0r: much better indeed! thanks for the suggestion!

A typo was detected thanks to that btw

RomanBredehoft (approved):

PR #738, opened 2024-06-20
feat: support resnet

jfrery:


andrei-stoian-zama (commented):
Looks good but:

- if we can't get the GPU time, maybe remove the GPU line
- add somewhere that rounding_threshold_bits for the configurations, next to the the n_bits dictionaries

fd0r:
CI failed too

andrei-stoian-zama (approved):
Great, it even has the GPU results

PR #739, opened 2024-06-20
chore: update Concrete Python to 2.7.0

RomanBredehoft:


pyproject.toml
```diff
@@ -34,7 +34,7 @@ readme = "README.md"
 # Investigate if it is better to fix specific versions or use lower and upper bounds  
 # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/2665
 python = ">=3.8.1,<3.11"
-concrete-python = {version="==2.8.0rc1", source = "zama-pypi-cpu"}
+concrete-python = {version="==2.7.0", source = "zama-pypi-cpu"}
```
bcm-at-zama, line 37: shouldn't you change that, @RomanBredehoft (to public pypi?)

RomanBredehoft: yes I read poetry's doc and I think you're right, I therefore added a comment

andrei-stoian-zama (approved):

PR #743, opened 2024-06-21
chore: remove sphinx doc management in release ci

RomanBredehoft:
We recently removed support for sphinx documentation (https://github.com/zama-ai/concrete-ml/pull/631), which means we no longer can upload sphinx html docs to the aws bucket

in particular, this recent change made the release ci fail : https://github.com/zama-ai/concrete-ml/actions/runs/9601275300/job/26479969530


fd0r (approved):
Not sure  what all these cloudfront stuff was supposed to do but looks good to me

PR #745, opened 2024-06-21
chore: move away from softprops/action-gh-release

fd0r:
softprops/action-gh-release was unreliable with SLSA files. We are moving to github CLI to create a release.


aquint-zama (approved):
lgtm

RomanBredehoft (approved):

PR #749, opened 2024-06-21
chore: fix slsa

fd0r:
Somehow what should be a file is a folder.

This workaround should allow the CI to work correctly but might not be the most elegant fix.


fd0r:
superseded by https://github.com/zama-ai/concrete-ml/pull/750

PR #761, opened 2024-06-25
fix: dynamic import of `transformers` in hybrid model

fd0r:
We currently have an import to `transformers` in the hybrid model source code but we don't have the library as a requirement, only a development dependency.

Importing the library dynamically should fix this.


jfrery (approved):
Thanks!

RomanBredehoft (dismissed):
thanks !

RomanBredehoft (approved):

PR #762, opened 2024-06-25
fix: dynamic import of `transformers` in hybrid model

fd0r:
We currently have an import to `transformers` in the hybrid model source code but we don't have the library as a requirement, only a development dependency.

Importing the library dynamically should fix this.

### DIFF
--- a/src/concrete/ml/torch/hybrid_model.py
+++ b/src/concrete/ml/torch/hybrid_model.py
@@ -18,7 +18,6 @@ import torch
 from brevitas.quant_tensor import QuantTensor
 from concrete.fhe import Configuration
 from torch import nn
-from transformers import Conv1D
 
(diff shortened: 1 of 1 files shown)


RomanBredehoft (dismissed):
thanks !

jfrery (dismissed):

jfrery:
Hmm we should have another model tested in the hybrid model to fix coverage

fd0r:
> Hmm we should have another model tested in the hybrid model to fix coverage

We would need to somehow modify dynamically the running python env to not use the transformers library.

I'll do a quick check to see if this is easy to implement but otherwise I'll just use a `pragma no cover`

RomanBredehoft (approved):

andrei-stoian-zama:
We decided to move the Conv1d stuff to the use case, we can close this PR, but thanks for the quick response

PR #763, opened 2024-06-25
chore: remove transformers import

jfrery:


src/concrete/ml/torch/hybrid_model.py
```diff
@@ -64,6 +63,37 @@ def underscore_str_to_tuple(tup: str) -> Tuple:
     return ast.literal_eval(tup.replace("po_", "(").replace("_pc", ")").replace("_", ", "))
 
 
+# FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/3858
+class Conv1D(nn.Module):
```
andrei-stoian-zama, line 67: I don't understand how this can work.. this is concrete.ml.torch.Conv1d and what you need to check with isinstance is transformers.Conv1D

andrei-stoian-zama (changes requested):

PR #765, opened 2024-06-26
chore: fix hybrid models

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/3905

cc @RomanBredehoft

### DIFF
--- a/src/concrete/ml/torch/hybrid_model.py
+++ b/src/concrete/ml/torch/hybrid_model.py
@@ -356,7 +356,7 @@ class HybridFHEModel:
             name: self._get_module_by_name(self.model, name) for name in self.module_names
         }
         self.remote_modules: Dict[str, RemoteModule] = {}
-        self.private_q_modules: Dict[QuantizedModule] = {}
+        self.private_q_modules: Dict[str, QuantizedModule] = {}
         self.configuration: Optional[Configuration] = None
         self.model_name = model_name
         self.verbose = verbose
@@ -508,12 +508,17 @@ class HybridFHEModel:
 
         model_path = Path(path)
         for module_name in self.module_names:
+            onnx_model = self.private_q_modules[  # pylint: disable=protected-access
+                module_name
+            ]._onnx_model
+
+            # mypy
+            assert onnx_model is not None
             input_shapes = [
                 tuple(elt.dim_value for elt in onnx_input.type.tensor_type.shape.dim)
-                for onnx_input in self.private_q_modules[  # pylint: disable=protected-access
-                    self.module_names[0]
-                ]._onnx_model.graph.input
+                for onnx_input in onnx_model.graph.input
             ]
+
             assert len(input_shapes) == 1, "Multi-input circuits not supported yet"
             model_module_path = model_path.resolve() / module_name
             model_module_path.mkdir(exist_ok=True)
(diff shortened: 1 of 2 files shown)


RomanBredehoft (dismissed):
thanks !

fd0r (changes requested):
The divergence in results was between torch and quantized?

I would still keep a torch mode here is instead of removing it entirely.

jfrery:
> The divergence in results was between torch and quantized?
> 
> I would still keep a torch mode here is instead of removing it entirely.

Not sure what you mean. I check the accuracy of simulation vs torch and disable vs torch.

Why would we keep a torch execution? We have the torch model for this

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -519,12 +508,17 @@ def _save_fhe_circuit(self, path: Path, via_mlir=False):
 
         model_path = Path(path)
         for module_name in self.module_names:
+            onnx_model = self.private_q_modules[  # pylint: disable=protected-access
```
RomanBredehoft, line 511: there is not need to access the private `_onnx_model`, `QuantizedModule` instances have a public property for them , you can directly call `onnx_model` normally

jfrery: Ah yes!

RomanBredehoft:
actually @jfrery , I think @fd0r is right we should keep the torch execution as well, but maybe through a new mode like `"torch"` or `"float"` ? 

because as I mentioned to you, currently the torch model is modified in place by the hybrid model. But that means, there's no real way of running the model in float after compiling it with the hybrid

jfrery:
> actually @jfrery , I think @fd0r is right we should keep the torch execution as well, but maybe through a new mode like `"torch"` or `"float"` ?
> 
> because as I mentioned to you, currently the torch model is modified in place by the hybrid model. But that means, there's no real way of running the model in float after compiling it with the hybrid

Very good point! We replace in-place so I am basically checking disable vs disable and simulate vs disable. I will add the torch inference.

That being said, simulate vs disable should raise the flaky we saw ...

src/concrete/ml/torch/hybrid_model.py
jfrery: Fixed

tests/torch/test_hybrid_converter.py
```diff
@@ -127,9 +126,9 @@ def run_hybrid_llm_test(
 @pytest.mark.parametrize(
     "list_or_str_private_modules_names, expected_accuracy, has_pbs",
     [
-        ("transformer.h.0.mlp", 0.934, True),
-        (["transformer.h.0.mlp", "transformer.h.1.mlp"], 0.42, True),
-        ("transformer.h.0.mlp.c_fc", 0.986, False),
+        ("transformer.h.0.mlp", 0.95, True),
+        (["transformer.h.0.mlp", "transformer.h.1.mlp"], 0.40, True),
```
RomanBredehoft, line 130: do we have an idea why this accuracy is so low then @jfrery ?

RomanBredehoft (approved):
thanks ! maybe wait for @fd0r feedback as well

tests/torch/test_hybrid_converter.py
jfrery: The "accuracy" here is the percent of similar labels in the top5 for all examples so a bit specific and I am not sure what to expect.

I think the rounding is really bad on MLP due to the outliers observed in LLMs. We might want to try smoothquant on gpt2 to see if this makes the rounding a bit less aggressive.

fd0r (approved):
LGTM

PR #773, opened 2024-06-28
chore: move away from `dev-dependencies`

fd0r:
`dev-dependencies` is deprecated from poetry.
This PR changes that.

Another benefit from it is that we should now be able to do: `python -m pip install concrete-ml[dev]` to install from pypi with all development dependencies.

closes https://github.com/zama-ai/concrete-ml-internal/issues/2005 
closes https://github.com/zama-ai/concrete-ml-internal/issues/2319 
closes https://github.com/zama-ai/concrete-ml-internal/issues/2685

chore: update licenses


jfrery (dismissed):

pyproject.toml
```diff
@@ -58,18 +58,32 @@ numpy = "1.23.5"
...
+name = "zama-pypi-gpu"
+url = "https://pypi.zama.ai/gpu"
+priority = "explicit"
+
+# GPU
+[tool.poetry.group.gpu]
+optional = true
+
+[tool.poetry.group.gpu.dependencies]
+concrete-python = {version="==2.7.0", source = "zama-pypi-gpu", markers = "sys_platform != 'darwin' and platform_machine != 'arm'"}
```
RomanBredehoft, line 77: do we want to add the gpu dependency in this exact PR ? feels like it should a PR on its own, with some associated tests

RomanBredehoft (changes requested):
thanks for the change ! Some questions, including why we are adding gpu deps here ! 

Also, for https://github.com/zama-ai/concrete-ml-internal/issues/2685, you can also update `script/make_utils/pytest_pypi_cml.sh`  (there's a fixme associated)

Makefile
```diff
@@ -49,7 +49,7 @@ setup_env:
 	if [[ $$(uname) != "Linux" ]] && [[ $$(uname) != "Darwin" ]]; then \
 		poetry install --only dev; \
```
RomanBredehoft, line 50: why do we have this actually ? is it for windows ? but why the `--only` ?

ci/aws_ami_build_component.yaml
```diff
@@ -43,7 +43,7 @@ phases:
             - virtualenv venv --python=python3.9
             - echo "source $(pwd)/venv/bin/activate" >> .bashrc
             - source venv/bin/activate
-            - python -m pip install concrete-ml==${CML_VERSION}
+            - python -m pip install "concrete-ml[dev]==${CML_VERSION}"
```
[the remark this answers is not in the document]
fd0r: Because we do sanity checks afterwards that require the dev deps.

Makefile
fd0r: I have no idea

pyproject.toml
```diff
@@ -58,18 +58,32 @@ numpy = "1.23.5"
...
 
-# Needed to have the latest Concrete Python versions
+# Zama sources
 [[tool.poetry.source]]
 name = "zama-pypi-cpu"
 url = "https://pypi.zama.ai/cpu"
 priority = "explicit"
 
-[tool.poetry.dev-dependencies]
+[[tool.poetry.source]]
```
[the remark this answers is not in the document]
RomanBredehoft: not sure about tests, but it could be a dedicated gh action that runs on a machine with gpu and where we install + activate the gpu option from Concrete 

still, I'm not sure adding gpu dependency in this pr makes much sense !

Makefile
fd0r: I checked with the autor and it's for windows support.
Not sure how tested this is tbh

RomanBredehoft (approved):
thanks !

PR #778, opened 2024-07-04
feat: support embedding layers

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4488

This allows us to use embedding layers in FHE by converting the embedding lookup into a onehot @ embedding_matrix operation.

The main change is the new `convert_first_gather_to_matmul` function in `preprocess_onnx_model` that replaces all Gather nodes by a a matmul and create the corresponding pre processing onnx that will be used to convert labels into one hot vectors.


src/concrete/ml/torch/compile.py
```diff
@@ -110,7 +110,7 @@ def build_quantized_module(
     # bit-width, but for tracing we only take a single one. We need the ONNX tracing batch size to
     # match the batch size during FHE inference which can only be 1 for the moment.
     dummy_input_for_tracing = tuple(
-        torch.from_numpy(val[[0], ::]).float() for val in inputset_as_numpy_tuple
+        torch.from_numpy(val[[0], ::]) for val in inputset_as_numpy_tuple
     )
```
jfrery, line 114: That change is breaking lots of tests. Not sure why this .float() is so important.

RomanBredehoft: what was the issue ?

jfrery: We need the data to be float32 here if there are float64 somehow. Otherwise torch isn't happy during tracing in the onnx dump.

RomanBredehoft: worth adding a comment about it right above so we remember it then !

src/concrete/ml/onnx/convert.py
```diff
@@ -235,13 +247,20 @@ def preprocess_onnx_model(onnx_model: onnx.ModelProto, check_model: bool) -> onn
             f"Available ONNX operators: {', '.join(sorted(IMPLEMENTED_ONNX_OPS))}"
         )
 
-    return equivalent_onnx_model
+    # Convert the first Gather node to a matrix multiplication with one-hot encoding
+    # In FHE, embedding is either a TLU or a matmul with a one-hot.
+    # The second case allows for leveled operation thus much faster.
+    onnx_preprocessing, equivalent_onnx_model = convert_first_gather_to_matmul(
```
RomanBredehoft, line 253: but we don't want that all the time though right ?

src/concrete/ml/onnx/convert.py
```diff
@@ -251,23 +270,39 @@ def get_equivalent_numpy_forward_from_onnx(
             Defaults to True.
 
     Returns:
-        Callable[..., Tuple[numpy.ndarray, ...]]: The function that will execute
-            the equivalent numpy function.
+        ONNXAndNumpyForwards: The function that will execute the equivalent numpy function.
     """
 
-    equivalent_onnx_model = preprocess_onnx_model(onnx_model, check_model)
+    onnx_preprocessing, equivalent_onnx_model = preprocess_onnx_model(onnx_model, check_model)
```
RomanBredehoft, line 276: I don't get why we need both onnx models here and in the following

src/concrete/ml/torch/compile.py
RomanBredehoft: also, why do you need to pass integers now (that's what you are trying to do right ?)

src/concrete/ml/quantization/post_training.py
```diff
@@ -710,6 +714,11 @@ def quantize_module(self, *calibration_data: numpy.ndarray) -> QuantizedModule:
             onnx_model=self.numpy_model.onnx_model,
         )
 
+        # Set the preprocessing onnx in the quantized module
+        quantized_module._onnx_preprocessing = (  # pylint: disable=protected-access
```
RomanBredehoft, line 718: what is going on here exactly ?

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -90,6 +92,8 @@ class QuantizedModule:
     input_quantizers: List[UniformQuantizer]
     output_quantizers: List[UniformQuantizer]
     fhe_circuit: Union[None, Circuit]
+    _onnx_preprocessing: Optional[onnx.ModelProto] = None
```
RomanBredehoft, line 95: these attributes should not be class attributes but class instances, so better to initialize them in the __init__ 

also, I would add a property setter for both instead of having to access these protected attributes in tests and so on

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -271,6 +288,24 @@ def post_processing(
...
+
+        Args:
+            values (numpy.ndarray): The input values to pre-process.
+
+        Returns:
+            Union[numpy.ndarray, Tuple[numpy.ndarray, ...]]: The pre-processed values.
+        """
+        if self._onnx_preprocessing is not None:
+            if self._preprocessing_numpy is None:
+                self._preprocessing_numpy = NumpyModule(self._onnx_preprocessing)
```
RomanBredehoft, line 305: shouldn't we just load consider a `_preprocessing_numpy` attribute only, and always build it using a onnx_processing object whenever we need (ie, directly in the dump/load methods, and/or through a public method for eg tests)

src/concrete/ml/torch/compile.py
jfrery: Because embedding layer want integers / labels.

src/concrete/ml/onnx/convert.py
jfrery: Yes so I went for the solution where now we can have some part of the model done as pre-processing. The fact that we use an onnx makes it super easy to serialize and re-use on the client machine.

The other option is do the modification to the onnx in the `post_training.py` but since we manipulate onnx in convert.py I thought it was better do to it before.

src/concrete/ml/onnx/convert.py
jfrery: Because we need the onnx_preprocessing for the pre-processing later. On the client machine and for any kind of forward of the quantized_module

src/concrete/ml/quantization/post_training.py
jfrery: Once the quantized module is created, we need to pass the pre-processing from the numpy_module to the quantized_module. You want a setter method?

src/concrete/ml/quantization/quantized_module.py
jfrery: Yes but that's true for all of them I suppose. We can do some refacto to clean this.

Agreed on the setter!

src/concrete/ml/quantization/quantized_module.py
jfrery: Isn't it what we do here? If _preprocessing_numpy is None we load it it. But if the onnx_preprocessing is None we shouldn't load anything.

src/concrete/ml/onnx/convert.py
RomanBredehoft: yes I think I get that part, but my point was more about that it looks like we are *always* converting any ONNX's first gather operator to matmul, since it's included in `preprocess_onnx_model` which is called everytime we convert to onnx, right ? I don't think that's intended !

src/concrete/ml/onnx/convert.py
RomanBredehoft: yeah make sense I guess

src/concrete/ml/quantization/post_training.py
RomanBredehoft: imo it would be better to pass it to the QM's init then, especially since we already do ` onnx_model=self.numpy_model.onnx_model` 

side note: would have been much clearer to instead have a `from_numpy_module` method for QMs, or just simply only give a `numpy_module` + `quant_layers_dict` in the init, but that's not urgent

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: what I meant is that when doing this we use 2 new attributes, but in reality only one matters (see the dumps/loads), so my guess is that we can avoid make things more complicated by only considering a single attribute instead of 2

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: see above (https://github.com/zama-ai/concrete-ml/pull/778/files/51d0d124b199c50bb81154dfd11811e352527ede#r1670354637): I think it makes sense to include `_onnx_preprocessing` in the init 

see below (https://github.com/zama-ai/concrete-ml/pull/778/files/51d0d124b199c50bb81154dfd11811e352527ede#r1670359773) : I think one of these 2 attributes could be removed for simplicity

src/concrete/ml/torch/compile.py
RomanBredehoft: ok I guess it makes sense, well as said above I would at least add a comment on that + the float thing then !

src/concrete/ml/onnx/convert.py
jfrery: Ah sorry. Yes we always use that optimization. Also currently gather isn't supported with encrypted indices.

RomanBredehoft: ah ok, and is it somewhat related to https://github.com/zama-ai/concrete-ml-internal/issues/4413 then ?

jfrery: Yes good point! And also related to https://github.com/zama-ai/concrete-ml-internal/issues/4532 where we will have to make Gather nodes compatible in FHE one way or another.

RomanBredehoft: worth adding one or 2 fixme here then !

jfrery: Done thanks

src/concrete/ml/torch/compile.py
jfrery: Added a comment

src/concrete/ml/quantization/quantized_module.py
jfrery: Ok for the first point but I not sure about the second. If you think we shouldn't return onnx + numpy forward for the preprocessing then we should do the same for the equivalent numpy forward, no?

src/concrete/ml/quantization/quantized_module.py
jfrery: Ok I just realized we are in the QuantizedModule and no in the NumpyModule. You are right let me do this.

jfrery: Actually I am not sure I get it. We shouldn't rebuild it everytime we need it that's very inefficient. So we need to check if there is a onnx. If there is one we can either build the numpy preprocessing or use it if it's already built.

jfrery: I have updated let me know what you think

src/concrete/ml/common/serialization/decoder.py
```diff
@@ -202,6 +203,9 @@ def object_hook(d: Any) -> Any:
                 model_class.__name__: model_class for model_class in serializable_classes
             }
 
+        if type_name == "Exactness":
```
RomanBredehoft, line 206: why do we store exactness objects in one of our object ?

src/concrete/ml/onnx/convert.py
RomanBredehoft: thanks

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -170,6 +178,7 @@ def dump_dict(self) -> Dict:
         metadata["ordered_module_input_names"] = self.ordered_module_input_names
         metadata["ordered_module_output_names"] = self.ordered_module_output_names
         metadata["quant_layers_dict"] = self.quant_layers_dict
+        metadata["_onnx_preprocessing_str"] = self._onnx_preprocessing
```
RomanBredehoft, line 181: better if we keep the same name between the metadata dict and the class

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -90,13 +92,16 @@ class QuantizedModule:
     input_quantizers: List[UniformQuantizer]
     output_quantizers: List[UniformQuantizer]
     fhe_circuit: Union[None, Circuit]
+    _onnx_preprocessing: Optional[onnx.ModelProto]
```
RomanBredehoft, line 95: here here are defined attributes that are not init input parameters, so better to remove this `_onnx_preprocessing: Optional[onnx.ModelProto]` but add typing to `self._onnx_preprocessing = onnx_preprocessing` below

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -142,6 +147,9 @@ def __init__(
         # Input-output quantizer mapping for composition is not enabled at initialization
         self._composition_mapping: Optional[Dict] = None
 
+        # Initialize _onnx_preprocessing
```
RomanBredehoft, line 150: would be great to add a small sentence saying what this is for

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -271,6 +282,26 @@ def post_processing(
...
+        For quantized modules, this method applies pre-processing if a NumpyModule exists,
+        or creates one if ONNX preprocessing is defined but the NumpyModule hasn't been created yet.
+
+        Args:
+            values (numpy.ndarray): The input values to pre-process.
+
+        Returns:
+            Tuple[numpy.ndarray, ...]: The pre-processed values.
+        """
+        if self._preprocessing_numpy is None and self._onnx_preprocessing is not None:
```
RomanBredehoft, line 297: so basically, are we expecting someone to run `pre_processing` multiple times in a row ? : 
- if not, then there's no need to have a `self._preprocessing_numpy` attibute anymore, we simple build and return the numpy module directly)
- if yes, then : 
    - we could keep building this numpy on the fly if it's quick 
    - if it's too slow, then why not instead directly build the numpy module in the QM's init and only handle `_preprocessing_numpy` (no need to keep `_onnx_preprocessing`) ? and for serialization you could dump the onnx processing (you could store `model` in `NumpyModule` if needed) and load back the numpy module in load.

it's just that we probably want to avoid adding more attributes to important classes like QMs if it's avoidable !

src/concrete/ml/torch/compile.py
RomanBredehoft: thanks ! just, unless you re not sure about it, could you instead write why it fails with fp64 and why we need integer support (more than just say that it fails without it) ? it's just that I feel we'll stumble again on that code one day and we'll not be sure anymore why it's needed

src/concrete/ml/torch/numpy_module.py
```diff
@@ -43,7 +43,12 @@ def __init__(
                 dummy_input is not None
             ), "dummy_input must be provided if model is a torch.nn.Module"
 
-            self.numpy_forward, self._onnx_model = get_equivalent_numpy_forward_from_torch(
+            (
+                self.numpy_preprocessing,
```
RomanBredehoft, line 47: I think `numpy_preprocessing` should be private + have a getter as well ! 

also, as stated in https://github.com/zama-ai/concrete-ml/pull/778/files#r1670793432, I think we should provide better naming here in order to make sure we do not confuse these with the QM's attribute(s)

src/concrete/ml/common/serialization/decoder.py
jfrery: I don't know. I wonder if we test serialisation on our qmodules. I just dumped and loaded and got an error that exactness could not be serialised so I fixed it.

RomanBredehoft: yes we do test serialization for QMs ! But I don't get where we would store an `Exactness` instance in QMs ! 🤔  Doesn't look right

src/concrete/ml/quantization/quantized_module.py
jfrery: Yes done.

src/concrete/ml/quantization/quantized_module.py
jfrery: What about all the over attributes above? Some of them appear in the init. But yes I can move it.

src/concrete/ml/quantization/quantized_module.py
jfrery: Well everytime a call to forward is done the pre-processing must be applied yes. We build a numpy module from onnx I don't think it's good practice to build it everytime.

Agreed to have it built in init and load time. Also agreed to remove the _onnx_preprocessing attribute and only use the _preprocessing_numpy.

jfrery: > side note : here QM's _preprocessing_numpy is a NumpyModule instance, but a NumpyModule instance also has a numpy_preprocessing attribute that is a callable. I feel like we could provide better names in order to avoid confusion !

I updated to `_preprocessing_module`. I see what you mean. Also it could be bad that a onnx_preprocessing gets its own numpy_preprocessing... Not sure how to handle that.

src/concrete/ml/torch/numpy_module.py
jfrery: why would it be more private than numpy_forward?

jfrery: I updated with your suggestion

src/concrete/ml/common/serialization/decoder.py
jfrery: I doubt we have QM tests or this would have been spotted.

jfrery: Alright so there was a probem with our QM serialisation tests indeed. We never test the case where we have rounding in the QM for serialisation. Fixed in https://github.com/zama-ai/concrete-ml/pull/778/commits/1cccb7c5307537501bd16a56f652a825c0ccf1e8

src/concrete/ml/torch/compile.py
jfrery: The error was a mismatch of between to matrix for the matmul. I didn't investigate as it was going down the onnx library before failing.

src/concrete/ml/quantization/post_training.py
jfrery: Done.

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -90,13 +92,15 @@ class QuantizedModule:
     input_quantizers: List[UniformQuantizer]
     output_quantizers: List[UniformQuantizer]
     fhe_circuit: Union[None, Circuit]
+    _preprocessing_module: Optional[NumpyModule] = None
```
RomanBredehoft, line 95: to keep consistency we should remove this + add the `Optional[NumpyModule]` typing to when we init the attribute

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -142,6 +146,14 @@ def __init__(
         # Input-output quantizer mapping for composition is not enabled at initialization
         self._composition_mapping: Optional[Dict] = None
 
+        # Initialize _preprocessing_numpy
+        # The onnx graph is used to pre-process the inputs before FHE execution
+        self._preprocessing_module = NumpyModule(onnx_preprocessing) if onnx_preprocessing else None
+
+        # Ensure that there is no preprocessing step
+        if self._preprocessing_module is not None:
```
RomanBredehoft, line 154: is this always true ? why ?

tests/quantization/test_quantized_module.py
```diff
@@ -469,27 +469,44 @@ def quantized_module_predictions_are_equal(
...
     """Test the serialization of quantized modules."""
 
     # Define the torch model
     torch_fc_model = model_class(activation_function=nn.ReLU)
     torch_fc_model.eval()
 
     # Create a random input
-    numpy_input = numpy.random.uniform(size=input_shape)
-    torch_input = torch.from_numpy(numpy_input).float()
+    if isinstance(model_class, partial) and model_class.func == EmbeddingModel:
```
RomanBredehoft, line 493: I don't think you need to add this test here , imo you test anything new ! This tests is about serializing QMs, with a few torch models (in order to get a relevant initialized QM). I don't see what `EmbeddingModel` brings more to the test, you already (rightly) test its compilation in `test_compile_torch` anyway ! imo it's safe to remove that part

if you were thinking about the QM rounding serialization issue, imo the `rounding_threshold_bits` fix you've added below is enough

jfrery: EmbeddingModel is the only model that will create a onnx preprocessing which we need to make sure can be serialised.

src/concrete/ml/quantization/quantized_module.py
jfrery: I have lost track here. We create a `QuantizedModule` object and we CAN pass or not an `onnx_preprocessing`. Its typing is

`onnx_preprocessing: Optional[onnx.ModelProto] = None,`

so yes it can be None. The line above does

```python
        self._preprocessing_module = NumpyModule(onnx_preprocessing) if onnx_preprocessing else 
```

src/concrete/ml/quantization/quantized_module.py
jfrery: I am not sure I follow the rules for attributes why would preprocessing_module not belong here vs the above attributes?

src/concrete/ml/onnx/onnx_model_manipulations.py
```diff
@@ -289,3 +289,103 @@ def _clean_graph_at_node_name(
 
     # Keep the output node
     keep_following_outputs_discard_others(onnx_model, [output_to_follow])
+
+
+# Function to convert the first Gather nodes
+# to matrix multiplications with one-hot encoding as a pre-processing step
```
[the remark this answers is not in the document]
jfrery: Done

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: ok sorry I think last time I must have read `if self._preprocessing_module.onnx_preprocessing is None` instead of `if self._preprocessing_module is not None:`, which made less much sense 😅  Thanks for explaining it ! looks good

tests/quantization/test_quantized_module.py
RomanBredehoft: hmm, sure ok, I first didn't find this relevant because it's the same kind of object than onnx_model, but you're right we never now it that changes later so better to be safe, sorry for the noise !

src/concrete/ml/quantization/quantized_module.py
RomanBredehoft: actually these attribute typings are not necessary and should be removed as they all are defined in the init (some are missing some typings but we should add the)

normally these class attribute typings are defined here when not done in the init (in case they are instantiated in subclasses' init for example, like done in built-in models), so it's an error to have them here

as for `_preprocessing_module` specifically, it does not make sense to put a default value if not relevant. And imo it is not relevant here because we instantiate it in the init anyway. So it would be preferable to remove it entirely. And I'll open a follow-up PR to remove the other ones !

src/concrete/ml/quantization/quantized_module.py
```diff
@@ -142,6 +146,14 @@ def __init__(
         # Input-output quantizer mapping for composition is not enabled at initialization
         self._composition_mapping: Optional[Dict] = None
 
+        # Initialize _preprocessing_numpy
```
RomanBredehoft, line 149: small bad refactor artifact : it should be `_preprocessing_module`

RomanBredehoft (changes requested):
a few last changes !

src/concrete/ml/quantization/quantized_module.py
jfrery: thanks!

RomanBredehoft (approved):

RomanBredehoft (approved):

RomanBredehoft (approved):

PR #806, opened 2024-07-23
chore: fix readme module names example for phi-1.5

jfrery:
layer names have been updated on HF


RomanBredehoft (approved):
now that's a long name

RomanBredehoft:
ah yes didn't see we are importing three layers now 😅

PR #807, opened 2024-07-24
feat: add brevitas channel-wise support

fd0r:


src/concrete/ml/onnx/convert.py
```diff
@@ -142,9 +146,14 @@ def get_equivalent_numpy_forward_from_torch(
...
+    else:
+        dummy_input = tuple(elt.to("cpu") for elt in dummy_input)
+
     # Export to ONNX
     torch.onnx.export(
-        torch_module,
+        torch_module.to("cpu"),
         dummy_input,
         str(output_onnx_file_path),
         opset_version=OPSET_VERSION_FOR_ONNX_EXPORT,
```
fd0r, line 159: We need the module and the inputs to be on CPU for the exporter to work properly

src/concrete/ml/pytest/torch_models.py
```diff
@@ -1600,3 +1609,176 @@ def forward(self, x):  # pylint: disable-next=no-self-use
             Tuple[torch.Tensor. torch.Tensor]: Outputs of the network.
         """
         return x, x.unsqueeze(0)
+
+
+# pylint: disable-next=too-many-ancestors
+class CommonIntWeightPerChannelQuant(Int8WeightPerTensorFloat):
+    """CommonIntWeightPerChannelQuant."""
+
+    scaling_per_output_channel = True
```
fd0r, line 1618: The per-channel quantizer from Brevitas

src/concrete/ml/pytest/torch_models.py
```diff
@@ -1600,3 +1609,176 @@ def forward(self, x):  # pylint: disable-next=no-self-use
...
+            bias=True,
+            input_bit_width=act_bit_width,
+            input_quant=CommonUintActQuant,
+            weight_accumulator_bit_width=acc_bit_width,
+            weight_bit_width=weight_bit_width,
+            weight_restrict_scaling_type=RestrictValueType.LOG_FP,
+            weight_quant=weight_quant,
+        )
+
+        self.apply(weight_init)
```
fd0r, line 1784: A LeNet provided by a user

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -420,7 +421,14 @@ def copy_function(x):
         # Note that here we do not rescale to the output_scale and we do not add a zero-point
         # Any following Gemm/MatMul/Conv layers will do the rescaling (during re-quantization)
         # by calling _prepare_inputs_with_constants(...quantize_real_values=True)
-        m_matmul = q_input1.quantizer.scale * q_input2.quantizer.scale
+        m_matmul: npt.NDArray[numpy.float64]
+        if q_input2.quantizer.scale.shape == tuple():
+            m_matmul = q_input1.quantizer.scale * q_input2.quantizer.scale
+        else:
+            # TODO: add assert on shapes
```
jfrery, line 428: TODO to remove or convert in fixme

jfrery (changes requested):
Thanks a lot for tackling this.

My main comment is: How do we make sure per channel quantization is indeed activated and used in concrete-ml?

Currently we only test that a per channel brevitas quantization can be compiled in CML but are we sure we use all the scales properly?

I am thinking, maybe have a simple model with a single conv and make sure that the number of scales == number of channel?

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -913,7 +915,7 @@ def can_fuse(self) -> bool:
     def make_output_quant_parameters(
         self,
         q_values: Union[numpy.ndarray, Any],
-        scale: numpy.float64,
+        scale: npt.NDArray[numpy.float64],
```
RomanBredehoft, line 918: is `npt` a common alias ? why not just put `numpy.ndarray` like usual ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -420,7 +421,14 @@ def copy_function(x):
         # Note that here we do not rescale to the output_scale and we do not add a zero-point
         # Any following Gemm/MatMul/Conv layers will do the rescaling (during re-quantization)
         # by calling _prepare_inputs_with_constants(...quantize_real_values=True)
-        m_matmul = q_input1.quantizer.scale * q_input2.quantizer.scale
+        m_matmul: npt.NDArray[numpy.float64]
+        if q_input2.quantizer.scale.shape == tuple():
```
RomanBredehoft, line 425: why this and not `isinstance(q_input2.quantizer.scale, numpy.float64)` ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -420,7 +421,14 @@ def copy_function(x):
...
+            m_matmul = q_input1.quantizer.scale * q_input2.quantizer.scale
+        else:
+            # TODO: add assert on shapes
+            weight_quant_scale = numpy.transpose(q_input2.quantizer.scale, axes=(1, 0))
```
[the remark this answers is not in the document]
RomanBredehoft: agree, might be worth a comment if that's the case

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -953,7 +961,17 @@ def q_impl(
         # This is going to be compiled with a PBS (along with the following activation function)
         # Note that we don't re-quantize the output of the conv, this will be done by
         # any Gemm/Add/Conv layers that follow
-        m_matmul = q_input.quantizer.scale * q_weights.quantizer.scale
+        m_matmul: npt.NDArray[numpy.float64]
+        if q_weights.quantizer.scale.shape == tuple():
```
RomanBredehoft, line 965: same

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -953,7 +961,17 @@ def q_impl(
...
+            m_matmul = q_input.quantizer.scale * q_weights.quantizer.scale
+        else:
+            # TODO: add assert on shapes
+            weight_quant_scale = numpy.transpose(
+                q_weights.quantizer.scale,
+                axes=(1, 0, 2, 3),
```
RomanBredehoft, line 971: same

tests/torch/test_brevitas_qat.py
```diff
@@ -606,3 +611,20 @@ def test_brevitas_power_of_two(
...
+    with torch.no_grad():
+        batch_size = 3
+        image_size = 1, 32, 32
+        images = torch.rand((batch_size, *image_size))
+        out = model(images).detach().numpy()
+        quantized_module = compile_brevitas_qat_model(model, images, rounding_threshold_bits=6)
+        out_qm = quantized_module(images.detach().numpy())
+        mse = ((out - out_qm) ** 2).mean()
+        # Arbitrary threshold to check that the predictions are relatively similar
+        assert mse < 1e-4
```
RomanBredehoft, line 630: probably better to use our tools like `check_float_array_equal`, which can provide rtol or atol, or `check_r2_score` or similar 

if not, then probably better to make this mse check a fixture so that we can re-use it elsewhere and not forget about this arbitrary value here

src/concrete/ml/quantization/base_quantized_op.py
fd0r: I was `npt` in some document somewhere and that help keep the code succint.

There is no way as far as I'm aware to specify the dtype of the np array using only `np.ndarray` and thought it would be better to specify it here.

src/concrete/ml/quantization/quantized_ops.py
fd0r: same answer, the assert above (to be added) will catch any issue but it should always have this shape for conv2d.

I'll add some errors in the case of conv1d or conv3d

tests/torch/test_brevitas_qat.py
fd0r: yeah true `check_float_array_equal` would probably be better indeed. I'll change that.

src/concrete/ml/quantization/quantized_ops.py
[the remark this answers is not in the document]
RomanBredehoft: which assert ? 🤔

jfrery (dismissed):
Removing my request as I will be off for some time. Good luck with the PR!

src/concrete/ml/quantization/base_quantized_op.py
RomanBredehoft: got it thanks !

PR #832, opened 2024-08-06
Relax python version restrictions

jdmarsh:
I have a client/server setup where we use the via_mlir flag to save/load the circuit. When the server.zip is generated with a different python version to what the server is running, this python version restriction prevents us from continuing. As long as the concrete/concrete-ml versions are identical, I don't see why the python version should restrict us from loading the MLIR in server side (Or why it should apply as a restriction even when not using via_mlir?). I've made a change to relax the python version restriction server side for MLIR cases.

Please let me know if there is a good reason for this restriction to apply when sending an MLIR server.zip generated with a different python version. In that case I will find another way forward.

### DIFF
--- a/src/concrete/ml/deployment/fhe_client_server.py
+++ b/src/concrete/ml/deployment/fhe_client_server.py
@@ -64,6 +64,7 @@ def check_concrete_versions(zip_path: Path):
     """
 
     with zipfile.ZipFile(zip_path) as zip_file:
+        is_mlir = "circuit.mlir" in zip_file.namelist()
         with zip_file.open("versions.json", mode="r") as file:
             versions = json.load(file)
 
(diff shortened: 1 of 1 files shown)


bcm-at-zama:
Thanks, we'll have a look

andrei-stoian-zama (commented):
I think this looks like a good approach. @jfrery , @umut-sahin what do you think ?

umut-sahin (commented):
Looks good IMO.

andrei-stoian-zama:
I re-created this PR with my account to let the CI run: https://github.com/zama-ai/concrete-ml/pull/853

Closing this one

andrei-stoian-zama:
Big thanks to @jdmarsh for this contribution !

PR #837, opened 2024-08-13
chore: fix wheel steps in weekly tests

RomanBredehoft:
not sure why this issue (https://github.com/zama-ai/concrete-ml/actions/runs/10343073361/job/28626854301) didn't happen last time I tried locally, now it should really fix the weekly ci


kcelia (approved):
This piece of code looks familiar, it seems to me that it was deleted at some point

PR #838, opened 2024-08-13
chore: change save_and_clear_private_info to consider brevitas models

kcelia:
Close https://github.com/zama-ai/concrete-ml-internal/issues/4571

### DIFF
--- a/tests/torch/test_hybrid_converter.py
+++ b/tests/torch/test_hybrid_converter.py
@@ -214,6 +214,7 @@ def test_hybrid_brevitas_qat_model():
         # Check that files are there
         assert (temp_dir_path / "model.pth").exists()
 
+
 # Dependency 'huggingface-hub' raises a 'FutureWarning' from version 0.23.0 when calling the
 # 'from_pretrained' method
 @pytest.mark.filterwarnings("ignore::FutureWarning")
--- a/tests/torch/test_hybrid_converter.py
+++ b/tests/torch/test_hybrid_converter.py
@@ -205,6 +205,14 @@ def test_hybrid_brevitas_qat_model():
     hybrid_model = HybridFHEModel(model, module_names="sub_module")
     hybrid_model.compile_model(x=inputs)
 
+    with tempfile.TemporaryDirectory() as temp_dir:
+        temp_dir_path = Path(temp_dir)
+
+        # Get the temp directory path
+        hybrid_model.save_and_clear_private_info(temp_dir_path)
+
+        # Check that files are there
+        assert (temp_dir_path / "model.pth").exists()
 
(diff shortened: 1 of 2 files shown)


RomanBredehoft (changes requested):
thanks ! small typo in the commit though

tests/torch/test_hybrid_converter.py
```diff
@@ -205,6 +205,15 @@ def test_hybrid_brevitas_qat_model():
...
     hybrid_model.compile_model(x=inputs)
 
+    with tempfile.TemporaryDirectory() as temp_dir:
+        temp_dir_path = Path(temp_dir)
+
+        # Get the temp directory path
+        hybrid_model.save_and_clear_private_info(temp_dir_path)
+
+        # Check that files are there
+        assert (temp_dir_path / "model.pth").exists()
```
bcm-at-zama, line 215: is it sufficient? shouldn't we check it works / it's usable etc? like, loading and using after having saved

kcelia: It's the same check as the standard pytorch model.

RomanBredehoft (approved):

tests/torch/test_hybrid_converter.py
kcelia: As Roman mentioned, this method isn't very useful since we don't specify the remote layers and there's no load function.

It would be better to include this in a separate PR.

bcm-at-zama: - but if the method is not useful, why do we have it? and if there is no load, I am not sure I understand the idea here. 
- and "It's the same check as the standard pytorch model.": so do we test what happens in `save_and_clear_private_info`? if I remove the body of save_and_clear_private_info, ie change its behavior, will a test catch it? it's what we would expect

RomanBredehoft: yes I agree that removing it makes more sense for now (else, we should provide a way to load hybrid models but not even sure how relevant this would be)

kcelia: Roman suggested to delete it, but I believe that we can enhance this functionality.
I haven't worked much on the hybrid approach, so I prefer to wait for the others to weigh in. That way, we can decide together whether to enhance it or remove it.

kcelia: No, we don't check what's going inside the body, as we don't have a load function.

kcelia: I don't see the importance of this function or if it truly adds any value.

Let’s wait for the others to decide what to do with it.

bcm-at-zama: ok let's wait

PR #842, opened 2024-08-20
test: concrete simulation p error is working

rudy-6-4:


kcelia (approved):

kcelia:
Please ignore my earlier approval; I hadn't noticed that the tests are not yet complete.

kcelia:
make conformance issues; I'll fix it

kcelia:
In local, make pcc is clean. But here it shows: 
```tests/quantization/test_quantized_ops.py:592: error: Argument "q_input" to "quantized_op_results_are_equal" has incompatible type "Tuple[Any, Any]"; expected "Optional[ndarray[Any, Any]]"  [arg-type]
``` 
any idea @RomanBredehoft  ?

tests/quantization/test_quantized_ops.py
```diff
@@ -100,7 +100,7 @@
 def quantized_op_results_are_equal(
     quantized_op_1: QuantizedOp,
     quantized_op_2: QuantizedOp,
-    q_input: Optional[numpy.ndarray] = None,
+    q_input: Optional[numpy.ndarray] = None,  # type: ignore[arg-type]
```
bcm-at-zama, line 103: why did it change

bcm-at-zama: if it's because of the xfail, could you create an issue to remove that and add a FIXME please?
maybe not @rudy-6-4 but someone from ML team

bcm-at-zama (commented):
- I wonder if a `poetry lock --no-update` was used. The --no-update is important, and as there are much too many changes, I guess it's not what was done
- also it's not expected that only Linux licenses have changed

So I don't think it can be merged as is.

bcm-at-zama:
@rudy-6-4 : if it's too complex to make this build green, no worries, the ML team can continue it, when they have some time. It's not urgent

tests/quantization/test_quantized_ops.py
kcelia: Because I have a mypy warning with this line, only on the CI. 
Locally, make pcc, is fine

kcelia: I did make force_licenses. 
I haven't update poetry.

Do you suggest to do first petry update --no-update then make licenses ?

bcm-at-zama: - changing the CP version shouldn't change mypy here, in my understanding
- and yes, it's not for this thread, but the poetry lock has to be done with --no-update

bcm-at-zama: we don't launch `make force_licenses.` anymore, there is an action for that now
if you don't know, maybe we can wait the rest of the team

RomanBredehoft:
yes @bcm-at-zama is right, you should try to check out why we have this error,  I don't think keeping this ignore is great .. also, what is this `mark.xfail` thing ? 

and yes to update licenses you just need to target the branch with https://github.com/zama-ai/concrete-ml/actions/workflows/update_licenses.yaml

rudy-6-4:
> yes @bcm-at-zama is right, you should try to check out why we have this error, I don't think keeping this ignore is great .. also, what is this `mark.xfail` thing ? cc @kcelia
> 
> and yes to update licenses you just need to target the branch with https://github.com/zama-ai/concrete-ml/actions/workflows/update_licenses.yaml

mark.xfail mean the test is executed and we know it fails, and it's reported if it no longer fails. until the test is fixed that's the best solution (for Concrete) so we can still use Concrete ML tests in concrete.

bcm-at-zama:
> mark.xfail mean the test is executed and we know it fails, and it's reported if it no longer fails. until the test is fixed that's the best solution (for Concrete) so we can still use Concrete ML tests in concrete.

yes yes, but what we say is that "we should add ignore's that we'll forget to remove, after the test is back", we need to keep track of it in an issue

tests/torch/test_compile_torch.py
```diff
@@ -675,6 +675,7 @@ def test_compile_brevitas_qat(
     )
 
 
+@pytest.mark.xfail
```
bcm-at-zama, line 678: we need an issue to fix that please

tests/quantization/test_quantized_ops.py
kcelia: I followed: poetry lock no update + make licenses
mypy no longer displays errors

bcm-at-zama: yes, was certainly due to an update in mypy, which is no more updated with --no-update

tests/torch/test_compile_torch.py
rudy-6-4: the issue already exists: https://github.com/zama-ai/concrete-ml-internal/issues/4576

bcm-at-zama: yes but the rule in CML is to add a FIXME on the source code to not forget
it's more for @kcelia than for you @rudy-6-4 here, don't worry

tests/torch/test_compile_torch.py
```diff
@@ -675,6 +675,9 @@ def test_compile_brevitas_qat(
     )
 
 
+# Due to a Concrete issue, this test is ignored
```
bcm-at-zama, line 678: strictly speaking, it's not a Concrete issue, it's: Concrete had a bug in simulation which has been fixed, and now, we need to update our test to be compatible with the fixed CP

tests/torch/test_compile_torch.py
```diff
@@ -675,6 +675,9 @@ def test_compile_brevitas_qat(
     )
 
 
+# Update this test to align with Concrete's simulation fix.
```
bcm-at-zama, line 678: yes thx!

PR #843, opened 2024-08-22
chore: Fix some elements in the .md that could cause confusion (repor…

kcelia:
- Fix some elements in the .md that could cause confusion (reported by a user).
    link: https://discord.com/channels/901152454077452399/1083021638544670720/1273533424187084800
-   Add a check to ensure the model is a PyTorch model, and a simple test.

### DIFF
--- a/docs/guides/hybrid-models.md
+++ b/docs/guides/hybrid-models.md
@@ -53,11 +53,12 @@ class FCSmall(nn.Module):
     def forward(self, x):
         return self.seq(x)
 
-model = FCSmall(10)
+dim = 10
+model = FCSmall(dim)
 model_name = "FCSmall"
 submodule_name = "seq.0"
 
-inputs = torch.Tensor(np.random.uniform(size=(10, 10)))
+inputs = torch.Tensor(np.random.uniform(size=(10, dim)))
 # Prints ['', 'seq', 'seq.0', 'seq.1', 'seq.2']
 print([k for (k, _) in model.named_modules()])
 
@@ -67,7 +68,6 @@ hybrid_model.compile_model(
     inputs,
     n_bits=8,
 )
-
 
 models_dir = Path(os.path.abspath('')) / "compiled_models"
 models_dir.mkdir(exist_ok=True)
@@ -110,7 +110,7 @@ You can develop a client application that deploys a model with hybrid deployment
 ```python
 # Modify model to use remote FHE server instead of local weights
 hybrid_model = HybridFHEModel(
-    model,
+    model,  # PyTorch or Brevitas model
     submodule_name,
     server_remote_address="http://0.0.0.0:8000",
     model_name=f"{model_name}",
@@ -122,7 +122,7 @@ Next, obtain the parameters necessary to encrypt and quantize data, as detailed 
 
 <!--pytest-codeblocks:skip-->
 
-```
+```python
 path_to_clients = Path(__file__).parent / "clients"
 hybrid_model.init_client(path_to_clients=path_to_clients)
 ```
@@ -131,7 +131,7 @@ When the client application is ready to make inference requests to the server, s
 
 <!--pytest-codeblocks:skip-->
 
-```
+```python
 for module in hybrid_model.remote_modules.values():
     module.fhe_local_mode = HybridFHEMode.REMOTE    
 ```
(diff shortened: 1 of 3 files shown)


docs/guides/hybrid-models.md
```diff
@@ -141,7 +141,7 @@ For inference with the `HybridFHEModel` instance, `hybrid_model`, call the regul
 <!--pytest-codeblocks:skip-->
 
 ```python
-hybrid_model.forward(torch.randn((dim, )))
+hybrid_model(torch.randn((dim, )))
```
bcm-at-zama, line 144: here, do we call the quantized module or really the simulate/fhe function?

docs/guides/hybrid-models.md
```diff
@@ -141,7 +141,7 @@ For inference with the `HybridFHEModel` instance, `hybrid_model`, call the regul
...
 ```
 
 When calling `forward`, the `HybridFHEModel` handles all the necessary intermediate steps for each model part deployed remotely, including:
```
bcm-at-zama, line 147: if the previous line is changed by removing the `forward`, then you would have to edit this line as well, which starts by

> When calling `forward`

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -343,21 +343,27 @@ class HybridFHEModel:
...
         self,
         model: nn.Module,
         module_names: Union[str, List[str]],
-        server_remote_address=None,
+        server_remote_address: Optional[str] = None,
         model_name: str = "model",
         verbose: int = 0,
     ):
+        if not isinstance(model, torch.nn.Module):
+            raise TypeError("The model must be a PyTorch model.")
```
bcm-at-zama, line 365: in the doc, you had commented

>    model,  # PyTorch or Brevitas model

so:
- should the error string contain "Brevitas" as well
- is the `if not isinstance(model, torch.nn.Module):` valid for Brevitas case as well

tests/torch/test_hybrid_converter.py
```diff
@@ -229,3 +227,14 @@ def test_gpt2_hybrid_mlp_module_not_found():
     fake_module_name = "does_not_exist"
     with pytest.raises(ValueError, match=f"No module found for name {fake_module_name}"):
         HybridFHEModel(model, module_names=fake_module_name)
+
+
+def test_invalid_model():
```
bcm-at-zama, line 232: thx for this

tests/torch/test_hybrid_converter.py
```diff
@@ -229,3 +227,14 @@ def test_gpt2_hybrid_mlp_module_not_found():
     fake_module_name = "does_not_exist"
     with pytest.raises(ValueError, match=f"No module found for name {fake_module_name}"):
         HybridFHEModel(model, module_names=fake_module_name)
+
```
bcm-at-zama, line 230: do we somewhere have tests with brevitas models, since we say it works for both torch and brevitas?

bcm-at-zama (commented):
Some comments, but take them with a grain of salt, it's been a long time I haven't looked at that precisely

src/concrete/ml/torch/hybrid_model.py
kcelia: > is the if not isinstance(model, torch.nn.Module): valid for Brevitas case as well

Yes, because, our Brevitas models inherit from torch.nn.Module

kcelia: updated the msg : "The model must be a PyTorch model." to "The model must be a PyTorch or Brevitas model."

tests/torch/test_hybrid_converter.py
kcelia: We have : test_hybrid_brevitas_qat_model

docs/guides/hybrid-models.md
kcelia: We call the quantized private modules, the modes are: simulate, disable, execute and calibrate.

 or really the simulate/fhe function?

bcm-at-zama: I thought that the simulate/fhe functions were used when we called `run`, which was below `forward` in the non-hybrid models, and that calling simply without the `run` was using the "all-in-CML (ie, no CP, no simulation nor FHE" quantized module function. But I may be wrong, to be checked with the rest of ML team maybe

kcelia: TBH, I don't know

jfrery: > here, do we call the quantized module or really the simulate/fhe function?

By default, it's the quantized module. I think it's better to add the forward in the hybrid model as said in this issue https://github.com/zama-ai/concrete-ml-internal/issues/4579.

andrei-stoian-zama (approved):

PR #849, opened 2024-08-29
Add gpu support

andrei-stoian-zama:
### Plan

1. Merge this PR
2. Re-run notebooks on main with GPU
3. Add GPU tests in the CI, when CP is updated

### New API

API is:

```
circuit = model.compile(inputset, device=device)
# or
q_module = compile_torch_model(net, x_train, rounding_threshold_bits=6, p_error=0.1, device=device)
# or
q_module = compile_brevitas_qat_model(net, x_train, rounding_threshold_bits=6, p_error=0.1, device=device)
```

where 

`device` is one of `cpu`/`cuda`.

Support: 
- all built-in models: supported
- custom models `compile_torch_model` / `compile_brevitas_qat_model`: supported
- deployment: supported
- dataframe: *not* supported

### Notebooks

Selected notebooks demo GPU support and can be executed on GPU with a new CI flow (refresh notebooks on GPU)

Closes https://github.com/zama-ai/concrete-ml-internal/issues/4496


use_case_examples/cifar/cifar_brevitas_training/experiments/CNV_2W2A_2W2A_20221114_131345/log.txt
bcm-at-zama, line 1: do we really want to commit this file?

jfrery (commented):
I propose the following:

Remove all changes in notebooks so that we keep using CPU by default and user can set `device = "cuda"` in the compile call if they want to.

For our tests, I would add the logic around the `CML_USE_GPU = 1` overwriting the `device` object in CML. Maybe in `check_compilation_device_is_valid_and_is_cuda` and `check_execution_device_is_valid_and_is_cuda`.

That way, we have:
- no change to notebook
- easy launch of our notebooks in the workflow for GPU backend
- less variable to understand for users in notebooks

If we want to keep one notebook with gpu enable when it can then we can just add the

```python
device = "cuda" if check_gpu_available() else "cpu"
```

which is very torch friendly.

use_case_examples/cifar/cifar_brevitas_training/experiments/CNV_2W2A_2W2A_20221114_131345/log.txt
jfrery: yes to be removed

jfrery (approved):
Thanks!

PR #853, opened 2024-09-02
Relax python version restrictions

andrei-stoian-zama:
https://github.com/jdmarsh

> I have a client/server setup where we use the via_mlir flag to save/load the circuit. When the server.zip is generated with a different python version to what the server is running, this python version restriction prevents us from continuing. As long as the concrete/concrete-ml versions are identical, I don't see why the python version should restrict us from loading the MLIR in server side (Or why it should apply as a restriction even when not using via_mlir?). I've made a change to relax the python version restriction server side for MLIR cases.
> 
> | Please let me know if there is a good reason for this restriction to apply when sending an MLIR server.zip generated with a different python version. In that case I will find another way forward.

Closes https://github.com/zama-ai/concrete-ml-internal/issues/4280

### DIFF
--- a/src/concrete/ml/deployment/fhe_client_server.py
+++ b/src/concrete/ml/deployment/fhe_client_server.py
@@ -64,6 +64,7 @@ def check_concrete_versions(zip_path: Path):
     """
 
     with zipfile.ZipFile(zip_path) as zip_file:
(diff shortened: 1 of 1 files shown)


fd0r (approved):
LGTM

Issue #862, opened 2024-09-06
Torch.where is not correctly supported


theostos:
## Summary

I think there is an issue with the support of "torch.where" within "compile_torch_model".
Torch.where is expecting a bool tensor for the "condition" parameter, while "compile_torch_model" is expecting a float tensor (maybe related to the discrepancy between the supported type of torch.where and numpy.where for the "condition" parameter). 
It is not possible to compile a torch model using torch.where because then:
-to compute the trace, torch requires a bool tensor.
-to quantize the model, concrete ml is expecting a float tensor.

## Description

- versions affected: concrete-ml 1.6.1
- python version: 3.9
- workaround:
I was able to make it work with a (very bad) workaround:
in _process_initializer of PostTrainingAffineQuantization (concrete.ml.quantization.post_training), recast "values" variable to numpy.float if array of bool.
(unfortunately overriding "_check_distribution_is_symmetric_around_zero" is not enough..)

<details><summary>minimal POC to trigger the bug</summary>
<p>

```python
import torch
from concrete.ml.torch.compile import compile_torch_model

    
class PTQSimpleNet(torch.nn.Module):
    def __init__(self, n_hidden):
        super().__init__()
        self.n_hidden = n_hidden
        self.fc_tot = torch.rand(1, n_hidden) > 0.5

    def forward(self, x):
        y = torch.where(self.fc_tot, x, 0.)
        return y

N_FEAT = 32

torch_input = torch.randn(1, N_FEAT)
torch_model = PTQSimpleNet(N_FEAT)

quantized_module = compile_torch_model(
    torch_model,
    torch_input
)
```

</p>
</details>

jfrery:
Thanks for reporting the bug! 

Indeed we don't support booleans params very well. We have an issue to implement the proper solution but it will be much harder than your work around which I think it sustainable.

I opened a PR with the workaround so that we fix this error.

PR #863, opened 2024-09-06
chore: fix where which use boolean params

jfrery:
closes https://github.com/zama-ai/concrete-ml/issues/862

### DIFF
--- a/src/concrete/ml/pytest/torch_models.py
+++ b/src/concrete/ml/pytest/torch_models.py
@@ -1494,6 +1494,27 @@ class ManualLogisticRegressionTraining(torch.nn.Module):
         return outputs.squeeze()
 
 
+class WhereNet(torch.nn.Module):
+    """Simple network with a where operation for testing."""
+
+    def __init__(self, n_hidden):
+        super().__init__()
+        self.n_hidden = n_hidden
+        self.fc_tot = torch.rand(1, n_hidden) > 0.5
+
+    def forward(self, x):
+        """Forward pass.
+
+        Args:
+            x (torch.Tensor): The input tensor.
+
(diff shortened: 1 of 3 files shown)


src/concrete/ml/pytest/torch_models.py
```diff
@@ -1494,6 +1494,27 @@ def predict(x, weights, bias):
...
+    def forward(self, x):
+        """Forward pass.
+
+        Args:
+            x (torch.Tensor): The input tensor.
+
+        Returns:
+            torch.Tensor: The output tensor after applying the where operation.
+        """
+        y = torch.where(self.fc_tot, x, 0.0)
```
andrei-stoian-zama, line 1514: the condition here is not done on the encrypted variable. is this the case we should test ?

jfrery: I am not sure I understand. self.fc_tot is the condition, a constant, not encrypted and x is the input encrypted tensor that will be 0 where condition matches?

andrei-stoian-zama: I think we should also test the case when the condition is on an encrypted variable

jfrery: Isn't this the case? x is encrypted and self.fc_tot isn't. Do you mean "is **not** on an encrypted variable?"

jfrery: Ok sorry you are right the conditions are in clear we are just selecting a part of the encrypted tensor.

andrei-stoian-zama (approved):

kcelia (approved):

PR #868, opened 2024-09-09
feat: GLWE mat mul for hybrid model

andrei-stoian-zama:
- the PR activates GLWE when possible
- the MLP lora works (it used to work before in @RomanBredehoft ’s version, I just cleaned it up and added text)
- LORA is supported for GLWE in hybrid model
- adds a test which uses GLWE for MLP inference and comparest to QM and pytorch fp32
- optimizes compilation when the hybrid model only has linear layers
- optimizes quantized clear execution (fhe=disable) when the hybrid model only has linear layers


src/concrete/ml/torch/hybrid_model.py
```diff
@@ -109,6 +113,49 @@ def convert_conv1d_to_linear(layer_or_module):
     return layer_or_module
 
 
+def enc_matmul(q_x, q_weight, crypto_params, compression_key, private_key, serialize=True):
```
andrei-stoian-zama, line 116: not used anymore

tests/torch/test_hybrid_converter.py
```diff
@@ -52,11 +57,16 @@ def run_hybrid_llm_test(
...
     )
 
+    logits_simulate = None
+
     with monkeypatch.context() as m:
         if not transformers_installed:
             m.setitem(sys.modules, "transformers", None)
             if has_pbs_reshape:
                 has_pbs = True
+        if not glwe_backend_installed:
```
andrei-stoian-zama, line 67: this simulates environments where `concrete-ml-extensions` is not available (e.g. MAC)

tests/torch/test_hybrid_converter.py
```diff
@@ -75,6 +85,13 @@ def run_hybrid_llm_test(
             assert "NoParametersFound" in error.args[0]
             pytest.skip(error.args[0])
 
+        # Check we can run the simulate locally
+        if has_pbs or not glwe_backend_installed:
+            logits_simulate = hybrid_model(inputs, fhe="simulate").logits
```
andrei-stoian-zama, line 90: if GLWE is enabled (and this model is fully linear), it can't be simulated - we don't yet know how to simulate the GLWE stuff

tests/torch/test_hybrid_converter.py
```diff
@@ -240,3 +272,85 @@ def test_invalid_model():
...
+    y1_test = y1_test.numpy()
+    acc_fp32 = numpy.sum(numpy.argmax(y_torch, axis=1) == y1_test)
+    acc_qm = numpy.sum(numpy.argmax(y_qm, axis=1) == y1_test)
+    acc_glwe = numpy.sum(numpy.argmax(y_glwe, axis=1) == y1_test)
+
+    # These two should be exactly the same
+    assert numpy.all(numpy.allclose(y_torch, y_hybrid_torch, rtol=1, atol=0.001))
+
+    # The clear quantization vs fp32 test has more tolerance
+    threshold_fhe = 0.01
```
andrei-stoian-zama, line 340: experimentally adjusted. 

the accuracy is fine even with this threshold on the logits

src/concrete/ml/quantization/quantizers.py
```diff
@@ -62,9 +63,16 @@ def fill_from_kwargs(obj, klass, **kwargs):
     # If the structure was created or modified by a call to this function, check
     # that it is completely filled
     if obj is not None:
-        for name in hints:
-            if getattr(obj, name) is None:
-                raise TypeError(f"Missing quantizer parameter {name}")
+        all_members_missing = all(getattr(obj, name) is None for name in hints)
```
andrei-stoian-zama, line 66: I wanted the `stats` to be optional -> to be able to create a quantizer out of analytically determined values, e.g. when no min/max is computed.

But I want the caller to either be able to use all the fields in `MinMaxStatistics` or none of them.

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -700,7 +700,7 @@ def _prepare_inputs_with_constants(
 
         return prepared_inputs
 
-    def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
+    def calibrate(self, *inputs: Union[QuantizedArray, numpy.ndarray]) -> numpy.ndarray:
```
andrei-stoian-zama, line 703: Allow calibration with `QuantizedArray`. Thus the calibration can be performed analytically. 

this is similar to what Brevitas does with `QuantTensor` which determines output quant parameters based on the input ones, without calibration.

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -720,12 +722,40 @@ def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
         if isinstance(raw_result, RawOpOutput):
             return raw_result
 
-        quantized_samples = QuantizedArray(self.n_bits, raw_result)
+        supports_analytical_quant = all(
```
andrei-stoian-zama, line 725: In theory all mixing ops could be calibrated analytically. But only Gemm implements the analytical formulas, all others will throw some kind of E_NOT_IMPL

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -720,12 +722,40 @@ def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
...
+                *inputs, calibrate=False, quantize_actual_values=True
+            )
+            quantizer = self.calibrate_analytical_output(*q_prepared_inputs)
+            self.output_quant_params = quantizer.quant_params
+            self.output_quant_stats = quantizer.quant_stats
+        else:
+            # These output quantization parameters are only used
+            # for operations that produce graph output operation
+            # and are a non-linear
+            quantized_samples = QuantizedArray(self.n_bits, raw_result)
```
andrei-stoian-zama, line 739: this is the previous "data-based calibration" that was done for op outputs. these values aren't actually used except for Univariate functions (mixing ops apply their input quantization)

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -491,6 +491,68 @@ def copy_function(x):
...
+        Args:
+            *inputs (QuantizedArray): quantized operation inputs. Quantized weights
+                are stored in the op instance
+
+        Returns:
+            res (UniformQuantizer): the quantizer of the operation's output values
+                that can be used to de-quantize these values.
+        """
+
+        q_input1 = inputs[0]
```
andrei-stoian-zama, line 506: implements the famous gemm calibration formula by identifying terms

deps_licenses/licenses_linux_user.txt
```diff
@@ -4,15 +4,16 @@ MarkupSafe, 2.1.5, BSD License
 PyYAML, 6.0.2, MIT License
 brevitas, 0.10.2, UNKNOWN
 certifi, 2024.8.30, Mozilla Public License 2.0 (MPL 2.0)
-charset-normalizer, 3.3.2, MIT License
+charset-normalizer, 3.4.0, MIT License
 coloredlogs, 15.0.1, MIT License
+concrete-ml-extensions, 0.1.2, BSD-3-Clause-Clear
```
andrei-stoian-zama, line 9: only for linux, we'll see in the weekly what happens on mac (should fall back to CP compilation)

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -674,9 +680,6 @@ def _prepare_inputs_with_constants(
                     new_input.quantizer.is_qat
                     and not input_.quantizer.is_precomputed_qat
                     and self.error_tracker is not None
```
andrei-stoian-zama, line 682: basically:
- only mixing ops and reshape ops (ops that quantize their inputs) call this function with `quantize_actual_values`
- only QuantizedArray that have `is_precomputed_qat` are now accepted for these functions if `is_qat` (QAT import was requested by the user).
- you don't neeed `QuantIdentity` before reshape/mixing layers applied on other mixing ops outputs because the "passthrough" for `is_precomputed_qat` implemented above.

docs/guides/prediction_with_fhe.md
```diff
@@ -112,11 +112,12 @@ class FCSmall(nn.Module):
         super().__init__()
         self.quant_input = qnn.QuantIdentity(bit_width=3)
         self.fc1 = qnn.QuantLinear(in_features=input_output, out_features=input_output, weight_bit_width=3, bias=True)
+        self.quant_2 = qnn.QuantIdentity(bit_width=3)
         self.act_f = nn.ReLU()
         self.fc2 = qnn.QuantLinear(in_features=input_output, out_features=input_output, weight_bit_width=3, bias=True)
 
     def forward(self, x):
-        return self.fc2(self.act_f(self.fc1(self.quant_input(x))))
+        return self.fc2(self.quant_2(self.act_f(self.fc1(self.quant_input(x)))))
```
andrei-stoian-zama, line 120: this was bad all along and only worked because the QAT quantization was guessed from the calibration data

src/concrete/ml/pytest/torch_models.py
```diff
@@ -1379,9 +1381,9 @@ def forward(self, x):
         x_pre = []
 
         for i in range(self.n_blocks):
-            x_block = x[:, i, :]
-            q1_out = self.quant_1(x_block)
-            fc1_out = self.fc1(q1_out)
+            q_x = self.quant_1(x)
```
andrei-stoian-zama, line 1384: this was wrong.. the input to the slicing should be quantized with QuantIdentity

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -559,7 +560,10 @@ def _prepare_quantized_input(self, input_: QuantizedArray) -> QuantizedArray:
         # but when parsing the ONNX graph, some options can be overwritten. Thus
         # when evaluating QAT layers we ignore one of these options to allow the
         # override.
-        if quant_opts.is_equal(input_.quantizer.quant_options, ignore_sign_qat=True):
+        if (
+            quant_opts.is_equal(input_.quantizer.quant_options, ignore_sign_qat=True)
+            or input_.quantizer.quant_options.is_precomputed_qat
```
andrei-stoian-zama, line 565: before we were overriding the op input quantization mainly for the model input. now we override every time there is a BrevitasQuant layer before the op

tests/torch/test_compile_torch.py
```diff
@@ -603,7 +603,7 @@ def test_compile_torch_or_onnx_activations(
 @pytest.mark.parametrize(
     "model",
     [
-        pytest.param(SimpleQAT),
+        pytest.param(StepFunctionPTQ),
```
andrei-stoian-zama, line 606: this was a non-brevitas QAT model for which QAT quantization had to be guessed. we don't support that anymore, so we use ptq here

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -375,25 +561,45 @@ def __init__(
         self.configuration: Optional[Configuration] = None
         self.model_name = model_name
         self.verbose = verbose
+
+        self.default_crypto_params_glwe = (
+            json.loads(fhext.default_params())  # pylint: disable=no-member
+            if _HAS_GLWE_BACKEND
+            else None
+        )
```
jfrery, line 569: that should be in the glwe class

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -585,10 +849,7 @@ def clear_private_info(module):
 
         # Save the model with a specific filename
         model_path = path / "model.pth"
-
-        # Save the model state dict due to a Brevitas issue
-        # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4572
-        torch.save(self.model.state_dict(), model_path.resolve())
+        torch.save(self.model, model_path.resolve())
```
jfrery, line 852: thanks!

tests/torch/test_hybrid_converter.py
```diff
@@ -106,40 +123,52 @@ def run_hybrid_llm_test(
...
         temp_dir_path = Path(temp_dir)
 
         # Get the temp directory path
-        hybrid_model.save_and_clear_private_info(temp_dir_path)
-        hybrid_model.set_fhe_mode("remote")
 
-        # At this point, the hybrid model does not have
-        # the parameters necessaryto run the module_names
-        module_names = module_names if isinstance(module_names, list) else [module_names]
+        if not has_pbs and glwe_backend_installed:
```
jfrery, line 151: don't we have a better way to check if glwe is used?

tests/torch/test_hybrid_converter.py
```diff
@@ -240,3 +272,85 @@ def test_invalid_model():
...
+        print(f"Value discrepancy detected for GLWE backend, with epsilon={threshold_fhe}")
+        print("Model output (torch fp32)", y_torch[diff])
+        print("Model output (glwe)", y_glwe[diff])
+        print("Model output (quantized clear)", y_qm[diff])
+
+    assert numpy.all(numpy.allclose(y_qm, y_glwe, rtol=1, atol=threshold_fhe))
+    assert numpy.all(numpy.allclose(y_torch, y_glwe, rtol=1, atol=threshold_fhe))
+
+    # Check accuracy between fp32 and glwe
+    assert numpy.abs(acc_fp32 - acc_glwe) < 0.01
```
jfrery, line 353: threshold_fhe?

jfrery (changes requested):

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -364,7 +550,7 @@ def __init__(
         if not isinstance(model, torch.nn.Module):
             raise TypeError("The model must be a PyTorch or Brevitas model.")
 
-        self.model = model
+        self.model = deepcopy(model)
```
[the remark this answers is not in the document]
andrei-stoian-zama: yes, it adds bugs too :) we'll figure it out in the new API Pr

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -414,9 +620,50 @@ def forward(self, x: torch.Tensor, fhe: str = "disable") -> torch.Tensor:
...
+                        private_key=private_key,
+                        compression_key=compression_key,
+                    )
+                )
+
+        result = self.model(x)
+
+        _optimized_linear_executor.set(None)
+
+        return result
```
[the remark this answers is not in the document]
andrei-stoian-zama: how can we get a reference to that executor context in the `forward()` of `self` with this setting ?  with contextvars there is a global object.

tests/torch/test_hybrid_converter.py
andrei-stoian-zama: this is not checking, it's enabling/disabling the glwe backend based on the value of `glwe_backend_installed`

src/concrete/ml/torch/hybrid_model.py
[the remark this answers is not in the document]
jfrery: We can do this I think:

```python
    @contextlib.contextmanager
    def _glwe_executor_context(self):
        try:
            _optimized_linear_executor.set(self.glwe_executor)
            yield
        finally:
            _optimized_linear_executor.set(None)
```

but up to you

src/concrete/ml/torch/hybrid_model.py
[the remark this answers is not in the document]
andrei-stoian-zama: fixed

src/concrete/ml/torch/hybrid_model.py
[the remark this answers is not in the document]
andrei-stoian-zama: I'll keep it in mind for the refactor

deps_licenses/licenses_linux_user.txt
kcelia: Did you run make licenses? 

I thought we were only supposed to do that through GitHub Actions. 
Or was it for a another purpose ?

docs/deep-learning/fhe_assistant.md
```diff
@@ -77,7 +77,7 @@ concrete_clf.compile(X, debug_config)
 
 #### 3. Quantization import failed
 
-**Error message**: `Error occurred during quantization aware training (QAT) import [...] Could not determine a unique scale for the quantization!`.
+**Error message**: `Error occurred during quantization aware training (QAT) import [...] Are you missing a QuantIdentity layer in your Brevitas model?`.
```
kcelia, line 80: We also have a non-intuitive  error that occurs when using 'view' instead of 'reshape'.

Error: ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size ... is different from ...)

Maybe we could add it ? CC @andrei-stoian-zama

docs/guides/prediction_with_fhe.md
kcelia: I'm really not a fan of using quantIdentity layers after quantLayers. 

It works fine without them, and including them just adds unnecessary complexity in my opinion.

src/concrete/ml/quantization/base_quantized_op.py
kcelia: Thanks for the explanation, maybe you should update the comment accordingly.

src/concrete/ml/quantization/quantized_ops.py
kcelia, line 506: sanity check: assert len(inputs) >= 2
no ?

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -491,6 +491,68 @@ def copy_function(x):
...
+
+        # Using snake case here to please the Python format, the original attrs don't have the '_'
+        # Use default false so we also support MatMul impl, MatMul does not have these flags
+        transpose_inputs2 = self.attrs.get("transB", False)
+
+        p = q_input2.qvalues.shape[-2]
+
+        assert q_input1.quantizer.scale is not None
+        assert q_input2.quantizer.scale is not None
+        m_matmul = q_input1.quantizer.scale * q_input2.quantizer.scale
```
kcelia, line 525: I would have added a comment (even though it's trivial, but it might help in the future):

out_scale = scale_input1 × scale_input2
​

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -491,6 +491,68 @@ def copy_function(x):
...
+
+        input2_q_values = (
```
kcelia, line 527: I would have grouped those rows, for better clarity:

# Transpose input2, if needed
transpose_inputs2 = self.attrs.get("transB", False)
input2_q_values = (
    numpy.transpose(q_input2.qvalues) if transpose_inputs2 else q_input2.qvalues
        )

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -491,6 +491,68 @@ def copy_function(x):
...
+        # Using snake case here to please the Python format, the original attrs don't have the '_'
+        # Use default false so we also support MatMul impl, MatMul does not have these flags
+        transpose_inputs2 = self.attrs.get("transB", False)
+
+        p = q_input2.qvalues.shape[-2]
+
+        assert q_input1.quantizer.scale is not None
+        assert q_input2.quantizer.scale is not None
+        m_matmul = q_input1.quantizer.scale * q_input2.quantizer.scale
+
```
kcelia, line 526: I would have added a comment to show the formula of out_zp (even though it's trivial, but it might help in the future):

out_zp = sum_weights − final_term
out_zp = zp_input1 ​× ∑input2_i − p × zp_input1 ​× zp_input2

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -1,32 +1,47 @@
...
 from typing import Dict, List, Optional, Tuple, Union
 
+try:
+    import concrete_ml_extensions as fhext
+
+    _HAS_GLWE_BACKEND = True
+except ImportError:  # pragma: no cover
+    fhext = None
+    _HAS_GLWE_BACKEND = False
+
```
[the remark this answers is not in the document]
kcelia: You can also use  `importlib` for conditional import, like:
```
import importlib

fhext = importlib.import_module('x') if importlib.util.find_spec("x") else None
_HAS_GLWE_BACKEND = fhext is not None
```

src/concrete/ml/quantization/quantized_ops.py
jfrery: I am not sure I understand the whole analytical part on this PR. Why are we doing this?

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -720,12 +725,40 @@ def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
...
+        supports_analytical_quant = all(
+            isinstance(qv, QuantizedArray) for qv in inputs
+        ) and isinstance(self, QuantizedMixingOp)
+        if supports_analytical_quant:
+            q_prepared_inputs = self._prepare_inputs_with_constants(
+                *inputs, calibrate=False, quantize_actual_values=True
+            )
+            quantizer = self.calibrate_analytical_output(*q_prepared_inputs)
+            self.output_quant_params = quantizer.quant_params
+            self.output_quant_stats = quantizer.quant_stats
```
jfrery, line 737: Why are we doing this? We should always have the output based on the calibration quantized params since we want a specific bit-width, no? This analytical output quantizer would mean we don't requantize the output to a lower bitwidth.

jfrery (changes requested):

src/concrete/ml/quantization/base_quantized_op.py
andrei-stoian-zama: yes, exactly, the analytically computed quantizers are only used for dequantization

andrei-stoian-zama: `I am not sure I understand the whole analytical part on this PR. Why are we doing this?`

because doing this avoids having to actually perform the computation of the layer outputs for the calibration. thus it's much faster. since we only need the parameters for dequantization we can skip the actual computation

src/concrete/ml/quantization/quantized_ops.py
```diff
@@ -491,6 +491,68 @@ def copy_function(x):
...
+        q_input1 = inputs[0]
+        assert isinstance(q_input1, QuantizedArray)
+        q_input2 = inputs[1]
+        assert isinstance(q_input2, QuantizedArray)
+
+        # In the operation Y = alpha * A' * B' + beta * C, q_bias is used for
+        # generalised matrix multiplication. q_bias is set to None for standard
+        # matrix multiplication (beta == 0 or only two inputs)
+        q_bias = None if len(inputs) == 2 or self.attrs["beta"] == 0 else inputs[2]
```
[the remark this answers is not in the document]
andrei-stoian-zama: if a bias is specified for this op there are 3 inputs

docs/deep-learning/fhe_assistant.md
andrei-stoian-zama: that's not related, it's due to shape inference during the onnx export.

src/concrete/ml/quantization/post_training.py
```diff
@@ -478,6 +515,17 @@ def _quantize_layers(self, *input_calibration_data: numpy.ndarray):
 
         constants: Set[str] = set(self.quant_params.keys())
 
+        # Check if the model has only GLWE supported linear layers.
+        # In this case, use analytical calibration which is much faster
+        fast_calibration = True
+        for node in graph.node:
```
jfrery, line 521: that's adding another full traversal of the onnx graph here. We should probably exit the loop when we encounter an op that's not supported

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -720,12 +725,40 @@ def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
         if isinstance(raw_result, RawOpOutput):
             return raw_result
 
-        quantized_samples = QuantizedArray(self.n_bits, raw_result)
+        supports_analytical_quant = all(
+            isinstance(qv, QuantizedArray) for qv in inputs
+        ) and isinstance(self, QuantizedMixingOp)
```
jfrery, line 730: Shouldn't this be in QuantizedMixingOp?

src/concrete/ml/quantization/base_quantized_op.py
jfrery: Alright so that's an optimization that will only be applied for linear only circuit. Though I am surprised it works since the condition is just for the op to be a QuantizedMixingOp. Why isn't the e.g. QuantizedAdd making our tests fail since it doesn't have an implementation for this calibrate_analytical_output ?

andrei-stoian-zama: because we only test for the QuantizedGemm / QuantizedMatmul case, while for the other MixingOps we test that an error is raised if one attempts to call calibrate_analytically on them

src/concrete/ml/quantization/post_training.py
andrei-stoian-zama: yes, you're right. traversing the graph in this way is pretty fast though, I'm guessing the graph is in memory and we're just processing some pointers.

src/concrete/ml/quantization/base_quantized_op.py
andrei-stoian-zama: yes, it could be a base function that returns False and an overload in the QuantizedMixingOp that implements the condition

src/concrete/ml/quantization/base_quantized_op.py
jfrery: The quantized add must be tested in torch. Why isn't this failing?

src/concrete/ml/quantization/base_quantized_op.py
```diff
@@ -720,12 +725,40 @@ def calibrate(self, *inputs: numpy.ndarray) -> numpy.ndarray:
         if isinstance(raw_result, RawOpOutput):
             return raw_result
 
-        quantized_samples = QuantizedArray(self.n_bits, raw_result)
+        supports_analytical_quant = all(
```
jfrery, line 728: Can you explain in a comment when we are supposed to have this true or false?

deps_licenses/licenses_linux_user.txt
andrei-stoian-zama: yes, I generated the license files in the action. but this library only works on linux so only the linux license file shows it

docs/guides/prediction_with_fhe.md
andrei-stoian-zama: it used to work because of some automatic detection of quant parameters that we were doing. but I removed that detection since it was slow. so now we need quantidentity in the right places

jfrery (approved):
Thanks!

PR #871, opened 2024-09-13
add monkeypatch context to test_load_fitted_sklearn_tree_models

kcelia:
This bug was reported by the [Concrete team](https://zama-ai.slack.com/archives/C044G400J6N/p1726212715764299), who encountered an issue with the value of the environment variable “TREE_USE_ROUNDING” during their tests.

The TREE_USE_ROUNDING variable is set to “1” by default in our source code, which means that tree-based models are always used with the rounding feature enabled.

However, in our Pytest tests (tests/sklearn/test_sklearn_models.py), particularly in the tests “test_load_fitted_sklearn_tree_models” and “check_rounding_consistency”, we temporarily change the value of this variable to test the rounding consistency feature.

So, the 2 tests mentioned above can conflict if they are executed in parallel, as both modify the TREE_USE_ROUNDING variable. To isolate the changes and ensure they only impact the specific test, it is recommended to use  a monkeypatch when safely modifying environment variables and  avoid potential conflicts during tests.

Hence, adding a monkeypath to "test_load_fitted_sklearn_tree_models" function (as in "check_rounding_consistency").

### DIFF
--- a/tests/sklearn/test_sklearn_models.py
+++ b/tests/sklearn/test_sklearn_models.py
@@ -1106,191 +1106,201 @@ def test_load_fitted_sklearn_tree_models(
     """Test `from_sklearn_model` functionnality of tree-based models."""
 
     numpy.random.seed(0)
-    os.environ["TREES_USE_ROUNDING"] = str(int(use_rounding))
+    with pytest.MonkeyPatch.context() as mp_context:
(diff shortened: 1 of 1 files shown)


andrei-stoian-zama (approved):
Thanks

PR #873, opened 2024-09-16
chore: fixing a typo

bcm-at-zama:
Had a look to https://github.com/zama-ai/concrete-ml/tree/main/use_case_examples/deployment/server and I think there is a typo.

By the way: maybe we are missing a README in https://github.com/zama-ai/concrete-ml/tree/main/use_case_examples/deployment/? Should we move https://github.com/zama-ai/concrete-ml/tree/main/use_case_examples/deployment/server/README there? Or have another one?

It's because people say they have difficulties in deploying, see https://discord.com/channels/901152454077452399/901152454622740593/1285222791104303134 and further messages.

### DIFF
--- a/use_case_examples/deployment/server/README.md
+++ b/use_case_examples/deployment/server/README.md
@@ -10,8 +10,8 @@ We show-case how to do this on 3 examples:
 You can run these example locally using Docker, or on AWS if you have your credentials set up.
 
 For all of them the workflow is the same:
-0\. Optional: Train the model
 
+1. Optional: Train the model
 1. Compile the model to an FHE circuit
 1. Deploy to AWS, Docker or localhost
 1. Run the inference using the client (locally or in Docker)


jfrery (approved):

bcm-at-zama:
@jfrery : maybe also you could tell me what you think about the second part of the comment, ie

> By the way: maybe we are missing a README in https://github.com/zama-ai/concrete-ml/tree/main/use_case_examples/deployment/? Should we move https://github.com/zama-ai/concrete-ml/tree/main/use_case_examples/deployment/server/README there? Or have another one?
> 
> It's because people say they have difficulties in deploying, see https://discord.com/channels/901152454077452399/901152454622740593/1285222791104303134 and further messages.

jfrery:
The readme doesn't really help.

I think we should have a docker image ready or simple script that does the server part and can be launched with only the `server.zip`.

CML is required for the client part so we can probably abstract away the complexity of deployment as well.

Let's discuss this.

bcm-at-zama:
My understanding was that:
- use_case_examples/deployment/breast_cancer/README.md
- use_case_examples/deployment/cifar/README.md
- use_case_examples/deployment/sentiment_analysis/README.md
were quite informative (aren't they?) so maybe we would have a use_case_examples/deployment/README which would link to them, and tell to read the details there?

jfrery:
We can have a readme at the top yes that makes sense. Let's merge this and we will create a high level readme.

andrei-stoian-zama (approved):

PR #874, opened 2024-09-19
chore: renaming the notebook for more clarity

bcm-at-zama:
The notebook had no obvious "FHE training" aspect in its name or in its title (Logistic Regression Training). I think it's a bit misleading for the quick readers, isn't it? Changing the name of the file is bad, it may break links, but what about changing its title to something like "Logistic Regression Training Over Encrypted Dataset"?

### DIFF
--- a/docs/advanced_examples/LogisticRegressionTraining.ipynb
+++ b/docs/advanced_examples/LogisticRegressionTraining.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Logistic Regression Training\n",
+    "# Logistic Regression Training on Encrypted Dataset\n",
     "\n",
     "This notebook shows how to train a logistic regression model on encrypted data using stochastic gradient descent (SGD). During this process,\n",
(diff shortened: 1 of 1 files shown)


docs/advanced_examples/LogisticRegressionTraining.ipynb
```diff
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Logistic Regression Training\n",
+    "# Logistic Regression Training over Encrypted Dataset\n",
```
jfrery, line 7: I have been told it's `on` and not `over` in this context.

jfrery (changes requested):

docs/advanced_examples/LogisticRegressionTraining.ipynb
bcm-at-zama: I would have kept in mind that both were ok, but ok I changed it to 'on'

jfrery (approved):

andrei-stoian-zama (approved):

PR #876, opened 2024-09-19
chore: add readme in root folder of deployment use case + fix cifar

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4600

ref #873 
cc @bcm-at-zama

### DIFF
--- a/use_case_examples/deployment/README.md
+++ b/use_case_examples/deployment/README.md
@@ -1,0 +1,42 @@
+# Deployment Examples
+
+This folder contains examples of how to deploy Concrete ML models using Fully Homomorphic Encryption (FHE). These examples demonstrate the process of training, compiling, and deploying models for various use cases.
+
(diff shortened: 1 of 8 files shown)


andrei-stoian-zama (approved):

andrei-stoian-zama (changes requested):
Wait.. why are we duplicating the cifar use case here in deployment ?

jfrery:
> Wait.. why are we duplicating the cifar use case here in deployment ?

We are duplicating the model. There was [a change](https://github.com/zama-ai/concrete-ml/pull/538/files#diff-0ce00a4975eb5fe4698e298d62c3c25aab55e01de9e16e91c4b384e2c1baa848L10-R10) to use the cifar10 training model but it was not imported.

andrei-stoian-zama (approved):

PR #883, opened 2024-09-24
docs: add mlp lora example

andrei-stoian-zama:
Adding an example for fine-tuning a MLP with LORA.


andrei-stoian-zama:
This was moved to @jfrery 's branch

PR #893, opened 2024-09-27
fix: fixing the call to train.py directly in python

bcm-at-zama:
else we have issues with the versions

Checking the breast cancer deployment on my mac:

```
cd use_case_examples/deployment/breast_cancer

# We need to force 3.8; eg, would not work with 3.10
rm -rf .venvdeployment
python3.8 -m venv .venvdeployment
source .venvdeployment/bin/activate

# 1.6.1 correspond to which version? Something set where, in the CML docker maybe?
pip uninstall concrete-ml; pip install concrete-ml==1.6.1

rm -rf ./dev 
python ./train.py 
ls ./dev 

    > client.zip  server.zip

python ../server/deploy_to_docker.py --path-to-model ./dev
```

works fine.


andrei-stoian-zama (changes requested):
Please keep the port 5000. Or maybe if you find a port that always works on MAC that's good too

use_case_examples/deployment/breast_cancer/client_via_tfhe-rs.py
```diff
@@ -0,0 +1,208 @@
+"""Client script.
```
bcm-at-zama, line 1: Ill remove this file

PR #907, opened 2024-10-01
fix: flaky cnn in simulate

andrei-stoian-zama:
The CNN test is poorly simulated (but works fine in FHE) and thus makes the test fail often. I don't know why it happens. All other models work fine.

### DIFF
--- a/tests/torch/test_compile_torch.py
+++ b/tests/torch/test_compile_torch.py
@@ -79,7 +79,7 @@ def create_test_inputset(inputset, n_percent_inputset_examples_test):
 
 
 def get_and_compile_quantized_module(
-    model, inputset, import_qat, n_bits, configuration, verbose, device
+    model, inputset, import_qat, n_bits, n_rounding_bits, configuration, verbose, device
 ):
     """Get and compile the quantized module built from the given model."""
     quantized_numpy_module = build_quantized_module(
@@ -87,6 +87,7 @@ def get_and_compile_quantized_module(
         inputset,
         import_qat=import_qat,
         n_bits=n_bits,
+        rounding_threshold_bits=n_rounding_bits,
     )
 
     p_error, global_p_error = manage_parameters_for_pbs_errors(None, None)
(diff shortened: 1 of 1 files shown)


tests/torch/test_compile_torch.py
```diff
@@ -146,15 +147,27 @@ def compile_and_test_torch_or_onnx(  # pylint: disable=too-many-locals, too-many
...
+    )
 
-        n_bits = (
-            {"model_inputs": 2, "model_outputs": 2, "op_inputs": 2, "op_weights": 2}
-            if qat_bits == 0
-            else qat_bits
-        )
+    n_rounding_bits = 6
 
+    # FHE vs Quantized are not done in the test anymore (see issue #177)
```
kcelia, line 169: Maybe you should, put the issue ID, in the right format

tests/torch/test_compile_torch.py
```diff
@@ -146,15 +147,27 @@ def compile_and_test_torch_or_onnx(  # pylint: disable=too-many-locals, too-many
     else:
         inputset = (numpy.random.uniform(-100, 100, size=(n_examples, *to_tuple(input_shape))),)
 
-    # FHE vs Quantized are not done in the test anymore (see issue #177)
-    if not simulate:
+    # Compile our network with the same bitwidth in simulation and FHE
+    if qat_bits == 0:
```
kcelia, line 151: are we really testing with qat_bits = 0 ?! 
For which case is it ?
Can you add a comment, to explain it, please.

andrei-stoian-zama: qat_bits==0 simply means use PTQ

tests/torch/test_compile_torch.py
andrei-stoian-zama: We have a syntax for TODO, which link to open issues . but this is not an open one, the reference here is for reference purposes only. so this is fine as it is

PR #921, opened 2024-10-15
chore: remove pip extra url

jfrery:
hopefully install deps is much faster


andrei-stoian-zama (approved):
thanks

jfrery:
install dependencies took 3min instead of 2 hours that's much better.

PR #942, opened 2024-11-14
chore: fix hybrid model glwe lora mlp

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4671

### DIFF
--- a/src/concrete/ml/torch/hybrid_model.py
+++ b/src/concrete/ml/torch/hybrid_model.py
@@ -267,6 +267,7 @@ class RemoteModule(nn.Module):
 
         elif self.fhe_local_mode == HybridFHEMode.REMOTE:  # pragma:no cover
             # Remote call
+            # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4672
             assert self.executor is None, "Remote optimized linear layers are not yet implemented"
             y = self.remote_call(x)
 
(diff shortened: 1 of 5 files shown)


src/concrete/ml/torch/hybrid_model.py
```diff
@@ -476,17 +460,16 @@ def forward(self, x: torch.Tensor, fhe: str = "disable") -> torch.Tensor:
 
                 # Loading keys from a file could be done here, and the
                 # keys could be passed as arguments to the Executor
-                executor = GLWELinearLayerExecutor()
-
+                self.executor = GLWELinearLayerExecutor()
                 if fhe_mode != HybridFHEMode.DISABLE:
-                    executor.keygen()
+                    self.executor.keygen()
```
andrei-stoian-zama, line 465: this is doing for every hybridmodel.forward no ? 

I wanted to avoid that: one keygen should be done a single time and the key re-used

Moreover, we should think about that, at some point,we need the option to load keys from a key store

andrei-stoian-zama (changes requested):
I think we need a way to ensure we don't keygen every time the `forward` of the hybrid model is called

src/concrete/ml/torch/hybrid_model.py
jfrery: That's not how the hybrid model with concrete works. We can't assume that all circuit can reuse the same key. But yes I get your point.

jfrery: I updated. We should only have the initialization and keygen once

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -400,13 +382,14 @@ def __init__(
...
         self.verbose = verbose
+        self.executor: Optional[GLWELinearLayerExecutor] = None
 
         self._replace_modules()
 
     def _replace_modules(self):
         """Replace the private modules in the model with remote layers."""
 
-        self._all_layers_are_pure_linear = True
+        self._has_large_linear_layers = True
```
andrei-stoian-zama, line 392: I think something more like `_has_only_large_linear_layers`

andrei-stoian-zama (changes requested):

src/concrete/ml/torch/hybrid_model.py
jfrery: Done

andrei-stoian-zama (dismissed):

andrei-stoian-zama (approved):

andrei-stoian-zama (approved):

PR #943, opened 2024-11-15
chore: update weekly and release machines

jfrery:
closes https://github.com/zama-ai/concrete-ml-internal/issues/4669
same vcpu machine but more available. ~Also use hpc7a for release~ Actually we need to switch region to get the hpc7a


andrei-stoian-zama (approved):
thanks!

PR #950, opened 2024-12-05
fix: py3.12 install deps

andrei-stoian-zama:
Python 3.12 deps don't install in the CI (but works locally).

First idea: update poetry


kcelia (approved):
thanks

tests/torch/test_compile_torch.py
```diff
@@ -863,7 +863,6 @@ def test_pretrained_mnist_qat(
     quantized_numpy_module = compile_onnx_model(
         onnx_model,
         inputset,
-        import_qat=True,
```
jfrery, line 866: I don't get this. We are in the test `test_pretrained_mnist_qat` why are we not using the import_qat now? And why did it work before?

andrei-stoian-zama: the model here is a non-brevitas, custom QAT, which we don't support anymore (to avoid computation cost and also because nobody used this feature which was very confusing)

jfrery: Why did that work before and not anymore?

andrei-stoian-zama: there was a similar test in the regular CI that was also changed the same way when merging the py3.12 PR. but this one which only runs in the weekly was not changed at that time.

jfrery: I see. Got it.

PR #957, opened 2024-12-09
fix: faster rounding test in weekly

andrei-stoian-zama:
Speed up the weekly test for rounding consistency


jfrery (commented):
I would remove the FHE run here. This test is probably one of the last running crt PBS which is taking very long.

kcelia (approved):

PR #962, opened 2024-12-17
feat: add logs + evaluation llama lora

jfrery:
- src/concrete/ml/quantization/linear_op_glwe_backend.py:

  - dynamic quantization
  - per token input quantization
  - per channel weight quantization .
 
- src/concrete/ml/torch/lora.py:
  - logging for training and evaluation.
  - dynamic quantization support (new param)
  - checkpoint saving and loadin.

- tests/torch/test_lora.py:

  - tests for dynamic quantization.
  - removed obsolete toggle_calibrate tests.
  - checkpoint saving and loading tests.

- use_case_examples/llm/qgpt2_class.py:

  - fixed forbidden word in comments.

- use_case_examples/lora_finetuning/README.md:

  - Added evaluation script description.
  - Updated training results visualization and final training losses.

- use_case_examples/lora_finetuning/eval.py:

  - evaluation script for LLaMA model training with different modes (torch, 8-bit, 16-bit).


src/concrete/ml/torch/hybrid_model.py
```diff
@@ -223,6 +223,16 @@ def _apply(self, fn, recurse=True):
...
         return self
 
+    def _ensure_module_on_device(self, x: torch.Tensor) -> None:
+        """Ensure the private module is on the same device as the input tensor.
+
+        Args:
+            x (torch.Tensor): The input tensor to match device with.
+        """
+        assert self.private_module is not None
+        if next(self.private_module.parameters()).device != x.device:
```
andrei-stoian-zama: can't there be multiple `parameters` in the `private_module`, why not iterate ?

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -612,7 +630,7 @@ def compile_model(
                     vals = self.private_q_modules[name].quant_layers_dict.values()
                     _, q_op = next(iter(vals))
                     const_inp = q_op.constant_inputs[1]  # Get the weights, the bias is in [2]
-                    const_inp.values = const_inp.qvalues.astype(numpy.float32)
+                    # const_inp.values = const_inp.qvalues.astype(numpy.float32)
```
andrei-stoian-zama, line 637: should't we allow this optimization if the quantization is static ?

src/concrete/ml/torch/lora.py
```diff
@@ -184,14 +223,6 @@ def _replace(module: nn.Module):
 
         _replace(model)
 
-    def toggle_calibrate(self, enable: bool = True):
```
andrei-stoian-zama, line 187: isn't this method used in the Gpt2 notebook?

use_case_examples/lora_finetuning/README.md
```diff
@@ -56,6 +56,29 @@ Each notebook includes its own data-set:
 
 Execute the Jupyter notebook `GPT2FineTuneHybrid.ipynb` to start the fine-tuning process. The notebook is structured into several steps:
 
+### Evaluation Script
```
andrei-stoian-zama, line 59: you mention the math problem training, but where is the notebook for it?

use_case_examples/lora_finetuning/README.md
```diff
@@ -78,9 +101,35 @@ After fine-tuning, the model's weights are distributed between the client and se
...
 
-### LLaMA Results
+#### Training Results Visualization
+
+![LoRA Fine-tuning Quantized Loss Comparison](./assets/fine_tuning_loss.svg)
+
+*LoRA Fine-tuning Quantized Loss Comparison*
+
+- **FP32**: Full precision (32-bit floating point) baseline training
+- **ST-ST**: Static per-Tensor quantization with fixed ranges
```
andrei-stoian-zama, line 113: do we need to showcase static quantization ? we know it doesn't work well.

andrei-stoian-zama (changes requested):
Do you need to upload the math problem notebook ? 

Could you re-run the LLAMA and GPT2 notebooks to make sure they work well. Maybe we should switch them to dyanmic quant as well ?

src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -30,7 +31,15 @@ def __init__(
...
         compression_key=None,
+        use_dynamic_quantization: bool = True,
     ):
+        """Initialize the GLWE linear layer executor.
+
+        Args:
+            private_key: Private key for encryption
+            compression_key: Compression key for encryption
+            use_dynamic_quantization: Whether to use dynamic quantization
+        """
```
kcelia, line 42: Is there a reason why `import concrete_ml_extensions as fhext` is not at the beginning of the script?

src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -56,71 +66,269 @@ def keygen(self):
             self.glwe_crypto_params
```
kcelia, line 66: mmm 
i thought that `self.fhext.create_private_key(` generates:
1-  the private key
Then to get the compressed key, we use: `fhext.deserialize_compression_key(server_key.serialize())`

src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -56,71 +66,269 @@ def keygen(self):
...
+        w_max_vals, _ = weight_float.max(dim=0, keepdim=True)
+
+        weight_scale = (w_max_vals - w_min_vals) / (q_max - q_min)
+        # Avoid division by zero.
+        weight_scale = torch.where(
+            (w_max_vals > w_min_vals), weight_scale, torch.ones_like(weight_scale)
+        )
+        weight_scale = weight_scale.squeeze(-1)  # shape: (out_dim,)
+
+        weight_zp = torch.round((q_min - w_min_vals) / weight_scale).to(torch.float32)
```
kcelia, line 107: In affine quantization, the zero point is an int, so maybe we should cast it to torch.int32

src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -129,42 +337,37 @@ def forward(
         q_x = q_module.quantize_input(x.cpu().numpy())
         assert q_x is not None
         assert isinstance(q_x, numpy.ndarray)
-
         q_x = numpy.transpose(q_x) if transpose_inputs1 else q_x
 
-        # Need to slice the last GLWE (this will be improved in later cml-extensions)
```
kcelia, line 135: why did you remove this comment ?
it's not the case anymore?

src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -129,42 +337,37 @@ def forward(
...
         num_valid_glwe_values_in_last_ciphertext = (
-            q_weight.shape[1] % self.poly_size or self.poly_size
+            qweight.shape[1] % self.poly_size or self.poly_size
         )
 
-        # The GLWE backend needs uint64 encoding for both neg/pos values
```
kcelia, line 140: same comment

src/concrete/ml/torch/hybrid_model.py
jfrery: Yes better check if any of the param is not on the same device.

src/concrete/ml/torch/hybrid_model.py
jfrery: Ok I add it back for static quantization. Not sure the gain though.

src/concrete/ml/torch/lora.py
jfrery: Good catch. I will update GPT2 notebook. We should rerun the two notebook anyway.

use_case_examples/lora_finetuning/README.md
jfrery: There is no notebook. The line after mentions the `eval.py` script

use_case_examples/lora_finetuning/README.md
```diff
@@ -78,9 +101,35 @@ After fine-tuning, the model's weights are distributed between the client and se
...
+  - Input: Static per-Tensor quantization
+  - Weight: Static per-Tensor quantization
+- **DTen-ST**: Dynamic/Static hybrid quantization
```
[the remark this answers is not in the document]
jfrery: I think the information should remain somewhere. Should I create another .md where I store this?

src/concrete/ml/quantization/linear_op_glwe_backend.py
jfrery: Yes, the library might not be installed / be available. Since this file is imported in the hybrid model, we don't want to import it by default.

src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -173,18 +376,126 @@ def forward(
...
+
+        Returns:
+            torch.Tensor: Output tensor after applying the linear operation in FHE mode
+        """
+        if self.private_key is None:
+            self.keygen()  # pragma: no cover
+
+        device = x.device
+        # Dynamic quantization for weights and input.
+        weight_q, weight_scale, weight_zp, sum_w = self._per_channel_weight_quantization(
```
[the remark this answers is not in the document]
jfrery: yes you are right. We do input dynamic quantization. Static per channel weight quantization.

These weight shouldn't be re-computed everytime. Ideally I should store them back in the quantized module. I will add a FIXME

src/concrete/ml/quantization/linear_op_glwe_backend.py
jfrery:
```
#[pyfunction]
fn cpu_create_private_key(
    crypto_params: &MatmulCryptoParameters,
) -> PyResult<(PrivateKey, CpuCompressionKey)> {
    let (glwe_secret_key, post_compression_glwe_secret_key, compression_key) =
        create_private_key_internal(crypto_params);

    return Ok((
        PrivateKey {
            inner: glwe_secret_key,
            post_compression_secret_key: post_compression_glwe_secret_key,
        },
        CpuCompressionKey {
            inner: compression_key,
            buffers: compression::CpuCompressionBuffers::<Scalar> { _tmp: PhantomData },
        },
    ));
}
```

create_private_key returns the compressed key as well

src/concrete/ml/quantization/linear_op_glwe_backend.py
jfrery: good catch. It's still the case.

src/concrete/ml/quantization/linear_op_glwe_backend.py
jfrery: yep... adding it back.

src/concrete/ml/quantization/linear_op_glwe_backend.py
kcelia: thanks

src/concrete/ml/quantization/linear_op_glwe_backend.py
jfrery: @andrei-stoian-zama figured this out. And doing a quick experiment with torch / numpy float32 vs int32 matmul confirms it.

I think both int32 matmul bypass BLAS optimization...

src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -129,42 +340,111 @@ def forward(
...
 
         # The GLWE backend needs uint64 encoding for both neg/pos values
-        q_weight = q_weight.astype(numpy.int64).astype(numpy.uint64)
+        # Convert weights to required type.
+        qweight = qweight.astype(numpy.int64).astype(numpy.uint64)
 
-        # Some models have (B, C, H)-size activations,
-        # for example LLMs: B=batch size, C=context length, H=hidden dime
-        # while other models only have (B, H)-size activations.
-        # Add a B=1 dimension if needed
```
kcelia, line 146: What about those comments ?

andrei-stoian-zama (changes requested):
1. there's a bug in the GPT2 notebook when LORA layers are activated.
```
# Inference using the fine-tuned model with LoRA weights
# Seed for reproducibility
torch.manual_seed(SEED)

fine_tuned_model.enable_adapter_layers()
prompt = "who invented FHE?"
generate_and_print(prompt, fine_tuned_model, tokenizer, SEED)
     
who invented FHE? , the,, a, in, at, and,
```

2. I think `use_dynamic_quantization` should be enabled by default, and it should be auto-disable if the model is not fully linear. 


3. In a second PR can we move the `eval.py` code to a notebook ?

src/concrete/ml/quantization/linear_op_glwe_backend.py
jfrery: I added them back

jfrery:
> 1. there's a bug in the GPT2 notebook when LORA layers are activated.
> 
> ```
> # Inference using the fine-tuned model with LoRA weights
> # Seed for reproducibility
> torch.manual_seed(SEED)
> 
> fine_tuned_model.enable_adapter_layers()
> prompt = "who invented FHE?"
> generate_and_print(prompt, fine_tuned_model, tokenizer, SEED)
>      
> who invented FHE? , the,, a, in, at, and,
> ```

No this is because we executed GLWE with 16 bits which has overflows. I updated the notebook with 8 bits + dynamic quant. It's much better.

>  2. I think `use_dynamic_quantization` should be enabled by default, and it should be auto-disable if the model is not fully linear.

GPT2 uses the hybrid model to train and not lora so it can't be default. Otherwise it's default for lora training yes. Dynamic quantization is only possible for GLWE currently which itself only work with linear layers. 

> 3. In a second PR can we move the `eval.py` code to a notebook ?

Yes we can.

PR #965, opened 2024-12-19
Update slab GitHub runner to 1.4.0

kcelia:
New version of our Slab CI Server, that brings:

- user cancellation won't induce zombie instances
- Github cancellation is handled, when you have a concurrency directive that kicks-in

The SHA reference is [here](https://github.com/zama-ai/slab-github-runner/commit/79939325c3c429837c10d6041e4fd8589d328bac) 

[Discussion](https://zama-ai.slack.com/archives/C01022E8SRJ/p1734595809899519)


script/actions_utils/update_slab_github_runner.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
kcelia, line 1: I added this file to automatically update the SHA.

Initially, I changed `zama-ai/action/workflows@...` (which has 18 occurrences). Based on that, I thought it would be useful to add a script to handle this automatically.

But, I later realized that the correct reference to update is `zama-ai/slab-github-runner@....`

Would you like to keep the script, or should we remove it? @andrei-stoian-zama?

kcelia: We agreed to remove it.

andrei-stoian-zama (approved):

jfrery (approved):

PR #963, opened 2024-12-18
fix: full gpu hybrid model

andrei-stoian-zama:
- Adds support for GPU (and other devices) for the local computation of the hybrid model.
  - use torch.Tensors for the GLWE backend 
  - keep the tensor device consistent
  - add a Torch quantizer
- Memory optimization:
  - Removes the ONNX model for quantized models built for fully linear layers in the hybrid model
  - Remove calibration data once hybrid model calibration is done
- Upgrades the use-case example CI to Ubuntu 22 (python 3.10)
- Improves the LLAMA lora fine tuning use case (adds FHE execution)

Closes https://github.com/zama-ai/concrete-ml-internal/issues/4682

PR #967, opened 2024-12-20
chore: update lora notebook + fix order inputs in llama

jfrery:


andrei-stoian-zama (dismissed):

use_case_examples/lora_finetuning/requirements.txt
```diff
@@ -4,6 +4,6 @@ peft==0.12.0
 Jinja2==3.1.4
 matplotlib==3.7.5
 datasets==3.1.0
-accelerate==1.2.0
+accelerate==1.0.1
```
kcelia, line 7: Which package the 1.2.0 version had conflicts with?

use_case_examples/lora_finetuning/GPT2FineTuneHybrid.ipynb
```diff
@@ -134,7 +133,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "9775e413ec264b2eb14ee53dbc381474",
```
kcelia, line 137: Since the PR is red, I suggest some updates:

# Apply LoRA  to the model.
# The 'target_modules' parameter can be set to "all-linear" to apply LoRA to all linear modules. 
# By default, only the 'c_attn' projection layers are fine-tuned with LoRA.

kcelia: You can remove the extra spaces between the comment and the function:
```
# Since we've already handled padding and labels, we can use a custom data collator


def data_collator(features):
```

use_case_examples/lora_finetuning/requirements.txt
jfrery: 1.2.0 isn't compatible with python 3.8. I will probably revert this and add a comment that we don't support 3.8 in this use case

kcelia: yes please add a comment

andrei-stoian-zama (changes requested):
I think this is superseded by #963

andrei-stoian-zama:
Closing since #963 was merged

PR #993, opened 2025-01-14
chore: add 3.12 tests in standard ci

jfrery:


.github/workflows/continuous-integration.yaml
```diff
@@ -211,7 +211,7 @@ jobs:
             LINUX_PYTHON_VERSIONS="${{ inputs.linux_python_versions }}"
             MACOS_PYTHON_VERSIONS="${{ inputs.macos_python_versions }}"
           elif [[ "${BUILD_TYPE}" == "pr" ]]; then
-            LINUX_PYTHON_VERSIONS="3.8"
+            LINUX_PYTHON_VERSIONS="3.8 3.12"
```
kcelia, line 214: Shouldn't we add Python 3.12 to push_to_main and push_to_release? 
Just asking

jfrery: I don't think it adds anything to test twice the same commit. We should probably disable push on main but well that's for another time.

andrei-stoian-zama (approved):

PR #997, opened 2025-01-16
feat: Add tfhe-rs compatibility for model inputs /outputs

andrei-stoian-zama:
Adds the `ciphertext_format` parameter to compile calls. 

Needs new fhext & new compiler to run.


src/concrete/ml/sklearn/base.py
```diff
@@ -506,9 +514,33 @@ def _get_module_to_compile(self) -> Union[Compiler, QuantizedModule]:
             Union[Compiler, QuantizedModule]: The module instance to compile.
         """
 
+    def _get_tfhres_module_to_compile(self, inputset):
```
bcm-at-zama, line 517: typo? tfhres -> tfhers

src/concrete/ml/sklearn/base.py
```diff
@@ -651,41 +799,59 @@ def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.
...
+                is_crt_encoding = self.fhe_circuit.statistics["packing_key_switch_count"] != 0
 
-                # If the inference should be executed using simulation
-                if fhe == "simulate":
-                    is_crt_encoding = self.fhe_circuit.statistics["packing_key_switch_count"] != 0
+                # If the virtual library method should be used
+                # For now, use the virtual library when simulating
+                # circuits that use CRT  encoding because the official simulation is too slow
+                # FIXME: https://github.com/zama-ai/concrete-ml-internal/issues/4391
+                if USE_OLD_VL or is_crt_encoding:
```
bcm-at-zama, line 822: I didn't know USE_OLD_VL what still there

tests/sklearn/test_sklearn_models.py
```diff
@@ -2352,3 +2354,73 @@ def test_xgb_serialization_errors(model_class, param, error_message):
...
+    model.compile(x, ciphertext_format=CiphertextFormat.TFHE_RS)
+
+    # Run the model in FHE for TFHE-rs inputs/outputs
+    y_pred_tfhers = model.predict(fhe_test_data, fhe="execute")
+    y_pred_disable = model.predict(fhe_test_data, fhe="disable")
+
+    model.compile(x)
+
+    # Check correctness with TFHE-rs inputs/outputs
+    assert numpy.all(y_pred_tfhers == y_pred_disable)
```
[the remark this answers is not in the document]
andrei-stoian-zama: fhe results (with tfhers inputs) should be equal disable for default p_error. Other tests check that concrete fhe and disable are also equal to disable. using fhe=disable here makes this test faster, it's already quite a slow one

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: it's disabled but still there

tests/sklearn/test_sklearn_models.py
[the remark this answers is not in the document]
bcm-at-zama: I have not understood your answer. 

My question is: do we test that results with results "with Concrete format" and "with TFHE-rs format" are the same? (either in simulate or in FHE). If we have that + existing tests that results in TFHE-rs are equivalent in simulate and fhe, that's fine with me.

src/concrete/ml/sklearn/base.py
```diff
@@ -506,9 +514,33 @@ def _get_module_to_compile(self) -> Union[Compiler, QuantizedModule]:
...
+            res = native_function_to_compile(concrete_x)
+            tfhers_res = tfhers.from_native(res[0], out_dtype_spec)
+            return tfhers_res
+
+        return Compiler(tfhers_tree_inference, {"inputs": "encrypted"})
+
     def compile(
         self,
         X: Data,
+        ciphertext_format: CiphertextFormat = CiphertextFormat.CONCRETE,
```
jfrery, line 543: can you allow simple string like we do for fhe arg? e.g. I would like to call compile with `model.compile(X, ciphertext_format="tfhe-rs")`

src/concrete/ml/sklearn/base.py
```diff
@@ -651,41 +799,59 @@ def predict(self, X: Data, fhe: Union[FheMode, str] = FheMode.DISABLE) -> numpy.
...
         # If the inference is executed in FHE or simulation mode
         if fhe in ["simulate", "execute"]:
             # Check that the model is properly compiled
             self.check_model_is_compiled()
 
-            q_y_pred_list = []
-            for q_X_i in q_X:
-                # Expected encrypt_run_decrypt output shape is (1, n_features) while q_X_i
-                # is of shape (n_features,)
-                q_X_i = numpy.expand_dims(q_X_i, 0)
```
jfrery, line 663: wasn't this necessary?

tests/sklearn/test_sklearn_models.py
```diff
@@ -2352,3 +2354,73 @@ def test_xgb_serialization_errors(model_class, param, error_message):
...
+    # If the model is not supported or if the n_bits is not supported
+    # an error is raised
+    if not n_bits == 8 or is_regressor_or_partial_regressor(model_class):
+        with pytest.raises(AssertionError, match=".*supported for 8-bit tree-based.*"):
+            model.compile(x, ciphertext_format=CiphertextFormat.TFHE_RS)
+        return
+
+    # Check that we can first compile to Concrete, then to
+    # TFHE-rs input/outputs then to concrete again
+    model.compile(x)
```
jfrery, line 2384: You could add model.predict with concrete maybe and add this to the comparisons. We expect them to be equal to disable and to execute with tfhe-rs input

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: Old vl is simply CP graph execution, so it is tested in Concrete. but you're right in the sense that we could easily remove it - we kept it thinking it might be useful at some point, but it's not anymore

tests/sklearn/test_sklearn_models.py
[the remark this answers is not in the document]
andrei-stoian-zama: we compare tfhe-rs inputs vs disable in this test and we compare concrete inputs vs simulate and disable in other tests.

tests/sklearn/test_sklearn_models.py
andrei-stoian-zama: ok, as Benoit asks for it too I'll add it

src/concrete/ml/sklearn/base.py
bcm-at-zama: > it is tested in Concrete
I don't know if they still test it. At some point, they'll certainly remove it as well

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: it's just been moved a bit lower in the file

src/concrete/ml/sklearn/base.py
```diff
@@ -597,6 +671,13 @@ def compile(
         self._is_compiled = True
         self._compiled_for_cuda = use_gpu
 
+        self._ciphertext_format = ciphertext_format
+        if ciphertext_format == CiphertextFormat.TFHE_RS:
+            self._tfhers_bridge = tfhers.new_bridge(self.fhe_circuit_)
+        else:
+            self._tfhers_bridge = None
+
+        #        print(self.fhe_circuit_.graph.format())
```
jfrery, line 680: to delete

src/concrete/ml/sklearn/base.py
```diff
@@ -580,6 +638,22 @@ def compile(
...
+        if ciphertext_format == CiphertextFormat.TFHE_RS:
+            assert all(
+                (
+                    self.input_quantizers[i].is_signed == self.input_quantizers[0].is_signed
+                    for i in range(1, len(self.input_quantizers))
+                )
+            )
+            is_signed = self.input_quantizers[0].is_signed
+
+            crypto_params = json.loads(fhext.get_crypto_params_radix())  # pylint: disable=no-member
```
jfrery, line 650: why do we use deai crypto params ?

jfrery (commented):
Thanks! I reviewed. I have a few comments + I am not sure I understand why we use concrete-ml-extension here?

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: we need to get the radix parameters from somewhere. Concrete doesn't set the parameters, so Concrete ML must set them, using default parameters from Rust. Fhext is the only Rust package where we can set these parameters.

jfrery: Got it

src/concrete/ml/sklearn/base.py
andrei-stoian-zama: fixed

jfrery (approved):

bcm-at-zama (commented):
Will you add some doc about that new feature as well? With execution times, to avoid they ask you why it's 4x slower

PR #1006, opened 2025-01-25
feat(deployment): Make serialization optional in FHEModelServer.run

DeVikingMark:
Add force_serialization parameter to FHEModelServer.run method to make
serialization/deserialization optional. This allows for better performance
in cases where serialization is not needed.

- Add force_serialization parameter with default value False
- Update method documentation
- Modify serialization logic to consider both input state and force_serialization flag
- Remove TODO comment as the functionality is now implemented

### DIFF
--- a/src/concrete/ml/deployment/fhe_client_server.py
+++ b/src/concrete/ml/deployment/fhe_client_server.py
@@ -141,23 +141,25 @@ class FHEModelServer:
             bytes, fhe.Value, Tuple[bytes, ...], Tuple[fhe.Value, ...]
         ],
         serialized_evaluation_keys: bytes,
+        force_serialization: bool = False,
     ) -> Union[bytes, fhe.Value, Tuple[bytes, ...], Tuple[fhe.Value, ...]]:
-        """Run the model on the server over encrypted data.
+        """Run the FHE circuit on encrypted data.
 
         Args:
-            serialized_encrypted_quantized_data (Union[bytes, fhe.Value, Tuple[bytes, ...], \
-                Tuple[fhe.Value, ...]]): The encrypted and quantized values to consider. If these
+            serialized_encrypted_quantized_data (Union[bytes, fhe.Value, Tuple[bytes, ...],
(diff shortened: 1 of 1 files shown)


andrei-stoian-zama:
Thank you for the PR!

Could you elaborate in which cases the serialization would not be needed? Would it be in the case when Client / Server run on the same machine ?

DeVikingMark:
> Thank you for the PR!
> 
> Could you elaborate in which cases the serialization would not be needed? Would it be in the case when Client / Server run on the same machine ?

@andrei-stoian-zama 

Yes, serialization isn't needed when Client and Server run on the same machine. I noticed that in local scenarios, the code was performing unnecessary serialization/deserialization operations that could be optimized.

In such cases, we can just pass fhe.Value objects directly, which is noticeably faster. This is especially useful during:
- local development and testing
- debugging
- running all components in a single process

That's why I added the force_serialization parameter - it helps optimize performance in these scenarios while maintaining backwards compatibility for distributed systems. Thanks for asking!

andrei-stoian-zama:
As I see it the code already does this: if you call server.run with fhe.Value, the result of the circuit execution will not be serialized back. this is done with the `        if inputs_are_serialized:` check at the end of the function.. can you please explain again how your use-case works ?

andrei-stoian-zama:
Thanks for explaining. We've made some changes to support TFHE-rs ciphertexts in the client/server (see the latest version of this file). These ciphertexts are not representable as fhe.Value and are always serialized. 

Serialization/deserialization is very fast and should not be a bottleneck. 

I would suggest:
- if, in your code, you need to accept serialized data and produce fhe.Value, you use the lower level api -> basically make a copy of the run function - you should be able to access all the required class members to fulfill your use case
- you simply deserialize the output of the run function when its input is serialized. this will be very fast

I think adding the function argument as you suggests complicates the API and may lead to confusion, especially since it would not work for the new TFHE-rs ciphertext types.  

Let me know if serialization is the performance bottleneck in your application and I'll re-open this discussion.

PR #1013, opened 2025-01-29
feat: CI on slab

andrei-stoian-zama:


fd0r (dismissed):

andrei-stoian-zama:
I merged #1016 which included these commits

PR #1022, opened 2025-02-21
fix: restore AdaptiveAvgPool2d in ResNet model

VolodymyrBg:
Replace fixed AvgPool2d with AdaptiveAvgPool2d in ResNet implementation.
This change allows the model to handle input images of varying sizes
by automatically adapting the pooling parameters to produce 1x1 output.
The adaptive pooling is more flexible and matches the original ResNet
architecture design.


jfrery:
Hey @VolodymyrBg,

Thanks for the PR. However, the change was made because concrete-ml doesn't support varying sizes onces compiled. We don't want to use adaptive pooling for now.

PR #1024, opened 2025-02-24
chore: eval llama lora to notebook

jfrery:


src/concrete/ml/quantization/linear_op_glwe_backend.py
```diff
@@ -92,19 +190,25 @@ def _per_channel_weight_quantization(
...
+
         # Avoid division by zero.
         weight_scale = torch.where(
             (w_max_vals > w_min_vals), weight_scale, torch.ones_like(weight_scale)
         )
         weight_scale = weight_scale.squeeze(-1)  # shape: (out_dim,)
 
-        weight_zp = torch.round((q_min - w_min_vals) / weight_scale).to(torch.float32)
+        # Quantization
+        weight_zp = torch.round(q_min - w_min_vals / weight_scale).to(torch.float32)
```
andrei-stoian-zama, line 209: couldn't you use the TorchUniformQuantizer here? (maybe just TODO in a next PR, no need to fix now)

src/concrete/ml/torch/hybrid_model.py
```diff
@@ -511,7 +516,15 @@ def forward(self, x: torch.Tensor, fhe: str = "disable") -> torch.Tensor:
         for module in self.remote_modules.values():
             module.executor = self.executor
 
-        result = self.model(x)
+        # Initialize the progress bar based on the number of remote modules
+        num_remote_modules = len(self.remote_modules)
+        with tqdm(total=num_remote_modules, desc="Remote Modules Inference") as pbar:
```
andrei-stoian-zama, line 521: compilation also has a tqdm on it, could your callback be used there too ?

`        for name in tqdm(self.module_names, desc="Compiling FHE layers"):`

PR #1034, opened 2025-03-04
Fix hybrid model requirements and clarify Python version in README

bonsainoodle:
This PR fixes the hybrid model use case by updating the requirements.txt, which was incomplete and had incorrect package versions.

Additionally, the README.md has been updated to explicitly specify the correct Python version (3.9), as using Python 3.10 caused compatibility issues.

Key changes:

Fixed missing and incorrect packages in requirements.txt
Added a note in README.md to recommend using Python 3.9 for this use case
This ensures developers can set up the project without unnecessary troubleshooting.


andrei-stoian-zama:
Thank you for the PR. However, I'm afraid we can't accept it:

The main issue with Python 3.10 is due to some interdependency issues between `apache_beam` and `multiprocess`. It seems this issue was introduced due to some updates in these packages and they dropped support for python 3.10. The notebook works with python3.12 for example. We can not pin package versions and a specific old version of Python (3.9) and block all other versions. It is up to the user to resolve dependency conflicts that are introduced by 3rd party package maintainers dropping support for older version.s

andrei-stoian-zama:
please see https://github.com/apache/beam/issues/24458

PR #1038, opened 2025-03-11
fix: change concrete version

andrei-stoian-zama:


pyproject.toml
```diff
@@ -38,8 +38,8 @@ python = ">=3.8.1,<3.13"
 # source, so we should only use this for Concrete Python's non-public releases
 # https://python-poetry.org/docs/1.7/repositories#project-configuration
 # concrete-python = {version="==2.7.0", source = "zama-pypi-cpu"}
-concrete-ml-extensions = "0.1.5"    
-concrete-python = {version="==2.9.0rc2.dev20250128", source = "zama-pypi-cpu"}
+concrete-ml-extensions = "0.1.6"
```
kcelia, line 41: why do you want to move to dev20250305 ? 
on cml we are still on dev20250128

andrei-stoian-zama: because I want to integrate the newer Concrete

kcelia (approved):

PR #1040, opened 2025-03-13
feat: use case to show post-processing CML circuit with TFHE-rs rust code

andrei-stoian-zama:
Adds a use-case that shows how to run a CML model, then export the ciphertexts to a TFHE-rs program that does additional processing.

First a CML decision tree is trained and exported to client/server deployment. The keys are generated and the inputs are encrypted.

Next, the model is run and the output is saved to disk. A TFHE-rs rust program reads the ciphertexts and evaluation keys and does argmax on the logics of the model. If the output class was 1, the program returns 16 zeros. Otherwise it returns 16 encrypted random values.

Todo:
- [x] fix pcc
- [x] make the rust code take the file names as arguments
- [x] add comments / conclusion
- [x] put the temporary files in temporary directory using tempfile.TemporaryDirectory
- [x] Add new test to fix coverage
- [] make the use case run in the CI: adding rust installation in the makefile (on going) -> conflict between cml and cml-ext !!


jfrery:
Had a look at the notebook. It's not clear what we are show casing here. Deserves some explanation I think.

use_case_examples/tfhers_interoperability/AuthenticationWithDecisionTree.ipynb
```diff
@@ -1,10 +1,31 @@
 {
  "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Introduction:\n",
+    "\n",
+    "Traditionally, Concrete ML (CML) relied exclusively on the [Concrete]() stack for model compilation. Today, we are expending its scope by introducing compilation support for TFHE-rs, enabling a new level of [interoperability]() between homomorphic encryption frameworks.\n",
```
andrei-stoian-zama, line 9: ```Traditionally, Concrete ML (CML) relied exclusively on the [Concrete]() stack for model compilation. Today, we are expending its scope by introducing compilation support for TFHE-rs, enabling a new level of [interoperability]() between homomorphic encryption frameworks.```

- don't add links yet, we don't have the docs for this yet
- don't use the "before/now" style -> this is a demo, not a changelog. Suggestion: remove this phrase. and rewrite `Interoperability between Concrete and TFHE-rs allows an application to combine Concrete ML with..`

use_case_examples/tfhers_interoperability/AuthenticationWithDecisionTree.ipynb
```diff
@@ -0,0 +1,695 @@
...
+    "## Introduction:\n",
+    "\n",
+    "\n",
+    "Concrete ML (CML) now supports model compilation to the TFHE-rs format, enabling deeper interoperability between the CML and TFHE-rs frameworks. This allows developers to combine CML’s intuitive, Python-based workflow with the performance and low-level control of TFHE-rs for advanced privacy-preserving applications.\n",
+    "\n",
+    "This notebook showcases a complete binary classification use case. We first perform encrypted inference using a CML decision tree model. The resulting ciphertexts are then exported to a TFHE-rs program for additional post-processing. In our example, this post-processing includes applying an `argmax` operation on the model's encrypted output, followed by a multiplication with a random plaintext token. If the predicted class is 0, the program returns a vector of zeros; otherwise, it returns the random token.\n",
+    "\n",
+    "This pattern represents a typical privacy-preserving server-side authentication flow, where access to a secure token is granted only if the encrypted model prediction satisfies a certain condition — without ever decrypting sensitive data throughout the process."
```
jfrery, line 14: It's not clear to me what `authentication` and `secure token` means here. If the secure token is meant to be shared as is with the server then there is a huge assumption that the server execution was correct.

use_case_examples/tfhers_interoperability/AuthenticationWithDecisionTree.ipynb
```diff
@@ -0,0 +1,695 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Introduction:\n",
+    "\n",
+    "\n",
+    "Concrete ML (CML) now supports model compilation to the TFHE-rs format, enabling deeper interoperability between the CML and TFHE-rs frameworks. This allows developers to combine CML’s intuitive, Python-based workflow with the performance and low-level control of TFHE-rs for advanced privacy-preserving applications.\n",
```
andrei-stoian-zama, line 10: `Concrete ML (CML) now supports model compilation to the TFHE-rs format,` -> `Concrete ML supports TFHE-rs format inputs and outputs for FHE ML models`

don't use `CML`, put `Concrete ML` everywhere please

also remove `performance` for TFHE-rs

use_case_examples/tfhers_interoperability/AuthenticationWithDecisionTree.ipynb
```diff
@@ -0,0 +1,693 @@
...
+       "DecisionTreeClassifier(n_bits={'op_inputs': 8, 'op_leaves': 8})"
+      ]
+     },
+     "execution_count": 11,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# For now, only 8-bit quantization is supported for TFHE-rs interoperability\n",
```
kcelia, line 472: @andrei-stoian-zama should we say why we are limited to 8 bits ?

andrei-stoian-zama: yes, that's fine

kcelia:
It looks like a flaky: https://github.com/zama-ai/concrete-ml/actions/runs/14104153158/job/39506912812?pr=1040 
pytest  tests/torch/test_compile_torch.py::test_compile_torch_or_onnx_networks[get_and_compile--FHE_simulation-BranchingGemmModule-5-relu]  --randomly-seed=1848616978  --randomly-dont-reset-seed is green on my side

conftest.py
```diff
@@ -533,6 +533,8 @@ def check_is_good_execution_for_cml_vs_circuit_impl(
         model: Union[Callable, QuantizedModule, QuantizedTorchEstimatorMixin],
         simulate: bool,
         n_allowed_runs: int = 5,
+        rtol: float = 1e-5,
```
andrei-stoian-zama, line 536: I don't think we need to change it because the flaky should be fixed by re-runnign when coverage is achieved

tests/deployment/test_client_server.py
```diff
@@ -403,6 +409,118 @@ def check_client_server_inference(
...
+        )
+    else:
+        assert isinstance(model, QuantizedModule)
+        q_y_pred_dev = model.fhe_circuit.encrypt_run_decrypt(q_x_test)
+
+    y_pred_dev = model.dequantize_output(q_y_pred_dev)
+    y_pred_dev = model.post_processing(y_pred_dev)
+
+    # Export intermediate files for TFHE-rs Rust inference ===
+    with open("evalkeys_tfhers.bin", "wb") as k, open("prediction_non_preprocessed.bin", "wb") as i:
```
andrei-stoian-zama, line 476: we can't test this in the CI, we don't have rust installed in the CI. we don't need to test the execution of the thfe-rs file, we can test only the serialization of the key

andrei-stoian-zama (commented):
One more line of coverage and it should be good

tests/deployment/test_client_server.py
kcelia: we can install rust no ?

kcelia: because, the only way i see to test this piece of code is to simulate a test with rust

```
    def get_tfhers_secret_key(self) -> Optional[bytes]:
        """Get the secret key that decrypts TFHE-rs ciphertexts.

        This secret key use used by the client object to encrypt TFHE-rs ciphertexts
        that are sent to the server, but also to decrypt outputs produced by
        a server-side computation, that can be a Concrete ML model or a TFHE-rs program.

        Returns:
            bytes: the serialized TFHE-rs secret key
        """
        return self._tfhers_sk
```

conftest.py
kcelia: The default output format when comparing two vectors is not very readable, especially for large vectors, it's hard to spot where the differences are. So I reused the same hyper-parameters as check_array and modified the display to highlight only the lines where values differ

PR #1043, opened 2025-03-19
Make deserialization and serialization optional in FHEModelServer.run

VolodymyrBg:
This PR adds an optional parameter auto_deserialize_serialize to the FHEModelServer.run method. When set to True (default), it maintains the current behavior of automatically deserializing input bytes and serializing output values. When set to False, it allows users to handle serialization and deserialization externally.


andrei-stoian-zama:
I'm sorry I don't understand how it works:

currently:
- if the inputs are fhe.Value (not serialized) -> the server.run uses them directly and does not serialize the FHE execution output
- if the inputs are bytes (serialized fhe.Value) -> the server run function deserializes, runs FHE, then serializes the results

I understand you want to handle serialization externally ->  thus you want to call this function with `bytes` ? but then we need to deserialize the bytes into fhe.value to run the circuit on them .. 

can you explain your use-case in mode detail please ?

PR #1060, opened 2025-04-02
fix: set executor during compile

andrei-stoian-zama:
In this way we can use the GLWE backend in `generate` right after compilation


jfrery (commented):
You can probably remove that part from the forward call then.

PR #1075, opened 2025-04-17
GPU tests in CI

kcelia:
Steps:
- Add new machine [g4dn.2xlarge](https://aws.amazon.com/ec2/instance-types/g4/) to `ci/slab.toml` under `[backend.aws.gpu_ciprofile]` tag, to pick up the SHA see: https://github.com/zama-ai/slab-github-runner
- Add the gpu marker, in conftest.py
- Target the tests to run with gpu
- Add gpu setup in .github/workflows/continuous-integration.yaml

Current issues: 
- [x] (solved) GPU not recognized in the slab runner
  The `start-runner-gpu-linux`  step runs on `ubuntu-24.04` and starts the GPU instance, `invidia-smi` won't work inside it.
- [x] No space left in the device
  The AMI size is defined at build time (see: `/dev/root` using `df -h`). Increasing the instance type won't affect disk space — we must expand the AMI storage and create a new snapshot:

      Reminder:
      A snapshot is an incremental backup of an EBS volume (the "hard drive" of an EC2 instance), capturing its exact state at a given moment.

      Steps:
      Launch an EC2 instanc:  Manually increase EBS storage via Configure Storage > Launch instance > Actions > Image & Templates > Create Image.

      <!> Don't forget to terminate the EC2 instance used for resizing and snapshot the ami image.
- [x] (solved) Error with concrete-python 2.24.0
    `pip install --extra-index-url https://pypi.zama.ai/gpu concrete-python` downloads `concrete-python 2.24.0` but CML is based on `concrete-python 2.10.0`. We encounter a breaking change -> `TypeError: fork() got an unexpected keyword argument 'optim_lsbs_with_lut'`

- [x] (solved) Reproducibility error with GPU
     ```FAILED tests/torch/test_hybrid_converter.py::test_hybrid_glwe_correctness[2048-False-10] - RuntimeError: Deterministic behavior was enabled with either `torch.use_deterministic_algorithms(True)` or `at::Context::setDeterministicAlgorithms(true)`, but this operation is not deterministic because it uses CuBLAS and you have CUDA >= 10.2. To enable deterministic behavior in this case, you must set an environment variable before running your PyTorch application: CUBLAS_WORKSPACE_CONFIG=:4096:8 or CUBLAS_WORKSPACE_CONFIG=:16:8. For more information, go to https://docs.nvidia.com/cuda/cublas/index.html#results-reproducibility```
  I added a fixture `enforce_gpu_determinism` 

All tests take less than 20 minutes.  

closes  [4723](https://github.com/zama-ai/concrete-ml-internal/issues/4723)
ref [previous work](https://github.com/zama-ai/concrete-ml/compare/main...feat/add_gpu_tests)

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -451,6 +451,136 @@ jobs:
             echo "No flaky tests were reported!"
           fi
 
+  # ----------------
+  start-runner-gpu-linux:
+      needs: [matrix-preparation]
+      permissions: {}
+      name: Start GPU SLAB runner
+      # This label is irrelevant once the SLAB runner starts.
+      runs-on: ubuntu-24.04
+      
+      timeout-minutes: 15
+      outputs:
+        label: ${{ steps.start-slab-runner.outputs.label }}
+      steps:
+        - name: Configure AWS credentials
+          uses: aws-actions/configure-aws-credentials@ececac1a45f3b08a01d2dd070d28d111c5fe6722
+          with:
+            aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+            aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+            aws-region: ${{ secrets.AWS_REGION }}
+              
+        - name: Start SLAB GPU runner
+          id: start-slab-runner
+          if: ${{ !cancelled() }}
+          uses: zama-ai/slab-github-runner@79939325c3c429837c10d6041e4fd8589d328bac
+          with:
+              mode: start
+              github-token: ${{ secrets.SLAB_ACTION_TOKEN }}
+              slab-url: ${{ secrets.SLAB_BASE_URL }}
+              job-secret: ${{ secrets.JOB_SECRET }}
+              backend: aws
+              profile: gpu_ciprofile 
+
+  # GPU-availability:
+  #   name: GPU Availability
+  #   runs-on: ${{ needs.start-runner-gpu-linux.outputs.label }}
+  #   needs: start-runner-gpu-linux
+    
+  #   steps:
+  #     - name: Check GPU availability
+  #       run: |
+  #         echo "Checking for GPU..."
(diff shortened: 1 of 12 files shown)


kcelia:
<!!!!!> The ci blocks on one test, i didn't get enough time to deeply look at it.

kcelia:
enabling all tests together, is more than 30 minutes...

conftest.py
```diff
@@ -282,17 +284,33 @@ def is_weekly_option(request):
 
 
 @pytest.fixture
-def get_device_for_compilation(request):
```
kcelia, line 285: it was used in one test, i find it right to change it.
Unless I miss something.

kcelia: I renamed get_device_for_computation to get_device because I wanted a name that would reflect a broader purpose than just computation.

Since I was mainly focusing on the GPU part, I didn’t anticipate the upcoming changes. I don’t have my laptop with me, so I fixed the CI as best as I could using my phone.

Feel free to revert if needed

andrei-stoian-zama (approved):
well done!

.github/workflows/continuous-integration.yaml
```diff
@@ -451,6 +451,136 @@ jobs:
...
+      - name: Install GPU concrete-python
+        id: install-gpu-concrete-python
+        run: |
+          # Show CP version
+          poetry run pip show concrete-python || echo "concrete-python not installed"
+          poetry run pip uninstall -y concrete-python
+
+          
+          # poetry run pip install --extra-index-url https://pypi.zama.ai/gpu concrete-python
+          poetry run pip install --extra-index-url https://pypi.zama.ai/gpu concrete-python==2.10.0
```
andrei-stoian-zama, line 549: we need to get the version that is currently install and install == that version

something like:

```
CONCRETE_WITH_VERSION= $(pip freeze | grep concrete-python)
poetry run pip uninstall -y concrete-python
poetry run pip install --extra-index-url https://pypi.zama.ai/gpu $CONCRETE_WITH_VERSION
```

PR #1091, opened 2025-05-21
Ci/fix permissions

aquint-zama:


.github/workflows/publish_aws_ami.yaml
```diff
@@ -23,24 +26,27 @@ jobs:
       # FIXME: We still need to check that it matches a version tag
       # This won't work on automatically triggered based on release
       - name: Checkout Concrete ML Repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          persist-credentials: 'false'
```
kcelia, line 31: By default the param is set to True, right ?
I m wondering :
does it mean that , before, some actions were able ( because they have access to GITHUB_TOKEN)  to push things that we didn't know ?

aquint-zama: by default it's set to true, it keep the github token locally in case you need it for further actions, the common issue being packaging it into a docker image or equivalent and sharing publicly a valid token (with all permissions as it was not scoped)

kcelia: thanks for the explanation

kcelia (commented):
We can run all workflows on this branch to ensure everything still works, or test them as we go @zama-ai/machine-learning

andrei-stoian-zama (approved):

kcelia (approved):

PR #1124, opened 2025-09-11
fix path certificats in CI

kcelia:


Makefile
```diff
@@ -50,6 +50,7 @@ setup_env:
 	poetry run python --version
 	poetry run python -m pip install keyring
 	poetry run python -m pip install -U pip wheel
+	pip install pip-system-certs
```
kcelia, line 53: that s the fix to get the right credentials

.github/workflows/continuous-integration-common.yaml
```diff
@@ -205,7 +205,20 @@ jobs:
...
+              echo "✅ Success on attempt $i"
+              break
+            fi
+            if [ $i -eq $max_retries ]; then
+              echo "❌ Failed after $max_retries attempts"
+              exit 1
+            fi
+            echo "⚠️ Failed attempt $i, retrying in 10s..."
+            sleep 10
+          done
```
kcelia, line 221: the loop I told you about

.github/workflows/continuous-integration-common.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          persist-credentials: 'false'
+          persist-credentials: 'true'
```
kcelia, line 96: I need to store the credentials for the git Los

jfrery: got it thanks

jfrery (changes requested):
Just one small change and it's good to go

jfrery (approved):

PR #1147, opened 2025-10-28
 remove dependabot updates

kcelia:


.github/dependabot.yaml
jfrery, line 1: We shouldn't delete everything I think.

jfrery (changes requested):

jfrery (approved):

